PR #107, opened 2014-12-05
Filter empty keys in 'tree' (folder nodes)

### LINKED ISSUE #106: tree api function returns kv folders as keys with empty values
```
{{range tree "consul-template/master/monitoring"}}
{{.Key | replaceAll "/" "."}} = {{.Value}}{{end}}
```

gives

```
graphite.port = 2003
graphite.server = "graphite"
graphite. = 
statsd.port = 8125
statsd.server = "statsd"
statsd. = 
 = 
```

ryanuber:
Fixes #106.

@sethvargo this should work for the 99% case, but what I realized in doing this is that we eliminate the possibility of getting the values of "folder" nodes in both `ls` and `tree`. I thought maybe we should add an extra condition, which would include the "folder" key if it had a value, but not if it didn't. This complicates things a little more, and I'm not sure we would ever need it, so it's not part of this PR. Just something to think about.

### DIFF
--- a/template.go
+++ b/template.go
@@ -139,7 +139,15 @@ func (t *Template) Execute(c *TemplateContext) ([]byte, error) {
 			return c.Services[d.Key()], nil
 		},
 		"tree": func(s string) []*util.KeyPair {
-			return c.KeyPrefixes[s]
+			var result []*util.KeyPair
+			// Filter empty keys (folder nodes)
+			for _, pair := range c.KeyPrefixes[s] {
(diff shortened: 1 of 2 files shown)


carlpett:
@ryanuber @sethvargo Coming along right here, some 550 days later :) How hard would it be to fix? Use-case is that I'm generating haproxy config from the kv, where the application developer creates a backend config by adding a folder and some options with keys within. However, when the defaults are good enough, we still need to add some key so that `tree` picks it up.

PR #141, opened 2015-01-07
Switch to using hashicorp/consul/api

ryanuber:
Depends on hashicorp/consul#581.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -14,8 +14,8 @@ import (
 	"syscall"
 	"time"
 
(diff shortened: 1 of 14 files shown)


ryanuber:
Unit tests fail because hashicorp/consul#581 is not yet merged. They should pass once we have that in place.

sethvargo:
:+1: I'll leave this up to you to merge once things are good, but I like this MUCH better!

PR #145, opened 2015-01-11
Add options for enabling HTTPS

### LINKED ISSUE #143: https api with consul-template
We put nginx in front of the consul web ui for a variety of reasons, and were wondering if there is something we are missing regarding using consul-template with an https instead of http api?

ryanuber:
Adds the `https` and `insecure` bool config options for talking to an HTTPS consul instance.
Fixes #143.

/cc @sethvargo

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,8 +33,8 @@ Usage
 | ------ | -------- |------------ |
 | `consul`    | _(required)_ | The location of the Consul instance to query (may be an IP address or FQDN) with port. |
 | `template`  | _(required)_ | The input template, output path, and optional command separated by a colon (`:`). This option is additive and may be specified multiple times for multiple templates. |
-| `https`     | | Use HTTPS while talking to Consul. Requires the Consul server to be configured to serve secure connections.
-| `insecure`  | | Ignore certificate warnings. Only used if `https` is enabled.
+| `ssl`       | | Use HTTPS while talking to Consul. Requires the Consul server to be configured to serve secure connections.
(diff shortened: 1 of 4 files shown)


re5et:
:+1:

README.md
```diff
@@ -33,6 +33,8 @@ Usage
 | ------ | -------- |------------ |
 | `consul`    | _(required)_ | The location of the Consul instance to query (may be an IP address or FQDN) with port. |
 | `template`  | _(required)_ | The input template, output path, and optional command separated by a colon (`:`). This option is additive and may be specified multiple times for multiple templates. |
+| `https`     | | Use HTTPS while talking to Consul. Requires the Consul server to be configured to serve secure connections.
+| `insecure`  | | Ignore certificate warnings. Only used if `https` is enabled.
```
sethvargo: I think we should name this something more specific like "--no-ssl-verify" or something. "Insecure" feels sad to me. What do you think? Any other ideas?

sethvargo: Also, I think the option should be "secure", not HTTPS. HTTPS describes the security protocol, but the config is describing the connection, which is "secure". Alternatively, I like "ssl" over "https"

ryanuber:
@sethvargo self signed certs are kind of sad. But good call, I changed the option names. I figured they are more obviously related if they both have `ssl` in them.

sethvargo:
:+1: I'll merge this tomorrow if you don't beat me to it

Issue #64, opened 2014-11-07
Dependency graph not properly built when nested inside dynamic conditional 


jrnt30:
We were attempting to dynamically add in ranges per DC in a somewhat "creative" solution to allow a single template to service a dynamic number of DCs.  We are massing the output from the /v1/catalog/datacenters endpoint into a JSON file to indicate the DC's health/addressability of the DC and attempting to use this file in conditionals to add in the range or not. 

What we've observed is the dependency graph does not get properly generated when the range is inside a conditional block that is reliant upon the parsed JSON. 

**Note: I don't know how feasible it would be, but allowing the creation of a dependency against a DC that currently does not exist would fix this issue by simply removing the need to do the conditional in the first place.**

Given the contents of dcs.json are: 
```json
{
  "us-east-1":true, 
  "us-west-2":false
}
```

**Fails when conditional is based solely on the dynamic value via the file pipeline/parsed JSON**
```liquid
{{ with $dcs := file "dcs.json" | replaceAll "-" "_" | parseJSON }}
  {{if $dcs.us_east_1}}
    {{ range service "consul@us-east-1"}}node {{.Node}}{{end}}
  {{end}}
  {{if $dcs.us_west_2}}
    us-west-2 SHOULDN'T APPEAR
  {{end}}
{{end}}
```

**Works "properly" because the dependency was created outside conditional, however not really an option because adding in a range to a non-addressable DC abends the consul-template**
```liquid
{{ range service "consul@us-east-1"}}{{end}}
{{ with $dcs := file "dcs.json" | replaceAll "-" "_" | parseJSON }}
  {{if $dcs.us_east_1}}
    {{ range service "consul@us-east-1"}}node {{.Node}}{{end}}
  {{end}}
  {{if $dcs.us_west_2}}
    us-west-2 SHOULDN'T APPEAR
  {{end}}
{{end}}
```

**Works properly with non-dynamic conditional**
```liquid
{{ with $dcs := true }}
  {{if $dcs }}
    {{ range service "consul@us-east-1"}}node {{.Node}}
    {{end}}
  {{end}}
{{end}}
```

armon:
I think the N pass does solve this. Even with conditionals, let's say you are in the "yes" path at time t0. Sometime in the future, at time t1 we switch to the "no" path (maybe a value changes in Consul), at that point the evaluator will notice we are missing a dependency, and will trigger an N+1 evaluation. Then we redo the evaluation once the data is ready. 

[...]

PR #146, opened 2015-01-12
Add "datacenters" template function to list datacenters

ebroder:
This is decidedly imperfect (see in particular the synthesized `QueryMeta` in `Datacenters.Fetch` - I was mostly just looking for something that would make it run to see if it worked), but I'm curious whether there's interest in a cleaned up version of this function.

I suspect in practice the utility of a `datacenters` function will be limited by #64, but it seems like an element of the consul catalog that's noticeably missing from the current functionality. Thoughts?

### DIFF
--- a/dependency/datacenters.go
+++ b/dependency/datacenters.go
@@ -1,0 +1,45 @@
+package dependency
+
+import (
+	"github.com/hashicorp/consul/api"
+	"log"
+	"sort"
+	"time"
+)
+
+type Datacenters struct {
+}
+
+func (d *Datacenters) Fetch(client *api.Client, options *api.QueryOptions) (interface{}, *api.QueryMeta, error) {
+	log.Printf("[DEBUG] (%s) querying Consul with %+v", d.Display(), options)
+
+	catalog := client.Catalog()
+	dcs, err := catalog.Datacenters()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	log.Printf("[DEBUG] (%s) Consul returned %d datacenters", d.Display(), len(dcs))
+	sort.Strings(dcs)
+	return dcs, &api.QueryMeta{LastIndex: uint64(time.Now().Unix())}, nil
+}
+
+func (d *Datacenters) HashCode() string {
+	return "Datacenters"
+}
+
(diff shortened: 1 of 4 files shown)


sethvargo:
Hey @ebroder 

Thank you for the PR! This looks really good. I'm going to defer to @armon, but my initial thought is "can I see a real-life use case". You are totally right that #64 is going to make this a blocker, but assuming 64 wasn't a thing, can you provide a "real-life" use case for when you would use this functionality? What would the in and out templates look like?

ebroder:
The specific thing I'm thinking of doing right now is generating a zone file for `.consul` so that I can isolate the DNS interface and caching of the DNS records from the consul cluster.

(We've recently been having some issues with consul's raft cluster losing consensus, and DNS and API queries failing with it. We definitely need to deal with that, and will follow up on that separately, but it's gotten me thinking about ways to make sure that our discovery code still works even when consul is having problems)

dependency/datacenters.go
```diff
@@ -0,0 +1,45 @@
...
+	sort.Strings(dcs)
+	return dcs, &api.QueryMeta{LastIndex: uint64(time.Now().Unix())}, nil
+}
+
+func (d *Datacenters) HashCode() string {
+	return "Datacenters"
+}
+
+func (d *Datacenters) Key() string {
+	return ""
```
armon: Probably still want a key even if hard coded

armon:
The main issue here is that the `catalog.Datacenters()` does not support blocking. This means we cannot block on an edge trigger from Consul efficiently. This dependency will currently sit in a hot loop continuously polling for updates.

armon:
I think the ideal answer is that Consul should support this upstream, but for a number of reasons that is very challenging. Simpler answer is that we should do a client side sleep and only poll for changes every ~15-30 seconds. It's pretty rare to add a new DC, so this seems fine to me.

ebroder:
Hmm, I'm not entirely sure what the right way is to implement that sleep. Is the right idea to just add something like

```go
if options.WaitTime != nil {
	time.Sleep(options.WaitTime)
}
```

at the beginning of the `Fetch` method?

PR #148, opened 2015-01-12
Add support for basic auth

bion:


sethvargo:
Hey @bion 

Thank you for the PR. This is looking really good!

What do you think about making this a single option (`--auth` or `--basic-auth`) that accepts `USERNAME[:PASSWORD]` and we can split on the first colon on the backend. I think this is more closely how curl behaves, so I would like to mirror that API as closely as possible. What do you think?

/cc @ryanuber

ryanuber:
@bion I agree with @sethvargo, it would be nice to have these as `user[:pass]` on the command line. I vote for the `-auth` flag.

cli.go
```diff
@@ -279,6 +283,10 @@ func bootstrap(config *Config, dry bool, once bool) (*Runner, *watch.Watcher, er
 			},
 		}
 	}
+	if (config.HttpAuth.Username != "") || (config.HttpAuth.Password != "") {
```
ryanuber: Probably don't even need to do this if check here.

config.go
```diff
@@ -32,6 +37,8 @@ type Config struct {
 	// Consul. This requires Consul to be configured to serve HTTPS.
 	SSL bool `mapstructure:"ssl"`
 
+	HttpAuth HttpAuth `mapstructure:"HttpAuth"`
```
ryanuber: The `mapstructure` tag maps the value to the key in a config file, which should follow the snake-chaining (lowercase with underscores) convention.

sethvargo: But if we follow the "single-value" for authentication, this would likely just become a string in the config.

config.go
```diff
@@ -65,6 +72,14 @@ func (c *Config) Merge(config *Config) {
 		c.SSLNoVerify = true
 	}
 
+	if (config.HttpAuth.Username != "") || (config.HttpAuth.Password != "") {
+		c.HttpAuth = config.HttpAuth
+	}
+
+	if config.HttpAuth.Password != "" {
```
ryanuber: This should probably be all-or-nothing for the best clarity. If user or password are present, then merge, else do nothing.

config_test.go
```diff
@@ -119,6 +119,30 @@ func TestMerge_HttpsOptions(t *testing.T) {
 	}
 }
 
+func TestMerge_BasicAuthOptions(t *testing.T) {
+	{
```
ryanuber: Nitpick: We don't need to wrap this test in a block since the variables aren't used in any other scope.

ryanuber:
@bion I added a few comments to the code as well. Thanks for doing this!

bion:
Thanks for the comments, guys. This is my first foray into Go so please forgive any naivete. I'll address these comments and @ you when it's ready for another look.

bion:
@ryanuber @sethvargo back to you guys.

sethvargo:
Hey @bion 

Thank you for making those changes. I am working on some internal refactorings for Consul Template and I will merge this after those have solidified.

sethvargo:
Cleaned up and merged in 9e819cc. Thanks!

PR #159, opened 2015-01-16
Try the new Docker containers on Travis

sethvargo:


sethvargo:
It's consistently slower :frowning:

PR #161, opened 2015-01-16
Quiesce on a per-template basis

### LINKED ISSUE #154: Quiescence should be per-template
Right now, quiescence is calculated globally for the runner, but it should really be on a per-template basis.

sethvargo:
Fixes #154

### DIFF
--- a/runner.go
+++ b/runner.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 	"time"
 
 	dep "github.com/hashicorp/consul-template/dependency"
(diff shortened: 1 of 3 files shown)


mitchellh:
So far so good to me. Only thing I don't see: is there any way a template is deregistered, and if so, we should remove it from the quiescence map.

sethvargo:
@mitchellh that's not something that is currently possible (directly). When we reload configuration, we make all new template objects, so the problem should be avoided.

mitchellh:
Right, but it looks like we tie the path directly into this map which is part of runner. Or do we create a new runner, too?

sethvargo:
@mitchellh we [create an entirely new runner object](https://github.com/hashicorp/consul-template/blob/master/cli.go#L148-L154)

mitchellh:
Then :+1:

PR #164, opened 2015-01-20
Fix documentation

guillaumelecerf:


sethvargo:
Good catch, thanks!

guillaumelecerf:
@sethvargo : thanks ;)

PR #177, opened 2015-02-02
Fix runner scope during SIGHUP handling. Fixes #175.

### LINKED ISSUE #175: Sending a HUP signal does not reload the configuration
I'm testing on a Mac, consul-template version 0.6.0.

I run it like this:

`~/consul-template ○ → ./consul-template -template "/tmp/sample.ctmpl:/tmp/sample.conf:date" -token mytoken`

I see it change my file correctly and execute my command. Then, I execute a `kill -HUP <PID>` to reload the configuration, and it dies:

    Received HUP, reloading configuration...
    ~/consul-template ○ →

ryanuber:
Fixes an issue where SIGHUP would try to reload, but end up stopping consul-template.

@sethvargo can you take a look?

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -158,7 +158,7 @@ func (cli *CLI) Run(args []string) int {
 			case syscall.SIGHUP:
 				fmt.Fprintf(cli.errStream, "Received HUP, reloading configuration...\n")
 				runner.Stop()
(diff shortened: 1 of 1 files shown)


sethvargo:
Thanks!

PR #180, opened 2015-02-04
Add missing locking around Brain's maps.

nelhage:
It's unclear whether a single Brain is intended to be used from multiple
goroutines at once, but Remember takes out the lock when it writes to
these maps, and the new batching goroutine in Runner.Start means it's
possible (if relatively unlikely) for Remember() calls to happen
concurrently with rendering templates.

Take out the locks to be safe.

### DIFF
--- a/template_functions.go
+++ b/template_functions.go
@@ -22,6 +22,9 @@ func datacentersFunc(brain *Brain,
 
 		addDependency(used, d)
 
+		brain.Lock()
+		defer brain.Unlock()
+
 		if result, ok := brain.datacenters[d.HashCode()]; ok {
 			return result, nil
 		}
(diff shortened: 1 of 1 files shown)


sethvargo:
Nice. Good catch! Do you think we should just make some kind of "getter" on the brain such that we don't have to lock/unlock when accessing? That way the template functions could just call `brain.Get(d)` for example?

What do you think?

nelhage:
That might make sense. If you did that, you could perhaps even collapse the `Brain`'s maps into a single `map[dep.Dependency]interface{}` and move the type assertions into the template functions, making `Brain` completely agnostic as to the structure of the saved data.

sethvargo:
Okay. I'll think about this a bit. This is definitely a bug and definitely something we should fix, but I don't have spare cycles to try out some of the different implementations that are floating in my head at the moment. I will re-visit this shortly.

PR #181, opened 2015-02-04
Add brain.recall

sethvargo:
Closes https://github.com/hashicorp/consul-template/pull/180

This adds a new Brain API that locks around getting data.

/cc @mitchellh @nelhage

### DIFF
--- a/template_functions.go
+++ b/template_functions.go
@@ -22,8 +22,8 @@ func datacentersFunc(brain *Brain,
 
 		addDependency(used, d)
 
-		if brain.Remembered(d) {
-			return brain.Recall(d).([]string), nil
+		if value, ok := brain.Recall(d); ok {
+			return value.([]string), nil
 		}
 
 		addDependency(missing, d)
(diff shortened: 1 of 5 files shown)


template_functions.go
```diff
@@ -22,8 +22,8 @@ func datacentersFunc(brain *Brain,
 
 		addDependency(used, d)
 
-		if result, ok := brain.datacenters[d.HashCode()]; ok {
-			return result, nil
+		if brain.Remembered(d) {
```
nelhage: This has the potential race condition for the metric to bet forgotten between `.Remembered()` and `.Recall()`, since you give up the lock in between. It may be unlikely with the current code, but it suggests this isn't quite the right API.

sethvargo: Okay, I just pushed up a change to make it more atomic.

sethvargo:
@nelhage I just pushed up another commit which makes the read from recall atomic. If you want to give this another pass-over, that'd be great! :smile:

mitchellh:
LGTM. A lot better.

nelhage:
yeah, I think this makes sense to me

PR #183, opened 2015-02-05
Add -max-stale and configurable -batch-size

sethvargo:
This adds a new option for `-max-stale` (config and CLI flag) that allows CT to query non-leaders. It also adds a new `-batch-size` option for the user to customize the size of the batch views are processed in.

/cc @mitchellh @armon

### DIFF
--- a/config.go
+++ b/config.go
@@ -48,6 +48,9 @@ type Config struct {
 	// ConfigTemplates is a slice of the ConfigTemplate objects in the config.
 	ConfigTemplates []*ConfigTemplate `mapstructure:"template"`
 
+	// BatchSize is the size of the batch when polling multiple dependencies.
+	BatchSize int `mapstructure:"batch_size"`
+
 	// Retry is the duration of time to wait between Consul failures.
 	Retry    time.Duration `mapstructure:"-"`
 	RetryRaw string        `mapstructure:"retry" json:""`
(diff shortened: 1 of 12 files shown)


config.go
```diff
@@ -139,6 +158,17 @@ func ParseConfig(path string) (*Config, error) {
...
 	config.Path = path
 
+	// Parse the MaxStale component
+	if raw := config.MaxStaleRaw; raw != "" {
+		stale, err := time.ParseDuration(raw)
+
+		if err == nil {
+			config.MaxStale = stale
+		} else {
+			errs = multierror.Append(errs, fmt.Errorf("max_stale invalid: %v", err))
```
armon: In Consul, if the value is less than 1000 we multiply by time.Second. Otherwise `max_stale = 5` means 5 nanoseconds, which is pretty useless and causes every query to be done twice. I'd add a similar guard.

sethvargo: @armon this is parsed as a time.Duration, so I think the user needs to specify units, don't they? Like `-max-stale 10s` or `max_stale = "10s"`. I don't think Go will let you put a unit-less value here.

armon:
LGTM otherwise.

config.go
armon: I think it does allow it. I might be wrong though.

armon: @sethvargo NVM, looks like not possible.http://play.golang.org/p/lCrLAgtfYF

sethvargo: If I pass in a non-string:

```
* 'max_stale' expected type 'string', got unconvertible type 'int'
```

If I pass in a string without units:

```
* max_stale invalid: time: missing unit in duration 2
```

Issue #188, opened 2015-02-10
`consul-template -once` will hang forever if the rendered template has not changed


nelhage:
Simple reproducer:

```console
$ echo HELLO > file.template && rm -f file
$ CONSUL_TEMPLATE_LOG=debug consul-template -template file.template:file -once
2015/02/10 03:30:30 [INFO] (runner) creating new runner (dry: false, once: true)
2015/02/10 03:30:30 [INFO] (runner) creating consul/api client
2015/02/10 03:30:30 [INFO] (runner) creating Watcher
2015/02/10 03:30:30 [INFO] (runner) starting
2015/02/10 03:30:30 [DEBUG] (runner) running initial templates
2015/02/10 03:30:30 [INFO] (runner) running
2015/02/10 03:30:30 [DEBUG] (runner) checking template file.template
2015/02/10 03:30:30 [DEBUG] (runner) checking ctemplate &{Source:file.template Destination:file Command:}
2015/02/10 03:30:30 [DEBUG] (runner) rendered: true
2015/02/10 03:30:30 [INFO] (runner) updating dependencies
2015/02/10 03:30:30 [INFO] (runner) once mode and all templates rendered
2015/02/10 03:30:30 [INFO] (runner) stopping
2015/02/10 03:30:30 [INFO] (watcher) stopping all views
$ CONSUL_TEMPLATE_LOG=debug consul-template -template file.template:file -once
2015/02/10 03:30:31 [INFO] (runner) creating new runner (dry: false, once: true)
2015/02/10 03:30:31 [INFO] (runner) creating consul/api client
2015/02/10 03:30:31 [INFO] (runner) creating Watcher
2015/02/10 03:30:31 [INFO] (runner) starting
2015/02/10 03:30:31 [DEBUG] (runner) running initial templates
2015/02/10 03:30:31 [INFO] (runner) running
2015/02/10 03:30:31 [DEBUG] (runner) checking template file.template
2015/02/10 03:30:31 [DEBUG] (runner) checking ctemplate &{Source:file.template Destination:file Command:}
2015/02/10 03:30:31 [DEBUG] (runner) rendered: false
2015/02/10 03:30:31 [INFO] (runner) updating dependencies
[HANG]
```

The problem seems to be that `r.render` returns `false` because there is nothing to do, and so we don't populate `r.renderedTemplates`, and so `allTemplatesRendered` returns false. So we hang forever waiting for new data, which will never arrive, because we didn't actually take out any watches.

[...]

PR #193, opened 2015-02-11
Record templates that would have rendered separate from those that did

sethvargo:
Fixes https://github.com/hashicorp/consul-template/issues/188

/cc @ryanuber @armon

### DIFF
--- a/runner.go
+++ b/runner.go
@@ -290,19 +290,28 @@ func (r *Runner) Run() error {
 			log.Printf("[DEBUG] (runner) checking ctemplate %+v", ctemplate)
 
 			// Render the template, taking dry mode into account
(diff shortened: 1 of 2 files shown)


armon:
LGTM

ryanuber:
@sethvargo LGTM. Is there any way to test the functionality, e.g., run in once mode against a template path which already has the most up-to-date content, ensure we don't hang and no commands run? Not sure if we have some helpers to make that easy or not.

sethvargo:
@ryanuber I started writing that test, but it was really fragile. I'll see what I can come up with.

sethvargo:
@ryanuber okay, I added a test.

PR #196, opened 2015-02-11
Send errors from the view to the watcher's error channel

sethvargo:
/cc @armon

### DIFF
--- a/watch/view.go
+++ b/watch/view.go
@@ -85,13 +85,8 @@ func (v *View) poll(viewCh chan<- *View, errCh chan<- error) {
 		case err := <-fetchErrCh:
 			log.Printf("[ERR] (view) %s %s", v.display(), err)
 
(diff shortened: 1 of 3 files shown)


watch/view.go
```diff
@@ -85,13 +85,8 @@ func (v *View) poll(viewCh chan<- *View, errCh chan<- error) {
...
 
-			// Intentionally do not send the error back up to the watcher and just
-			// retry. Eventually, once Consul API implements errwrap and multierror,
-			// we can check the "type" of error and conditionally alert back, but for
-			// now we will just swallow the error and continue.
-			// if err.Contains(Something) {
-			//   errCh <- err
-			// }
+			// Push the error back up to the watcher
+			errCh <- err
```
armon: I'd guard the push with the stopCh as well, in case the watcher is shutdown

PR #201, opened 2015-02-19
Added canonical import path to main.

jimmcslim:
Allows source to be easily built by centurylink/golang-builder (Docker
image) without a Go environment otherwise installed.


sethvargo:
Thanks!

Issue #205, opened 2015-02-24
Provide insight into resource utilization implied by templates


ryanbreen:
A poorly written template can [D]DoS the Consul servers, representing a real risk to the health of the cluster.  consul-template should provide some feedback when the template(s) will create a high number of views.  In that case, you might even refuse to run unless something like a `-force` flag is passed.

Additionally, it would be cool if `-dry` provided accounting for the number of views.  That would make it a pseudo-query planner and a more useful tool when attempting to minimize the impact of templates.

(Note that I'm using "views" as a proxy for "things that do potentially expensive stuff.  If there are leading indicators of template complexity other than view count, they would be cool to include in this accounting.)

sethvargo:
@ryanbreen thank you for opening an issue. While this is a great feature request, I am afraid it is not possible due to the dynamic nature of how Consul Template compiles dependencies. This would have been possible in the original versions of Consul Template, but since [v0.6.0](https://github.com/hashicorp/consul-template/releases/tag/v0.6.0), Consul Template implements n-pass evaluation. This allows the user to do something like:

```liquid
{{range ls "active_services"}}
{{range service .Key}}
{{.Address}}:{{.Port}}
{{end}}
{{end}}
```

This template is rather simple - it queries a list of key-value pairs in the KV store called `active_services` and, for each of those entires, it attemps to watch that service. This hypothetical template would allow operators to add/remove services from this template simply by changing a value in the KV store. So why is this relevant?

If `active_services` contains 1 key, this template requires 2 views (`ls` + `service` x 1). But if the KV store has 50 keys, this template requires 51 views (`ls` + `service` x 50). Even more complicated, those are _dynamic_ values, so the template could require 5 services to start (which might be deemed "healthy" or "ok"), but then over time grow to 500. This is a classic n+1 problem and I believe this is something that users of Consul Template (and related tooling) need to be cogniscenet of when using the tool. I hope this clarifies why we cannot know the complexity of an operation due to the highly dynamic and ever-changing nature of dependencies.

I would not be opposed to adding some warnings to the README, etc, but I do not think there is an appropriate or viable UX where Consul Template needs to decide for the user. @armon and I have discussed this pretty extensively and have come to the conclusion this is a problem that should be solved in Consul moreso that Consul Template. Consul 0.5 is a step in the right direction and much more performant than older versions, but we have some ideas about how to make the Consul Template use case even better.

[...]

PR #206, opened 2015-02-24
Warn the user if they have over 128 watches running

sethvargo:
Relates to https://github.com/hashicorp/consul-template/issues/205

### DIFF
--- a/runner.go
+++ b/runner.go
@@ -127,20 +127,20 @@ func (r *Runner) Start() {
 			}
 		}
 
-		// If we are running in once mode and all our templates have been rendered,
-		// then we should exit here.
-		if r.once && r.allTemplatesRendered() {
-			log.Printf("[INFO] (runner) once mode and all templates rendered")
-			r.Stop()
-			return
-		}
-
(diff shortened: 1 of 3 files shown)


watch/watcher.go
```diff
@@ -136,6 +136,16 @@ func (w *Watcher) Remove(d dep.Dependency) bool {
 	return false
 }
 
+// Size returns the number of views this watcher is watching.
+func (w *Watcher) Size() int {
+	w.Lock()
+	defer w.Unlock()
+
+	log.Printf("[INFO] (watcher) getting size")
```
armon: Probably overly verbose. I don't think this log is ever useful.

runner.go
```diff
@@ -129,6 +135,14 @@ func (r *Runner) Start() {
 			return
 		}
 
+		// Warn the user if they are watching too many dependencies.
+		if r.watcher.Size() > saneViewLimit {
```
armon: Maybe move this before the `once` check so that the output still shows up for that case?

armon:
LGTM

Issue #210, opened 2015-03-03
Service health filters don't trigger for warning/critical state


cwstrommer:
Consul version: 0.7.0 (also tested on 0.5.0)

I'm configuring a script to put haproxy backends in maintenance mode (as opposed to removing them for haproxy.cfg and reloading the config).  The template looks something like this:

    {{ range service "nginx" "passing" }}echo "enable server nginx:8080/{{ .ID }}" | socat stdio /run/haproxy/admin.sock
    {{ end }}
    {{ range service "nginx" "warning, critical" }}echo "disable server nginx:8080/{{ .ID }}" | socat stdio /run/haproxy/admin.sock
    {{ end }}

When the service is in passing state I see the proper enable line.  However when the service is in warning or critical state I get nothing at all.  
If I change the condition on the second line to either "warning" or "critical" by itself nothing changes.  
If I instead set the second line's conditions to "passing, warning, critical" I do get the line, but then I will also get it when the service is passing.  
I suspect that #116 is missing combinations for non-passing conditions.

PR #212, opened 2015-03-05
Fix service filtering

sethvargo:
This fixes a pretty critical bug in how we filter services, specifically around how we handle "any" vs "passing" vs "(other things)".

"any" is a special case in that it cannot and should not be specified with other status filters. In the past, it was permitted to do this:

```liquid
{{ range services "nginx" "any, passing" }}{{ end }}
```

This led to user confusion. Users thought this meant "give me any service that is passing", but it really means "give me a service that has any status or services that are passing", but since passing services are a subset of any service it was really "give me all services regardless of their status". This was especially confusing for new users and the documentation was a bit unclear. Consul Template now returns an error when you specify additional filters using "any":

```text
* service filter: cannot specify extra keys when using "any"
```

Hopefully this will help eliminate some of the confusion around the behavior of "any". I have additionally updated the documentation.

"passing" is also a special case that was improperly handled before. "passing" actually has a dual meaning - when we query Consul, we have the option to return only passing services. When we filter services on Consul Template's end, we also have the option to filter _only_ passing services. Furthermore, there is a fundamental difference between:

```liquid
{{ range services "nginx" "passing" }}
```

and

```liquid
{{ range services "nginx" "passing, unknown" }}
```

The former says "give me all nginx services which have all checks passing". The latter says "give me all nginx services which has at least one check that is passing or at least one check that is unknown". Previously, we did not differentiate between these two cases. This PR changes Consul Template to handle "passing" vs "passing, (other)" differently.

Finally, this PR adds an additional check to the status filter parsing to ignore empty statuses, for example:

```liquid
{{ range services "nginx" "unknown,   , critical" }}
```

This is useful if the user is building the filter from the result of a subquery where they are dynamically building the list of acceptable status filters.

- Fixes https://github.com/hashicorp/consul-template/issues/210

### DIFF
--- a/dependency/health_service.go
+++ b/dependency/health_service.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	"github.com/hashicorp/consul/api"
+	"github.com/hashicorp/go-multierror"
 )
 
 // Ripped from https://github.com/hashicorp/consul/blob/master/consul/structs/structs.go#L31
@@ -52,7 +53,7 @@ func (d *HealthServices) Fetch(client *api.Client, options *api.QueryOptions) (i
 	log.Printf("[DEBUG] (%s) querying Consul with %+v", d.Display(), options)
 
 	health := client.Health()
-	entries, qm, err := health.Service(d.Name, d.Tag, d.Status.onlyAllowPassing(), options)
+	entries, qm, err := health.Service(d.Name, d.Tag, d.Status.passingOnly(), options)
 	if err != nil {
 		return nil, qm, err
 	}
@@ -62,7 +63,7 @@ func (d *HealthServices) Fetch(client *api.Client, options *api.QueryOptions) (i
 	services := make([]*HealthService, 0, len(entries))
 
 	for _, entry := range entries {
-		if !d.Status.accept(entry.Checks) {
+		if !d.Status.Accept(entry.Checks) {
 			continue
 		}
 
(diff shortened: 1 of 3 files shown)


dependency/health_service.go
```diff
@@ -197,44 +190,107 @@ func ParseHealthServices(s ...string) (*HealthServices, error) {
...
+			continue
+		}
+
+		// Record the case where we have the "any" status - it will be used later.
+		if trim == HealthAny {
+			hasAny = true
+		}
+
+		// Validate that the service is actually a valid name.
+		if trim != HealthAny &&
```
ryanuber: Minor, but maybe use a `switch` here and default to appending the error?

ryanuber:
The code looks great! I think having the multiple meanings for `passing` is a little bit confusing, though. If this is a backwards-compat thing, then maybe this is the only way, but it would be nice if there was some distinction between like `passing` and `passing:all` or something like that.

Otherwise, this LGTM.

dependency/health_service.go
sethvargo: Oh yea, that's much cleaner :smile:

sethvargo:
@ryanuber yea, I agree. The distinction _does_ exist though, it's just really confusing (and even harder to explain). If you don't pass an argument to `service`, it defaults to "all checks passing". If you pass "passing" to `service`, it defaults to "any check passing". I'm going to update the docs to make that clearer.

tgwizard:
Oh, tricky bug! Does this mean that `"passing, warning"` will return services that have at least one check (e.g. a node check) that is passing or warning, but any number of critical checks? If so, how would you specify "give me the services that have only passing or warning checks, but not any critical checks"?

sethvargo:
@ryanuber @tgwizard okay, I've refactored this a bit to pre-calculate the service's status by iterating over all checks and assuming:

```text
passing < warning < unknown < critical
```

So a service that has checks that are critical will always be marked as critical, etc. I think this will have the desired end result.

tgwizard:
Cool!

So now these return the same services / conceptually does the same filtering, right?

```
{{service "webapp"}}
{{service "webapp" "passing"}}
```

Should the README be updated to reflect that?

sethvargo:
@tgwizard well, sorta. The [README](https://github.com/hashicorp/consul-template#service) is updated to say:

> There is an architectural difference between the following:
> 
> ```liquid
> {{service "webapp"}}
> {{service "webapp" "passing"}}
> ```
> The former will return all services which Consul considers "healthy" and passing. The latter will return all services registered with the Consul agent and perform client-side filtering. As a general rule, you should not use the "passing" argument alone if you want only healthy services - simply omit the second argument instead. However, the extra argument is useful if you want "passing or warning" services like:

They _should_ be the same, but the difference is in the amount of memory Consul Template is going to consume, the size of the response from Consul, and the amount of data Consul needs to watch, does that make sense?

tgwizard:
Ah, with that part in the README I think it is clear enough.

PR #241, opened 2015-03-29
Improve byKey function to return a map

### LINKED ISSUE #202: Index KV data in template
Say I had some extra per-node data in a KV tree like /weights/node_name.  I wish I could query this entire tree with a single watch:

```
{{$weights := tree "/weights"}}
```

And then lookup the per-node keys while looping over services:

```
{{range service "release.webapp"}}
    {{$weight := or (index $weights (print "/weights/" .Node)) 100}}
    server {{.Node}} {{.Address}}:{{.Port}} weight {{$weight}}{{end}}
```

It doesn't work because `tree` returns a slice, not a map: `error calling index: cannot index slice/array with type string`.

One idea is a `byKey` function that indexes the result of `tree` by they Key property of each entry:

```
{{$weights := tree "/weights" | byKey}}
```

Probably better would be a treeMap function that is similar to tree, but returns a map of subkeys under the lookup path.  This would make the map less clunky to index:

```
{{$weights := treeMap "/weights"}}
{{range service "release.webapp"}}
    {{$weight := or (index $weights .Node) 100}}
    server {{.Node}} {{.Address}}:{{.Port}} weight {{$weight}}{{end}}
```

### LINKED ISSUE #211: Race condition in byKey function
Originally reported https://github.com/hashicorp/consul-template/pull/209#issuecomment-76486361, there is a bug in the `byKey` function (or somewhere else in the runner) that is causing partial output to be rendered.

Use the following script with `GOMAXPROCS=1` and see that partial output is rendered:

```shell
#!/bin/sh
set -e

DATADIR=/tmp/consul-testing-1

echo STARTING CONSUL
consul agent -server -bootstrap -data-dir $DATADIR &
CONSUL_PID=$!
sleep 5

echo
echo CREATING KEYS
curl -XPUT localhost:8500/v1/kv/groups/elasticsearch/es1
curl -XPUT localhost:8500/v1/kv/groups/elasticsearch/es2
curl -XPUT localhost:8500/v1/kv/groups/elasticsearch/es3
curl -XPUT localhost:8500/v1/kv/services/elasticsearch/check_elasticsearch
curl -XPUT localhost:8500/v1/kv/services/elasticsearch/check_indexes
sleep 2

echo
echo STARTING CONSUL-TEMPLATE
consul-template \
  -template ./template.ctmpl:./out \
  -log-level debug &
CONSUL_TEMPLATE_PID=$!
sleep 2

echo
echo RESULT
cat out

kill -9 $CONSUL_PID
kill -9 $CONSUL_TEMPLATE_PID

rm -f out
rm -rf $DATADIR
```

sethvargo:
- Fixes #202 
- Fixes #211 

/cc @ryanuber @etuttle

### DIFF
--- a/template_functions.go
+++ b/template_functions.go
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	dep "github.com/hashicorp/consul-template/dependency"
-	"github.com/mitchellh/copystructure"
 )
 
 // now is function that represents the current time in UTC. This is here
@@ -254,8 +253,8 @@ func treeFunc(brain *Brain,
 //
 // Note that the top-most key is stripped from the Key value. Keys that have no
 // prefix after stripping are removed from the list.
-func byKey(pairs []*dep.KeyPair) (map[string][]*dep.KeyPair, error) {
-	m := make(map[string][]*dep.KeyPair)
+func byKey(pairs []*dep.KeyPair) (map[string]map[string]*dep.KeyPair, error) {
+	m := make(map[string]map[string]*dep.KeyPair)
 	for _, pair := range pairs {
 		parts := strings.Split(pair.Key, "/")
 		top := parts[0]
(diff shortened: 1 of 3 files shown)


template_functions.go
```diff
@@ -266,11 +266,21 @@ func byKey(pairs []*dep.KeyPair) (map[string][]*dep.KeyPair, error) {
 		}
 
 		if _, ok := m[top]; !ok {
-			m[top] = make([]*dep.KeyPair, 0, 1)
+			m[top] = make(map[string]*dep.KeyPair)
+		}
+
+		m[top][key] = &dep.KeyPair{
```
ryanuber: You could probably also:
```
newPair := *pair
newPair.Key = key
m[top][key] = &newPair
```
Up to you, both would work fine.

ryanuber:
Code-wise this LGTM. Since the data type is changing, does any backwards compatibility break? It looks like it might not, just double-checking. Otherwise :+1: :ship:

template_functions.go
sethvargo: Black magic! :smile:

sethvargo:
Go's templating language handles the map vs slice thing pretty well. Existing templates will continue to function since it will yield the pairs, but if you specify two variables to the range function it will also yield the key.

We also haven't released CT with `byKey` yet - it's only ever been on master. While I hate to introduce breaking changes, the only people using `byKey` are building from master.

PR #255, opened 2015-04-16
Add an --ssl-cert argument which will send a client certificate to the c...

danielrowles-wf:
...onsul

server, and an --ssl-ca-cert argument which will validate the consul server
certificate against a specific list of CAs.


sethvargo:
Hey @danielrowles-wf,

This is looking really good. Thank you for the PR. Would you mind updating the documentation in the README (there's a table that lists all the configuration options), and maybe add an example somewhere? Thanks!

danielrowles-wf:
@sethvargo I've added some simple documentation to the README file. I can add full documentation on setting up a CA, and enabling SSL in your consul server if you like, but I'm not sure that really belongs in this repo. Let me know if you'd like me to add that, though, and, if so, where you'd like it.

sethvargo:
Nope, this looks fine. Thanks!

PR #257, opened 2015-04-21
Fixed references to "webapp" and "nyc1"

andrewwatson:
I noticed while walking someone through the examples that they didn't work anymore


sethvargo:
Thanks! We updated the demo cluster awhile back :smile:

PR #273, opened 2015-05-07
added orkey for key precedence functionality

Ecsy:
This achieves key precedence. We needed to overwrite values based on our production configuration using 'or' would make templates hard to understand - think of nesting 3 or more keys with or, so we created a shorthand for that functionality.

### DIFF
--- a/template.go
+++ b/template.go
@@ -89,6 +89,7 @@ func funcMap(brain *Brain, used, missing map[string]dep.Dependency) template.Fun
 		"file":        fileFunc(brain, used, missing),
 		"key":         keyFunc(brain, used, missing),
 		"ls":          lsFunc(brain, used, missing),
+		"merge":       mergeFunc(brain, used, missing),
 		"nodes":       nodesFunc(brain, used, missing),
 		"service":     serviceFunc(brain, used, missing),
 		"services":    servicesFunc(brain, used, missing),
--- a/template_functions.go
+++ b/template_functions.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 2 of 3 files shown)


sethvargo:
Hey @Ecsy 

Thank you for the Pull Request. We talked about this internally and I do not think we want to support this functionality for a few reasons:

1. We prefer to stay within Go's templating language as much as possible. Although sometime's limiting, the new syntax adds an additional API function that users must be aware of and does not save that much additional typing or offer additional readability in our opinion. You are saving, on average, 5 characters per key, which is difficult to justify the tradeoff of the overhead of maintaining another API function that is easily achievable using the built-in Go template functions.

2. I think the `or` syntax makes it very clear that you are querying _two_ keys (i.e. creating two watches in Consul). The `orkey` functionality could mislead a user into thinking they are only creating one watch. We have issues with users DDOSing their Consul clusters because of misconceptions about how CT works, and I would prefer to avoid introducing more areas for confusion.

3. "Key precedence" is something that should probably be avoided when configuring infrastructure at scale. If you are trying to use Consul to mimic something like Puppet's Hiera or Chef's Ohai, you are going to experience degraded performance and your cluster will be very difficult to reason about. If you have so may layers of nesting that a single `or` query becomes too complex, that is a problem with the design decision, not a technical one. In other words, I would suggest rethinking the decision to use Consul's KV store in this manner.

Does that make sense?

pepov:
Hey all, what about enconsul's merged prefixes? Isn't it something that could be added here as well? I suppose it could be implemented using path watches and select the referenced keys only locally. We really need this, we have no choice but to support merging multiple levels.

Ecsy:
Hi all,

1. I can not disagree with you on this.

2. You should document this behaviour more precisely for the users. This could be another discussion how to solve it, but not in this topic.

3. After reviewing our consul-template usage patterns, "precedence" was not the right word for our use case. What we really need is a sort of `tree` with a merge capability so that way we could pass it to a `range` function. I will reference it as `merge` from now on. This would be the equivalent what envconsul does with prefixes. That way the number of watches required would be the number of keys we pass to this merge function. This could be well documented and it would reduce the watches used by a magnitude in our case.

Consul KV store contains the following structure:
```
path/default/a=1
path/default/b=2
path/override/b=4
path/override/c=3
```
With the following template:
```
{{$key1 := "path/default/"}}
{{$key2 := "path/override/"}}

{{range ls $key1}}
	{{.Key}}={{or (key (printf "%s/%s" $key2 .Key)) .Value}}
{{end}}

{{range ls $key2}}
	{{if ne (key (printf "%s/%s" $key1 .Key)) ""}}
		{{.Key}}={{.Value}}
	{{end}}
{{end}}
```
This could be much easier by doing it this way:
```
{{range merge $key1 $key2}}
	{{.Key}}={{.Value}}
{{end}}
```
The result would be:
```
a=1
b=4
c=3
```

I hope you got the idea.
Please share your thoughts about this.

Ecsy:
Hi @sethvargo,
Do you have any update on this?

sethvargo:
Hi @Ecsy 

I haven't had a chance to think through it yet. I'm working on some other things and will circle back to Consul Template later this week/next. Thanks!

Ecsy:
Hi @sethvargo 

Two weeks have passed. When do you have time to revisit this issue?
Thanks

armon:
@Ecsy After some further discussion internally, we are not ready to merge this. For one, every new template function adds more complexity, and we haven't often had this behavior requested. It also encourages this merging to be done client side (e.g. in consul-template) across many clients instead of being done ahead of time and written to Consul. Doing the flattening ahead of time spares the clients from having to run multiple key watches, and reduces the query load on the Consul servers.

Long term, I believe there are lots of cases where a small amount of custom logic in CT would be helpful, but it is not sustainable for us to keep adding new template functions. To that end, we are investigating a plugin model, or an embedded language, such that you could write a custom "merge" function like this.

Issue #275, opened 2015-05-08
Improve error reporting


c4milo:
I have an error in one of my templates it seems, but I can't really tell what's going on based on these reported errors: 

```
Consul Template returned errors:
runner: config: walk error: 1 error(s) occurred:

* Line 5, column 11: syntax errorConsul Template returned errors:
runner: config: walk error: 1 error(s) occurred:

* Line 5, column 11: syntax errorConsul Template returned errors:
runner: config: walk error: 1 error(s) occurred:

* Line 5, column 11: syntax errorConsul Template returned errors:
runner: config: walk error: 1 error(s) occurred:

* Line 5, column 11: syntax errorConsul Template returned errors:
runner: config: walk error: 1 error(s) occurred:

* Line 5, column 11: syntax errorConsul Template returned errors:
runner: config: walk error: 1 error(s) occurred:

* Line 5, column 11: syntax errorConsul Template returned errors:
runner: config: walk error: 1 error(s) occurred:

* Line 5, column 11: syntax errorConsul Template returned errors:
runner: config: walk error: 1 error(s) occurred:

* Line 5, column 11: syntax errorConsul Template returned errors:
runner: config: walk error: 1 error(s) occurred:

* Line 5, column 11: syntax errorConsul Template returned errors:
runner: config: walk error: 1 error(s) occurred:
```

This is my template: 

```
auto_accept: {{key "q/stg/salt/master/auto_accept"}}

file_roots:
  base:
    - {{key "q/stg/salt/master/file_roots/base"}}

fileserver_backend:
  - {{key "q/stg/salt/master/backend"}}

gitfs_remotes:
  - {{key "q/stg/salt/master/backends/git/remote"}}

gitfs_user: {{key "q/stg/salt/master/backends/git/user"}}
gitfs_password: {{key "q/stg/salt/master/backends/git/password"}}
gitfs_root: {{key "q/stg/salt/master/backends/git/root"}}
gitfs_base: {{key "q/stg/salt/master/backends/git/base"}}
gitfs_provider: {{key "q/stg/salt/master/backends/git/provider"}}

```

ryanuber:
Hey @c4milo, do you use configuration files to start consul-template? It looks like the error may be in one of the configuration files provided. I agree though, the error could be more helpful. Marking as an enhancement!

c4milo:
[...]

Issue #271, opened 2015-04-30
Vault section in config file being overwritten


ChrisAubuchon:
Using config:
```
consul = "127.0.0.1:8500"

vault {
        address = "https://localhost:8200"
        token = "<token>"
        ssl {
                enabled = true
                verify = false
        }
}

ssl {
        enabled = false
}
```

Running consul-template with:
```
consul-template -log-level=debug -config=<config_path>
```

Results in:
```
2015/04/30 18:58:28 [DEBUG] runner: final config (tokens suppressed):
{
  "path": "",
  "consul": "127.0.0.1:8500",
  "vault": {
    "address": "",
    "ssl": {
      "enabled": true,
      "verify": true,
      "cert": "",
      "ca_cert": ""
    }
  },
```

ChrisAubuchon:
The merge at runner.go:392 seems to be where the Vault configuration is overwritten.

mdelagralfo:
Yeah I think the problem is that anything defined in DefaultConfig() is overriding the contents of the config file.  For example I get the same result trying to set the SSL flag to true.  Using config:

```
ssl {
    enabled = true
}
```

Running consul-template with: 

```consul-template -config consul_template_config.hcl -log-level=debug -once```

Results in:

```
{
 ...
  "ssl": {
    "enabled": false,
    "verify": true,
    "cert": "",
    "ca_cert": ""
  },
  ...
```

To me, it looks like the defaults are treated as command-line arguments and take precedence over the config files.

sethvargo:
The problem is that Go's default boolean value is `false`. So when creating a new config, we cannot differentiate between 'the user explicitly set this to false' and 'the user did not set this'. This makes default values effectively trinary. I have a good solution, I just haven't had time to implement it yet :smile:

mdelagralfo:
Yeah I noticed the same challenges trying to get the CLI config right.  I may attempt a PR if I figure something out.  Right now it's impossible to get consul-template talking to Vault on another server without hacking the source.

sethvargo:
@ChrisAubuchon @mdelagralfo ^

kolis:
This is a pretty serious bug if I want to use the vault functionality, considering that the latest release version of consul-template does not include this fix. This means when I want to use consul-template I must build it myself, instead of just using the prebuild package.

When can we expect a new release package?

PR #278, opened 2015-05-10
Use a trinary config for booleans

sethvargo:
- Fixes https://github.com/hashicorp/consul-template/issues/271
- Fixes https://github.com/hashicorp/consul-template/issues/275

/cc @KFishner @ryanuber

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -156,13 +156,13 @@ func (cli *CLI) parseFlags(args []string) (*Config, bool, bool, bool, error) {
 	flags.StringVar(&config.Consul, "consul", config.Consul, "")
 	flags.StringVar(&config.Token, "token", config.Token, "")
 	flags.Var((*authConfigVar)(config.Auth), "auth", "")
-	flags.BoolVar(&sslEnabled, "ssl", true, "")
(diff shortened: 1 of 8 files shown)


mdelagralfo:
I gave it a quick test, and the Vault settings from a config file seem to be respected now.  Great!

cli.go
```diff
@@ -173,7 +176,21 @@ func (cli *CLI) parseFlags(args []string) (*Config, bool, bool, bool, error) {
 
 	// Deprecated options
 	var deprecatedSSLNoVerify bool
-	flags.BoolVar(&deprecatedSSLNoVerify, "ssl-no-verify", !config.SSL.Verify, "")
+	flags.BoolVar(&deprecatedSSLNoVerify, "ssl-no-verify", false, "")
+
+	// Because Go's default value for booleans are false, we need to track
+	// auxillary values to see if they were actually provided. It has a pretty
+	// high degree of sadness.
```
ryanuber: Hm, this comment definitely describes the following block :(

ryanuber: @sethvargo can't we just use the `true` default value for the flag, and then merge the other configs against the CLI config?

PR #282, opened 2015-05-12
Add split/join functions to template

zemm:
This adds 4 standard functions to be used at template: `split`, `splitWith`, `join` and `joinWith`.
`split` and `join` are directly exposed `strings.Split` and `strings.Join`, but since it seems they cannot be intuitively piped, I also added local versions `splitWith` and `joinWith` with argument order reversed.

I will add updated documentation later, possibly waiting for an input first on weather this (or the two versions) are ok, bad idea, can be done with single functions (I'm new to Go) or some other concerns.

### Motivation
We store multiline values at some keys, and need to print the lines at the template with pre- or postfixes. Join was more of a complementary addition.

#### `split`
```liquid
{{range split (key "foo/options") "\n"}}
option {{.}};{{end}}
```
Results:
```text
option foo;
option bar;
```

#### `splitWith`
```liquid
{{range key "foo/options" | splitWith "\n"}}
option {{.}};{{end}}
```
Results:
```text
option foo;
option bar;
```

#### `join`
```liquid
{{join $items ";"}}
```
Results:
```text
foo;bar
```

### `joinWith`
```liquid
{{$items | joinWith ";"}}
```
Results:
```text
foo;bar
```

### DIFF
--- a/template.go
+++ b/template.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"path/filepath"
+	"strings"
 	"text/template"
 
 	dep "github.com/hashicorp/consul-template/dependency"
@@ -108,5 +109,9 @@ func funcMap(brain *Brain, used, missing map[string]dep.Dependency) template.Fun
 		"toLower":         toLower,
 		"toTitle":         toTitle,
 		"toUpper":         toUpper,
+		"split":           strings.Split,
+		"splitWith":       splitWith,
+		"join":            strings.Join,
+		"joinWith":        joinWith,
 	}
 }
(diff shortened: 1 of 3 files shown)


sethvargo:
Hi @zemm 

This looks really good, but I think I am :-1: on including the `*with` functions. I think the API should probably look like this:

```liquid
{{range key "foo/options" | split "\n"}}{{end}}
{{printf $items | join "\n"}}
```

sethvargo:
Merged in 873014e

zemm:
Oh, I was just about to remove the standard ones, and rename the `With`-versions :) The readme update was also missing. I'll still push the readme update, but should I still do the removal+renaming or not?

sethvargo:
Hey @zemm 

I did all that in a follow-up commit to yours: https://github.com/hashicorp/consul-template#split. It's all good! :)

zemm:
Oh I see. Thanks! Gotta learn to read.

PR #298, opened 2015-05-21
add fqdnRandservice

maguec:
This allows a user to hash the hostname to semi-randomize the elements chosen from a slice akin to https://docs.puppetlabs.com/references/latest/function.html#fqdnrand

It also allows truncation.

Example use case is in resolv.conf. I only want to select 2 of the services running bind and set 8.8.8.8 as the default nameserver.

```
{{range datacenters}}search {{.}}.node.consul service.{{.}}.consul {{end}}
{{ range service "bind" | fqdnRandService 2 }}
nameserver {{.Address}}
{{end}}
nameserver 8.8.8.8
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,4 +1,4 @@
-package main // import "github.com/hashicorp/consul-template"
+package main 
 
 import (
 	"os"
--- a/main.go
+++ b/main.go
@@ -1,4 +1,4 @@
-package main 
+package main // import "github.com/hashicorp/consul-template"
 
 import (
 	"os"
--- a/template.go
+++ b/template.go
@@ -102,6 +102,7 @@ func funcMap(brain *Brain, used, missing map[string]dep.Dependency) template.Fun
 		"loop":            loop,
 		"join":            join,
 		"parseJSON":       parseJSON,
+		"fqdnRandService": fqdnRandService,
 		"regexReplaceAll": regexReplaceAll,
 		"regexMatch":      regexMatch,
 		"replaceAll":      replaceAll,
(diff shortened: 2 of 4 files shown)


maguec:
This appears to be an AWS error ../vault/physical/s3.go:104: undefined: aws.Error

armon:
@sethvargo We should figure out why there is a dependency on the `vault` package and remove that, otherwise it will bloat the CT binary

sethvargo:
@armon CT uses the Vault API client to talk to Vault. I bet there is something in the vault api client that is pulling in something from Vault core then :/ /cc @mitchellh

sethvargo:
@armon determined that it's being used in tests (only), not the final binary.

sethvargo:
Hi @maguec 

Thank you for the Pull Request. Because this has not been a widely-request function and in order to limit the scope of Consul Template, I am going to request you implement this as a [plugin](https://github.com/hashicorp/consul-template#authoring-plugins) instead of adding this to Consul Template core.

Please let me know if you have any questions, and I am sorry that we will not be merging this.

123BLiN:
Sorry to bring this old post up but we are interested in such feature!
@maguec did you happened to make a plugin for this?
In our case we need to consistently choose only one node from all service nodes to forward traffic based on service name - this should give us some kind of load balance for a bunch of services that are not supporting traffic balancing - but we want failover.

maguec:
@123BLiN I did not rewrite as a plugin

With newer versions of consul using a prepared query with DNS seems to work for some cases
https://www.hashicorp.com/blog/consul-0-6/#prepared-queries

PR #305, opened 2015-06-04
Add LookupIP, LookupIPv4, and LookupIPv6 functions

williambailey:
These functions are handy in the following situations when having to work with external services.

1. You have something like haproxy which you are using to forward connections to an external service. Being able to lookup the available ip addresses of that service allows you to load balance your outgoing connections across all of the ip addresses available. 
2. When you have an external service that provides DNS records with low TTLs and you want to reconfigure/bounce an application should the DNS update.
3. Both of the above.


tgwizard:
:+1: on this, we have the same use-case.

dependency/lookup_ip.go
```diff
@@ -0,0 +1,168 @@
...
+					return true
+				}
+			}
+			return false
+		}(d, ip)
+
+		if changed {
+			return ip, nil
+		} else {
+			time.Sleep(time.Minute)
```
tgwizard: Perhaps the sleep time here should be related to the TTL of the DNS records?

sethvargo:
Hi @williambailey 

Thank you for the Pull Request. We will not be merging this into Consul Template core, but please let me explain further.

In these situations where a resource "behaves like" a dependency with global data but is not part of Consul's core, we recommend making this it's own utility that writes data to Consul. In this specific case, your lookup functions would run as a service on a node, writing data into Consul. Your templates then watch that key that is being written.

Please let me know if you have any questions.

williambailey:
Thanks for the reply @sethvargo. As stated in the PR my main use case for these functions were to make working with external services, themselves registered in consul, a bit nicer in the situation where those services use DNS with TTLs as a kind of poor mans failover/load balancing mechanism. Having to use some utility app that queries the external service from consul, before resolving the IP addresses and then updating the KV store only to then have any template watch those KVs, and potentially also the external service, seems quite a bit more awkward to get up and running... however it's certainly doable.

Looking forward to the next release.

PR #306, opened 2015-06-06
Add node function.

williambailey:
Query Consul for a single node in the catalog. Essentially this exposes the `/v1/catalog/node/<node>` API to templates.


williambailey:
Not sure why that travis build failed. Going to close and reopen to get Travis to rebuild.

Issue #258, opened 2015-04-21
Can't create yaml templates from deeper directories of kv


onnimonni:
Hey!

This might be possible but I couldn't find anything from documentation or source code. I tried to use ```func byKey``` in inner loops but without success.

**What I would like to achieve:**
Generate >=4 levels deep yaml from consul KV.

For example how to turn
**KEY** = **VALUE**
- domains/example.com/container = 12345
- domains/example.com/ssl/key = example.com.key
- domains/example.com/ssl/crt = example.com.crt
- domains/another.com/container = 123456

**to yaml:**
```yaml
domains:
  example.com:
    container: 12345
    ssl:
      key: example.com.key
      crt: example.com.crt
  another.xyz:
    container: 123456
```

**What I tried to do:**
```go
{{range $key, $pairs := tree "domains" | byKey}}{{$key}}:
{{range $pair := $pairs}}  {{.Key}}:{{.Value}}
{{end}}{{end}}
```

**What I get instead:**
```yaml
domains:
  example.com:
    container: 12345
      ssl/key: example.com.key # <============= Should be ssl: key: without '/'
      ssl/crt: example.com.crt # <============= Should be ssl: crt: without '/'
  another.xyz:
    container: 123456
```

sethvargo:
This looks like a duplicate of https://github.com/hashicorp/consul-template/issues/243. In short, I do not think it is possible due to Go's templating language wanting a pre-defined structure, but I have not had a chance to really dig in and try it.

onnimonni:
Well The tree func could be remodeled to allow recursion or inner loops. I think yaml is pretty standard and it would be awesome to have to have tree_to_json and tree_to_yaml methods available into consul-template

sethvargo:
@onnimonni I don't disagree :smile:, but in my experience, Go's templating language expects the map to be balanced, so even if I returned an arbitrarily-nested data structure, it would crumble when the template tries to process it. As I said - I'm still trying out different approaches, but I just wanted to clarify that it's not a trivial issue to circumvent :smile:

onnimonni:
Thanks in advance :)

pdf:
I quite like Go, but every time I look at Template I want to stab my face off.  I probably would have looked at using something like [pongo2](https://github.com/flosch/pongo2), since it's much more flexible, but that's obviously problematic at this stage because people are already using this project and effort has been spent trying to hammer text/template into the hole. :anguished:

PR #311, opened 2015-06-08
Add serializers for outputting JSON and YAML from KV

sethvargo:
Fixes https://github.com/hashicorp/consul-template/issues/258

/cc @ryanuber


onnimonni:
Thanks man!

sean-:
:+1: Ansible users of the world thank you (once the next version is cut *grin*).

panda87:
@sean- Can you pls explain how this helps to Ansible users? how can we replace or use consul template k\v variables into the .j2 file? what should I add?

Thanks

PR #313, opened 2015-06-09
Add plugin system

sethvargo:
This adds a basic plugin system based around string message-passing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -677,7 +677,7 @@ See Go's [time.Format()](http://golang.org/pkg/time/#Time.Format) for more infor
 Takes the result from a `tree` or `ls` call and converts it into a JSON object.
 
 ```liquid
-{{ tree "config" | toJSON }} // e.g. {"admin":{"port":1234},"maxconns":5,"minconns":2}
+{{ tree "config" | explode | toJSON }} // e.g. {"admin":{"port":1234},"maxconns":5,"minconns":2}
 ```
 
 Note: This functionality should be considered final. If you need to manipulate keys, combine values, or perform mutations, that should be done _outside_ of Consul. In order to keep the API scope limited, we likely will not accept Pull Requests that focus on customizing the `toJSON` functionality.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -743,6 +758,72 @@ minconns: 2
...
+- `INPUT` - input from the template - this wil always be JSON if provided
+
+#### Important Notes
+
+- Plugin output must be returned as a JSON string (even if it is empty)
+- Only stdout will be parsed for output - be sure to log all errors, debugging
+  messages onto stderr to avoid errors when Consul Template tries to parse the
+  JSON
+- Always `exit 0` or Consul Template will assume the plugin failed to execute
+
```
ryanuber: Maybe we should add a note somewhere in here about security implications, something like:
```
- Plugins execute user-provided scripts and pass in potentially sensitive data from Consul.
  Neither is validated or protected by consul-template, so all necessary precautions and
  considerations should be made.
```

williambailey:
Nice addition. :sunglasses:  Quick question. Do you have any thoughts about long running plugins vs. fire and forget plugins? It might be nice to have the option to have a plugin invoked and then processed as a dependency where we can read multiple JSON outputs (using the encoding/json package func (*Decoder) Decode) from STDOUT to trigger template cycles.

ryanuber:
@sethvargo this looks good. One thought is, should we let the plugins just return a string? My thinking is that we have this `parseJSON` helper that can be a part of the pipeline, and it would help simple cases that don't need to return a map-like structure. We allow any string as input, so it might make sense for the output to be the same. What do you think?

sethvargo:
@ryanuber good point. Updated to just be a string

sethvargo:
@williambailey the plugin model will only support the fire-and-forget model. I will elaborate more before the final release, but the "plugin as a dependency" model will not be supported in Consul Template core. It is a rabbit hole and goes against the principles of Consul. 

If you have a plugin that periodically does something, that plugin should write data _to Consul_ and the template should just watch that key for changes.

README.md
```diff
@@ -743,6 +758,75 @@ minconns: 2
...
+$ NAME [INPUT...]
+```
+
+- `NAME` - the name of the plugin - this is also the name of the binary
+- `INPUT` - input from the template - this wil always be JSON if provided
+
+#### Important Notes
+
+- Plugins execute user-provided scripts and pass in potentially sensitive data
+  from Consul or Vault. Nothing is validated or protected by Consul Remplate,
```
ryanuber: I guess the OCD filter didn't catch this one :) "Consul Remplate"

ryanuber:
@sethvargo LGTM

PR #320, opened 2015-06-16
Make Runner callable from outside code. Add the option to start and stop services

marouenj:
We'd like to share the following propositions:
- Currently, CT Runner's lifecycle runs inside the Cli container. It would be nice if it is callable from anywhere, easy to plug into other systems. One good example is to be able to control Runner from within a consul agent.
- Although CT offers the possiblity to run any command upon a new rendering of the template, when it comes to changing the state of the service, it only make sense to restart it. Because, first, successively starting/stopping a service is irrelevant. Besides, passing a custom script (that knows which command to execute) is not straighforward to implement and is service-specific and thus requires more effort from the service owner.

In this PR:
We moved 'brain', 'config', 'runner' and 'template' to a new package called 'core'. Runner can thus be imported using

    consultemplate "github.com/marouenj/consul-template/core"

We changed the structure of the template config from

    type ConfigTemplate struct {
        Source      string `json:"source" mapstructure:"source"`
        Destination string `json:"destination" mapstructure:"destination"`
        Command     string `json:"command,omitempty" mapstructure:"command"`
    }

to

    type ConfigTemplate struct {
        Source         string `json:"source" mapstructure:"source"`
        Destination    string `json:"destination" mapstructure:"destination"`
        StartCommand   string `json:"startcommand" mapstructure:"startcommand"`
        RestartCommand string `json:"restartcommand" mapstructure:"restartcommand"`
        StopCommand    string `json:"stopcommand" mapstructure:"stopcommand"`
        First          bool
    }

Cli's instantiation of the runner is intentionally kept the same to bring as few changes as possible in this PR. When executing consul-template the usual way, the command will go into 'RestartCommand' instead of 'Command'.

But the Runner, when called from outside and passed a full config template, can
- start a service on the first rendering of the template
- restart it on subsequent ones
- stop the service
- enter the cycle again

If the PR is accepted, the Cli can be changed as well to make use of the new feature.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ build:
 	go build -o bin/$(NAME)
 
 test: deps
-	go list ./... | xargs -n1 go test -timeout=5s
+	go list ./... | xargs -n1 go test -timeout=25s
 	go list ./... | xargs -n1 go vet
 
 xcompile: deps test
--- a/brain.go
+++ b/brain.go
@@ -3,7 +3,7 @@ package main
 import (
 	"sync"
 
-	dep "github.com/hashicorp/consul-template/dependency"
+	dep "github.com/marouenj/consul-template/dependency"
 )
 
 // Brain is what Template uses to determine the values that are
--- a/brain_test.go
+++ b/brain_test.go
@@ -4,7 +4,7 @@ import (
 	"reflect"
 	"testing"
 
-	dep "github.com/hashicorp/consul-template/dependency"
+	dep "github.com/marouenj/consul-template/dependency"
 )
 
 func TestNewBrain(t *testing.T) {
--- a/cli.go
+++ b/cli.go
@@ -10,9 +10,9 @@ import (
 	"sync"
 	"syscall"
 
-	"github.com/marouenj/consul-template/logging"
-	"github.com/marouenj/consul-template/core"
-	"github.com/marouenj/consul-template/watch"
+	"github.com/hashicorp/consul-template/logging"
+	"github.com/hashicorp/consul-template/core"
+	"github.com/hashicorp/consul-template/watch"
 )
 
 // Exit codes are int values that represent an exit code for a particular error.
--- a/cli.go
+++ b/cli.go
@@ -11,6 +11,7 @@ import (
 	"syscall"
 
 	"github.com/marouenj/consul-template/logging"
+	"github.com/marouenj/consul-template/core"
 	"github.com/marouenj/consul-template/watch"
 )
 
@@ -82,7 +83,7 @@ func (cli *CLI) Run(args []string) int {
 	}
 
 	// Initial runner
-	runner, err := NewRunner(config, dry, once)
+	runner, err := core.NewRunner(config, dry, once)
 	if err != nil {
 		return cli.handleError(err, ExitCodeRunnerError)
 	}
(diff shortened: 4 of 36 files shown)


sethvargo:
Hi @marouenj 

Thank you for the Pull Request. Consul Template is not meant to be a process supervisor, but rather work closely alongside a process supervisor using the script command. It is outside of the scope of Consul Template to have full process-lifecycle management, since that is best delegated to something like Upstart, runit, etc. The start/stop/restart pattern does not align with the use cases for Consul Template.

That being said, I am interested in hearing about your use case a bit more. Maybe we can reach a happy compromise :smile:!

marouenj:
Seth,

Thank you for your reply.

Let me elaborate more.

This PR is not about making CT a process supervisor. It still works closely alongside a process supervisor to start/stop/restart a service.

With this PR, we bring synchronization between CT and the service: They wake up together. They exit together (provided it is agreed that the service be managed 'through' CT). I think this sounds as reasonable as: manage a db through a web interface rather than use raw sql

This way, the following unwanted scenario (maybe more exist) ceases to exist:
- the service is turned down (intentionally)
- the respective CT is still listening (left unintentionally)
- a dependency is updated (could be a human action. Could also be microservices talking to each other through consul. Both common)
- CT will run the command against our will (the command could be an upstart restart command, which would start the service against our will)

Besides, the changes in this PR won't affect the previous use cases of CT. To prove that, using this PR's version to run the usual command will just work as usual. In other terms, The Cli's behaviour hasn't been altered yet. Only the core (Runner precisely) changed slightly (reallocating channels, restarting watchers, ...)

Of course the CLi can be changed to offer access to the new features of Runner. It hasn't been done so because it's not the ultimate purpose. And here I come to explain my use case. Please find it [here](https://github.com/hashicorp/consul/pull/1039) in the form of a PR to consul.

This remains a personal point of view and, honestly, my use case still leaves in virtual box. But managing a service through consul (itself using the core of CT) makes life easier. It enforces a convention and brings synchronization to a set of previously decoupled actions, namely:
- registering/deregistering the service
- running/stopping it
- watching or not for config updates

And don't worry about any compromise. No need to. At worst, it has been the best initiation into go (by digging into clean code)

Best,
Marouen

sethvargo:
Hi @marouenj 

If you do not want Consul Template to make changes until you are ready, you can use the `-once` mode which will only render the template an execute the command one time. It will not watch Consul for changes. When you want to pick up new changes, you just re-run Consul Template with the `-once` flag to pick up the most recent configuration.

I am aware that the CLI API was not altered. I just do not think this is something Consul Template should be concerning itself with. Consul Template does not have a notion of a "service", which I where I think the confusion lies. The "command" attribute is intentionally arbitrary so that users can execute _anything_ when a template changes. As such, you could change a key in Consul that would send emails, reboot a machine, or more. Of course, one of those commands could be a service restart or reload, but that is not an explicit requirement for Consul Template. There is, in fact, an intentional decoupling from a service-oriented model. Users can choose how they want to start/restart/reload/stop services using any script they so choose. This also reduces the amount of complexity that Consul Template needs to deal with.

If you would like to use the core of Consul Template (specifically the `dependency` package), you can still do so. We use Consul Template's `dependency` package in envconsul, consul-replicate, and even some internal projects. It is the core for many community projects as well. It seems like you are fairly comfortable with writing Go code form your Pull Request, so I would recommend writing a tool that is specific to your use case :smile:.

Please let me know if you have any questions.

PR #325, opened 2015-06-19
Fix node service list sort when multiple services have the same name.

williambailey:
Change how we sort `NodeService` items in a `NodeServiceList`. This should allow multiple services with the same name to have a consistent ordering based on the unique service `ID`.

Authored using the github web UI and I haven't had a change to build locally. Apologies in advance if this breaks things.

### DIFF
--- a/dependency/catalog_node.go
+++ b/dependency/catalog_node.go
@@ -14,6 +14,7 @@ type NodeDetail struct {
 }
 
 type NodeService struct {
+	ID string
 	Service string
 	Tags    ServiceTags
 	Port    int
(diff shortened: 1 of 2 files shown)


dependency/catalog_node.go
```diff
@@ -152,5 +154,5 @@ type NodeServiceList []*NodeService
 func (s NodeServiceList) Len() int      { return len(s) }
 func (s NodeServiceList) Swap(i, j int) { s[i], s[j] = s[j], s[i] }
 func (s NodeServiceList) Less(i, j int) bool {
-	return s[i].Service <= s[j].Service
```
sethvargo: I think we might want to order by service first, then fallback to ID. What do you think?

williambailey: Seeing as ID defaults to Service if not provided when said service was registered I don't think that ordering by service then falling back to ID gives us anything as, for the most part, it will be the same value. However i'm not too fussed either way as long as the the ordering is the same between template renderings. :grin:

sethvargo: @williambailey I'm mostly worried about backwards-compatability. I don't want template rendering (order) to change if a user updates Consul Template as that would be unexpected. Can we leave service in there for now and maybe we can remove it for the 1.0 release.

williambailey:
@sethvargo - This should good. Any other comments?

PR #332, opened 2015-07-07
Reload configuration on SIGHUP

sethvargo:
/cc @phinze


phinze:
LGTM, though the build is unhappy

Issue #263, opened 2015-04-26
var increment within a range 


mrh666:
Hi there, 

I can't find out how to increment variable within a range. And how to suppress last ","?
Let's say I need to have an output within a config:
1:consul_node_1,2:consul_node_2,3:consul_node_3

How I can do it? This one doesn't work, just for example: {{range service "consul@dc1"}}{{$somewar}}:{{.Address}},{{$somevar += 1}}{{end}} ?

sethvargo:
Hi @mrh666 

Thank you for opening an issue. You will need to do something like the following:

```liquid
{{range $index, $service := service "consul@dc1" }}
{{if ne $index 0}},{{end}}{{.Address}}
{{end}}
```

Does that help?

mrh666:
sethvargo, Thank you, I've got an correct output, without last comma from:

```
{{range $index, $service := service "consul@dc1" }}{{if ne $index 0}}{{$index}},{{end}}:{{.Address}}{{end}}

:172.17.0.11,:172.17.0.22,:172.17.0.3
```
But how can I add a an incremental var there to have an output like:

```
1:172.17.0.11,2:172.17.0.22,3:172.17.0.3
```
I've tried to use different go syntax, but no luck. May be you can provide me some link to detailed docs for templates or examples?

sethvargo:
Hi @mrh666 

The documentation for text/template can be found here: http://golang.org/pkg/text/template/. I do not think it is possible to do what you are trying to do - that is just a limitation of Go's templating language.

sielaq:
Actually there is add function in go template 
you can see it in tests:
http://golang.org/src/text/template/exec_test.go?h=add#L355
and in html/template examples here:
http://gohugo.io/templates/go-templates/

but probably its missing in text/templates so consul-template cannot inherit it.

sethvargo:
@sielaq hugo uses it's own template extensions. I've been thinking of writing a verbose template plugin that includes many of the functions of CT, plus things like `add`, `subtract`, etc.

mrh666:
Anyway, using ```{{$index}}``` I have a desired output (starting from 0):

```
servers-spec={{range $index, $service := service "zookeeper"}}{{if ne $index 0}},{{end}}{{$index}}\:{{.Address}}{{end}}
```

```
servers-spec=0\:zookeeper-node-1,1\:zookeeper-node-2,2\:zookeeper-node-3
```

digitalronin:
[...]

PR #360, opened 2015-08-13
add and subtract int64 helpers

sheldonkwok:
I found these useful and I noticed someone mentioned wanting add/subtract functionality in https://github.com/hashicorp/consul-template/issues/263


sethvargo:
Hi @sheldonkwok 

Thank you for the PR. I need to think about this a bit more. I'm really hesitant to add these types of functions because, in my experience, if you need this level of complexity it's a sign of a larger problem. I'm going to triage the Consul tooling next week :smile:

ChristianKniep:
:+1: I love incrementing - just saying...

Issue #359, opened 2015-08-12
Support for renewing vault token


AndrewKlopper:
As far as I can tell, consult-template does not have support for renewing the vault token used to retrieve secrets. Since all tokens except those using the 'root' policy have a maximum lease time of 30 days, this means that the vault token used by consul-template will eventually expire and require a restart of the consul-template process. This in turn will result in losing all in-memory secret leases held by consul-template, meaning all config files managed by consul-template that contain dynamic secrets will have to be regenerated, and the associated processes restarted.

Am I missing something, or is there a way around this to avoid unnecessary interruptions in service?

sethvargo:
Hi @AndrewKlopper 

CT automatically renews _secrets_ requested from Vault, but it does not renew the token used to communicate. This is a bug. Tagged. Thanks! :smile:

PR #367, opened 2015-08-19
Add a way to renew Vault tokens

sethvargo:
This is a configuration option that defaults to true. Users can opt-out by setting the configuration option.

### DIFF
--- a/dependency/vault_token.go
+++ b/dependency/vault_token.go
@@ -1,0 +1,84 @@
+package dependency
+
+import (
+	"fmt"
+	"log"
+	"sync"
+	"time"
+)
+
+// VaultToken is the dependency to Vault for a secret
+type VaultToken struct {
+	sync.Mutex
+
+	leaseID       string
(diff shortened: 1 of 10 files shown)


sethvargo:
Fixes https://github.com/hashicorp/consul-template/issues/359

dependency/vault_token.go
```diff
@@ -0,0 +1,84 @@
...
+	// Create our cloned secret
+	secret := &Secret{
+		LeaseID:       token.LeaseID,
+		LeaseDuration: token.LeaseDuration,
+		Renewable:     token.Renewable,
+		Data:          token.Data,
+	}
+
+	leaseDuration := secret.LeaseDuration
+	if leaseDuration == 0 {
```
armon: We can probably just cancel the renew at this point right?

blalor:
What happens if consul-template is unable to renew the lease?  Does it exit?

Issue #375, opened 2015-09-01
function "contains" not defined


sirlatrom:
The docs presently specify a function ["contains"](https://github.com/hashicorp/consul-template/blob/6d91335c188c80611b29af1d14b115c00ef706e5/README.md#contains) with the following example given:

**contains-test.hctmpl:**

    {{range service "consul"}}{{ if .Tags | contains "production" }}
    # ...
    {{ end }}{{end}}

However, when used in a template, it yields the following output:

    $ consul-template -dry -once -template contains-test.ctmpl
    2015/09/01 08:27:36 [DEBUG] (logging) setting up logging
    2015/09/01 08:27:36 [DEBUG] (logging) config:

    {
      "name": "consul-template",
      "level": "WARN",
      "syslog": false,
      "syslog_facility": "LOCAL0"
    }

    Consul Template returned errors:
    template: template: contains-test.ctmpl:1: function "contains" not defined

The [changelog for the v0.9.0 release](https://github.com/hashicorp/consul-template/releases/tag/v0.9.0), which mentions a "contains" function, speaks of it as a `.Contains` function specifically on `ServiceTags`, which indeed works.

**ServiceTags.Contains-test.hctmpl:**

    {{range service "consul"}}{{ if .Tags.Contains "production" }}
    # ...
    {{ end }}{{end}}

Given the above template, consul-template returns no errors:

    consul-template -dry -once -template ServiceTags.Contains-test.ctmpl -consul lspdocker001:8500
    2015/09/01 08:37:25 [DEBUG] (logging) setting up logging
    2015/09/01 08:37:25 [DEBUG] (logging) config:

    {
      "name": "consul-template",
      "level": "WARN",
      "syslog": false,
      "syslog_facility": "LOCAL0"
    }

PR #376, opened 2015-09-01
Correct README re: contains function

sirlatrom:
Presently, there is no "contains" function - rather, there is a ServiceTags.Contains function.

This PR corrects the docs, albeit in a manner that makes the heading inconsistent the wording of the surrounding text (the other functions are all 'global' and have single-word, lower-case names).

See #375.

### DIFF
--- a/README.md
+++ b/README.md
@@ -483,13 +483,13 @@ Takes the list of services returned by the [`service`](#service) or [`services`]
 {{end}}{{end}}
 ```
 
-##### `contains`
-Determines if a needle is within an iterable element.
+##### `.Tags.Contains`
+Determines if a needle is within an service's tags.
 
 ```liquid
-{{ if .Tags | contains "production" }}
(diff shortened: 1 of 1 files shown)


sethvargo:
There is a contains function, but it hasn't been released yet: https://github.com/hashicorp/consul-template/blob/master/CHANGELOG.md#v0100dev-unreleased. 

Thank you for reporting! :smile:

sirlatrom:
I guess that's a lesson to me to read the [readme from the release tag](https://github.com/hashicorp/consul-template/blob/14cf110504f3089a3c96a36126bd6d5fe39df37a/README.md) rather than the front-page. Or to build from source :-)

sethvargo:
@sirlatrom hehe. I try to call this out really big at the top of the README, but sometimes people scroll right past because of force of habit. We will be releasing a new version soon though :smile:

PR #384, opened 2015-09-03
Fix race condition in cli_test by using io.Pipe instead of a byte buffer

jefferai:


sethvargo:
https://github.com/hashicorp/consul-template/commit/2e2aa83639cc31d8436673eacb59092819270dcf

PR #394, opened 2015-09-10
Cleanly handle config reread

### LINKED ISSUE #393: HUP doesn't properly reread the config
HUP rereads the config file but merges it with the existing config.  If a template has been removed, it will still be in the config and consul-template will crash with:

    2015/09/10 16:46:22 [INFO] (runner) creating Watcher
    Consul Template returned errors:
    open /tmp/ct/hector/agg/product_group_auto-3.properties: no such file or directory

Working on a patch.

wwalker:
Please apply this to master, to fix #393

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -64,6 +64,8 @@ func (cli *CLI) Run(args []string) int {
 	}
 
 	// Setup the config and logging
+	// Save original config (defaults + parsed flags) for handling HUP
+	original_config := config
 	config, err = cli.setup(config)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


wwalker:
This patch fixes the problem described below:

Given a config file, passed on the command line:
    
    template {
      source = "original-template"
      destination = "original-target"
    }

If we then update the config file:
  
    template { 
      source = "different-template"
      destination = "different-target"
    }

Now, we HUP the running consul template, rather than having one template in the new config, we now have both.

PR #410, opened 2015-09-23
Adds a toMap function which converts a ServiceTags element into a map.

etlweather:
This is needed to use the [Registrator](http://gliderlabs.com/registrator) / Consul-Template tools in order to dynamically configure Apache. Since Consul does not support Registrator's Attributes, and can only work with labels, a method of having key-values is necessary. By creating `key=value` labels, one can achieve this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -749,6 +749,19 @@ Takes the argument as a string and converts it to lowercase.
 
 See Go's [strings.ToLower()](http://golang.org/pkg/strings/#ToLower) for more information.
 
+##### 'toMap'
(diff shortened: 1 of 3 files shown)


sethvargo:
Hey @etlweather 

Thank you for the Pull Request. I understand this is a real problem you are facing, but I am against adding this kind of function to Consul Template. Instead, we are going to work to add this kind of metadata support as a first-class citizen in Consul itself. Please track https://github.com/hashicorp/consul/issues/1107 for that work. Thank you! :smile:

Issue #368, opened 2015-08-20
Checking for Consul k/v keys produces log spam


JayH5:
A common pattern is to check for the presence of a key and if it is not present, use a default value.
```
{{if key "service/app/max_conns"}}
  max_conns={{key "service/app/max_conns"}}
{{else}}
  max_conns=500
{{end}}
```

But doing this produces a constant stream of log messages:
```
2015/08/20 11:27:56 [WARN] ("key(service/app/max_conns)") Consul returned no data (does the path exist?)
2015/08/20 11:28:58 [WARN] ("key(service/app/max_conns)") Consul returned no data (does the path exist?)
...
```

Should this really be a `WARN` message? It seems like `DEBUG` is more suited.

sethvargo:
Hi @JayH5 

This is working as intended. If you search through old issues (I don't have the number off hand), users wanted a warning because of typographical errors. 

Also note, I believe you could refactor this as:

```liquid
max_conns={{or (key "service/app/max_conns") 500}}
```

That won't make the warning go away though.

It's challenging to find a balance to suit everyone's needs, but I think warning for a non-existent key is the correct UX, even if it's chatty. What do you think?

JayH5:
Thanks for the reply.

I did have a search through the issues before creating this one but didn't manage to find that issue.

I still think `DEBUG` is more suited to this case... `WARN` to me signifies something like a "recoverable error", not someone using a function in one of the ways it was intended (but not the *main* way).

Otoh, I guess people hardly ever think to set the log level as low as debug so I suppose this is reasonable...

sean-:
If a key doesn't exist, can we suppress future notifications for N seconds that way some folk could set it to 86400 and only get this message once a day?  It's not the space that's the concern, it's the signal-to-noise ratio that's important.

sethvargo:
@JayH5 @sean- what if we move it to an info level - then you can just run with log-level warn if you don't care

JayH5:
@sethvargo Moving from `WARN` to `INFO` would not make much difference to me. I wouldn't switch the log level to `WARN` as then we would lose basically any messages that would let us know consul-template is working -- we'd only get messages when something goes wrong.

I'm still pro making it `DEBUG` but I wouldn't lose much sleep over it if that didn't happen :sweat_smile:

sethvargo:
[...]

PR #417, opened 2015-10-08
Add key_or_default

sethvargo:
This new function builds on the existing `key` function, but provides an easy
way for users to specify a default value. This replaces the common pattern of:

    {{if key "service/app/max_conns"}}
      max_conns={{key "service/app/max_conns"}}
    {{else}}
      max_conns=500
    {{end}}

or

    {{ or (key "service/app/max_conns") "5" }}

with

    {{ key_or_default "service/app/max_conns" "5" }}

Fixes GH-368


ryanuber:
This LGTM!

PR #422, opened 2015-10-11
add timestamp special case 'unix'

macb:
Ran into a case where a unix timestamp would be very useful but wouldn't be able to get that out of the timestamp formatting. Thought it might warrant a special case to `timestamp` as opposed to its own function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -716,13 +716,20 @@ Returns the current timestamp as a string (UTC). If no arguments are given, the 
 {{timestamp}} // e.g. 1970-01-01T00:00:00Z
 ```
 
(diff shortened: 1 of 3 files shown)


sethvargo:
Hi @macb 

Thank you for the Pull Request. Can you explain your use case a bit more? The `timestamp` function is not really the core function of Consul Template; it's mostly used for debugging purposes and I would prefer to keep the API minimal if possible.

macb:
Sure @sethvargo. I ran into this when I was wanting to generate a bind file based on the consul node and service catalogs. I wanted a value to use as the domain's `serial`, which is an always incrementing integer. This seemed like a relatively straight forward way to generate that value instead of actually trying to increment an int from some place.

PR #426, opened 2015-10-14
Support for minimum of entries required.

takeda:
This feature introduces `minimum` function. It allows user to specify minimum number of nodes required for a valid configuration. If number of nodes returned is less than specified value, the template won't be updated and an error message will be generated. This can be used as an extra sanity check for critical services.

Example usage:
```nginx
upstream frontend { {{range service "web@sfo1" | minimum 3}} 
        server {{.Address}}:{{.Port}};{{end}}
}

server {
        listen 80 default_server;

        location / { 
                proxy_pass http://frontend;
        }
}
```

### DIFF
--- a/runner.go
+++ b/runner.go
@@ -272,9 +272,18 @@ func (r *Runner) Run() error {
 		// Attempt to render the template, returning any missing dependencies and
 		// the rendered contents. If there are any missing dependencies, the
 		// contents cannot be rendered or trusted!
-		used, missing, contents, err := tmpl.Execute(r.brain)
+		used, missing, contents, control, err := tmpl.Execute(r.brain)
 		if err != nil {
 			return err
+		}
+		if control != nil {
+			switch control.controlType {
+			case ControlTypeSkip:
+				log.Printf("[ERR] skipping template generation: %s", control.reason)
+				continue
+			default:
+				return fmt.Errorf("Invalid control type: %v", control.controlType)
+			}
 		}
 
 		// Add the dependency to the list of dependencies for this runner.
--- a/template.go
+++ b/template.go
@@ -18,6 +18,18 @@ type Template struct {
 	contents string
 }
 
+const (
+	// skip generating the template and log message.
+	ControlTypeSkip = iota
+)
+
+type ControlType struct {
+	// type of the issued control signal.
+	controlType uint
+	// reason that will be displayed to the user.
+	reason string
+}
+
 // NewTemplate creates and parses a new Consul Template template at the given
 // path. If the template does not exist, an error is returned. During
 // initialization, the template is read and is parsed for dependencies. Any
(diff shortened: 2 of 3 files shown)


sethvargo:
Hi @takeda 

Thank you for the pull request! We discussed this as a team, and we do not think this is Consul Template's responsibility. You should add resilience to your templates to gracefully deprecate if services are not available. You should also set up alerting around minimum services if that is a requirement for your organization :smile:

carlivar:
@sethvargo are there examples anywhere of resilience in templates for graceful degradation? That sounds more difficult and complex versus this simple approach. How might a template know the "last snapshot" of available services?

Alerting of minimum services does not seem relevant to this feature. That can (and should) be done independently.

sethvargo:
For example:

```liquid
{{ $services := service "web@sfo1" }}
{{ if gt 3 (len $services) }}
{{ range $services }}
  //
{{ end }}
{{ else }}
  // Graceful error message or static error
{{ end }}
```

The approach in this PR will kill the CT process if less than the minimum number of services are met. This somewhat goes against the concepts of CT. If you want to render templates without continuous watching, you can just use the `-once` flag.

takeda:
@sethvargo, the main reason for this PR is to make consul-template skip generating the template if certain conditions are not met, without crashing consul-template. I can submit another PR which would implement `skip` command which can be invoked in similar way for example:

```
{{ $services := service "web@sfo1" }}
{{ if gt 3 (len $services) }}
{{ range $services }}
  //
{{ end }}
{{ else }}
  skip "Not enough nodes in web@sfo1"
{{ end }}
```
I went with `minimum` function, because using it is more compact, but this approach could work too. Should I create another PR with that feature?

sethvargo:
Hi @takeda 

I think this requirement is outside the scope of Consul Template. If you have a load balancer, for example, and you don't have enough nodes in the balancer, rendering the template with stale data is not a good practice. If those nodes are truly down, they should not be receiving traffic.

takeda:
@sethvargo The feature requested supposed function as an assertion/fail safe mechanism. For example let say someone deployed a change (through configuration management) that caused the agent process on all nodes for particular service to die.

So let say a service normally has 12 nodes and suddenly consul is reporting that there are no healthy nodes, this most likely means something went south and better not make things worse by updating config. The service might be still perfectly fine, and since HAProxy does healtcheck itself having stale entries is way better than no entries.

123BLiN:
+1, same with nginx, the problem is that if no nodes returned - nginx will not be reloaded at all:
`nginx: [emerg] no servers in upstream`
this leads to lost all other fails of other services because nginx will never be reloaded. Things should not be like so, but I'm going to use @sethvargo workaround provided above. 
Something like:
```
{{ $services := service .Name }}
{{ if gt 0 (len $services) }}
{{ range $services }}
  server  {{.Address}}:{{.Port}};
{{ end }}
{{ else }}
  {{ range $services "any"}}
    server  {{.Address}}:{{.Port}};
  {{ end }}
{{ end }}
```
Thanks!

sethvargo:
Hi @123BLiN 

For nginx, you should use `nginx -t -c` as part of your reload command. For example:

```
command = "/usr/sbin/nginx -t -c /path/to/rendered.conf && /usr/sbin/nginx reload"
```

123BLiN:
Hi @sethvargo !

First of all thanks for the great work youve done!
If I understand you correct you advised to check nginx conf before reload, we are already check it, the problem is that if config wrong - no reload. So for example if we have service 2 down on all nodes then template is broken, but if after that will fail service 1 on one node - this will not be triggered :(

However your workaround works es expected, thanks!

PR #448, opened 2015-10-29
Introduce TLS/SSL Mutual Authentication

tnn:
Setting the config option ssl.key or CLI flag
'-ssl-key=/my/private/key.pem' allows the HTTPS client to authenticate
itself towards a Consul server which has the option "verify_incoming": true

CC: @sorenmat

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,5 +24,9 @@ _testmain.go
 .proc
 *.conf
 
+# IDE files
+*.iml
+*.idea
+
 bin/
 pkg/
--- a/cli.go
+++ b/cli.go
@@ -210,6 +210,13 @@ func (cli *CLI) parseFlags(args []string) (*Config, bool, bool, bool, error) {
 	}), "ssl-cert", "")
 
 	flags.Var((funcVar)(func(s string) error {
+		config.SSL.Key = s
+		config.set("ssl")
+		config.set("ssl.key")
+		return nil
+	}), "ssl-key", "")
+
+	flags.Var((funcVar)(func(s string) error {
 		config.SSL.CaCert = s
 		config.set("ssl")
 		config.set("ssl.ca_cert")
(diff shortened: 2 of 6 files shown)


sethvargo:
Hi @tnn 

Thank you for the pull request. I am a bit confused here though. How is this different than the existing `cert` option? It uses the exact same protocol for parsing the keypair and loading it into the tlsconfig.

```diff
+			cert, err := tls.LoadX509KeyPair(config.SSL.Cert, config.SSL.Key)
+			if err != nil {
+				return nil, err
+			}
+			tlsConfig.Certificates = []tls.Certificate{cert}
```

tnn:
@sethvargo Look closer :-)  `tls.LoadX509KeyPair(config.SSL.Cert, config.SSL.Key)` vs . `tls.LoadX509KeyPair(config.SSL.Cert, config.SSL.Cert)`

sethvargo:
I _think_ that's what we do. Could you give that a try?

tnn:
I guess I should have tried that before. It does actually work concatenating the two files together, but only because consul-template rely on the internal implementation of `tls.X509KeyPair(certPEMBlock, keyPEMBlock)`. The API of tls.LoadX509KeyPair is `(certFile, keyFile)`.

tnn:
@sethvargo So do you want this to be upstreamed or not?

sethvargo:
@tnn let met think about it :smile:

sethvargo:
I worry about it confusing users, but it's obviously already confusing users

tnn:
I understand your concern from existing users, but:
consul:
```
-ca_file
-cert_file
-key_file
```
curl:
```
--cert
--key
--cacert
```
terraform:
```
tls_private_key
tls_self_signed_cert
```
docker:
```
--tlscacert=ca.pem
--tlscert=server-cert.pem
--tlskey=server-key.pem
```
The only other software that I've touched with this behaviour is HAProxy. I was at least confused - but you don't need to remove the existing functionality (and this PR doesn't do that).

PR #450, opened 2015-10-30
Use renew-self instead of renew

sethvargo:


slackpad:
This change LGTM but the build failure looks like something is broken (not related to this change), but a real borkage and not just a flaky test.

sethvargo:
@slackpad there are vault issues with aws-sdk-go that need to be patced

sethvargo:
These have been fixed upstream and the build is green :smile:

PR #465, opened 2015-11-17
Adds support for de-duplication of template rendering between multiple instances

armon:
This adds a new de-duplication mode which is used to have a single instance of consul-template run the watchers, and the other instances receive the data in a single KV key. This works by:

* Generating a KV prefix based on the hash of the template `prefix = consul-template/dedup/MD5/`
* Locking a key for the template (`prefix/lock`)
* If leader, run views normally, and on render write the dependencies as (`prefix/data`)
* If follower, watch template data (`prefix/data`), re-render on data updates
* Vault data is excluded from this, as it should not be shared

In this system, instead of running 50 queries per template on 50 nodes (2500 queries), we instead run 50 queries on one node, and 100 KV watches on the others, for a total of 150. This results in a dramatic reduction of read pressure on Consul.

TODO:
- [x] Unit tests for DeduplicationManager
- [x] Unit tests for Runner changes

### DIFF
--- a/brain.go
+++ b/brain.go
@@ -53,6 +53,16 @@ func (b *Brain) Recall(d dep.Dependency) (interface{}, bool) {
 	return b.data[d.HashCode()], true
 }
 
+// ForceSet is used to force set the value of a depdency
+// for a given hash code
+func (b *Brain) ForceSet(hashCode string, data interface{}) {
+	b.Lock()
+	defer b.Unlock()
+
+	b.data[hashCode] = data
+	b.receivedData[hashCode] = struct{}{}
+}
+
 // Forget accepts a dependency and removes all associated data with this
 // dependency. It also resets the "receivedData" internal map.
 func (b *Brain) Forget(d dep.Dependency) {
--- a/brain.go
+++ b/brain.go
@@ -53,7 +53,7 @@ func (b *Brain) Recall(d dep.Dependency) (interface{}, bool) {
 	return b.data[d.HashCode()], true
 }
 
-// ForceSet is used to force set the value of a depdency
+// ForceSet is used to force set the value of a dependency
 // for a given hash code
 func (b *Brain) ForceSet(hashCode string, data interface{}) {
 	b.Lock()
--- a/cli.go
+++ b/cli.go
@@ -356,6 +356,9 @@ Options:
   -max-stale=<duration>    Set the maximum staleness and allow stale queries to
                            Consul which will distribute work among all servers
                            instead of just the leader
+  -dedup                   Enable de-duplication mode. Reduces load on Consul when
+                           many instances of Consul Template are rendering a common
+                           template.
   -ssl                     Use SSL when connecting to Consul
   -ssl-verify              Verify certificates when connecting via SSL
   -ssl-cert                SSL client certificate to send to server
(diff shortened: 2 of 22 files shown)


brain.go
```diff
@@ -53,6 +53,16 @@ func (b *Brain) Recall(d dep.Dependency) (interface{}, bool) {
 	return b.data[d.HashCode()], true
 }
 
+// ForceSet is used to force set the value of a depdency
```
sethvargo: dependency

cli.go
```diff
@@ -236,6 +236,7 @@ func (cli *CLI) parseFlags(args []string) (*Config, bool, bool, bool, error) {
 
 	flags.Var((funcBoolVar)(func(b bool) error {
 		config.Syslog.Enabled = b
+		config.set("syslog")
```
sethvargo: Oops. I'm surprised no one opened an issue on that. Good catch!

dedup.go
```diff
@@ -0,0 +1,427 @@
...
+		// Pull the current value from the brain
+		val, ok := d.brain.Recall(dp)
+		if ok {
+			td.Data[dp.HashCode()] = val
+		}
+	}
+
+	// Encode via GOB and LZW compress
+	var buf bytes.Buffer
+	compress := lzw.NewWriter(&buf, lzw.LSB, 8)
```
sethvargo: I'm curious the reason you chose lzw instead of gzip?

dedup.go
```diff
@@ -0,0 +1,427 @@
...
+	select {
+	case <-lockCh:
+		return false
+	default:
+		return true
+	}
+}
+
+// UpdateDeps is used to update the values of the dependencies for a template
+func (d *DedupManager) UpdateDeps(t *Template, deps []dep.Dependency) {
```
sethvargo: I think this function should return an error instead of just logging. We can catch that error in the runner.

sethvargo:
Hey @armon 

This is looking really good. At a high level, my only question is how this relates to quiescence timing. I'm envisioning a scenario where a quiescence timer causes an invalid/partial/non-existent template to be rendered because the leader hasn't written the data yet.

dedup.go
armon: LZW is way cheaper CPU wise, Gzip just has diminishing returns but does work better. I'm not opposed to switching, but figured LZW is lots of bang for the buck with minimal CPU cost.

armon:
@sethvargo I think the quiescence timers work like normal, since the DeduplicationManager shims in around the brain and doesn't touch that logic at all. Basically instead of updating one dependency at a time via a watcher, we are updating all them at once via the shared data file.

armon:
@sethvargo When you get a chance, this should be good to go now!

sethvargo:
Hey @armon 

I fixed the travis script, but it looks like there are some build failures. I can reproduce those failures locally too. Could you please take a look? Thanks! :smile:

bastichelaar:
Hi @armon, the deduplication mode works great! Load drops significantly. We didn't encounter any issues so far. I think this PR will make a lot of users very happy :+1:

armon:
@sethvargo Should be all working now!

ryanuber:
This LGTM, interesting approach!

sethvargo:
@armon looks like tests are still failing :frowning:

sethvargo:
And could you please add some documentation to the README for this? Thanks! :smile:

chrishaessig:
Thanks for adding this !

PR #466, opened 2015-11-18
Introduce a "joiner" function into template functions.

jessemyers:
The joiner design is inspired by Jinja2's [joiner](http://jinja.pocoo.org/docs/dev/templates/#joiner).

It allows insertion of a separator into template output in contexts where a loop index is not available
and where the existing, string-only "join" function cannot be used.

An example use case is iterating through Consul's KVs by Key and then joining the underlying values.

    {{ range $key, $pairs := tree "foo" | byKey }}
        $key = {{ $comma := joiner "," }}{{ range $pairs }}{{ sep $comma }}{{ .Value }}{{ end }}
    {{end}}

In this case, no loop index is available (because the tree returns a hash) and string-based join
is not useable because the value types are not strings.


jessemyers:
This PR is my first dig into the Go pool. Please don't hold back if something is not idiomatic.

ryanuber:
Hey @jessemyers,

Thanks for the PR. I agree that what you are trying to accomplish is less than a pleasure to implement with `text/template`'s syntax. A helper for this might be useful, but we definitely have some problems in this implementation. The goroutine started by the `joiner` function will never end. Over time, many goroutines will accumulate, causing a resource leak.

I feel like there may be a more generally useful way we could wrap maps with a template function, though I'm not sure what exactly it looks like. I'll think more on this.

jessemyers:
Thanks for the reply. I was worried about the goroutine never ending, but
I'm new enough to go not to know how serious of a problem that is.

I did think about an implementation that relies on a fixed length; that's
often known in advance.

sethvargo:
Hi there,

Once again, thank you for the Pull Request, we really appreciate it! Unfortunately, I do not think we should add this to Consul Template's API. This has not been an often-requested feature, and the implementation is difficult to get perfect. Instead, I would recommend using Go's templating language directly for this, something like:

```go
{{ range $key, $pairs := tree "foo" | byKey }}
{{ $key }} = "{{ range $i, $pairs }}{{ if gt 0 $i }},{{ end }}{{ .Value }}{{ end }}"
{{end}}
```

In this approach, you only append a leading comma if it's not the first element. This avoids the need to know the list's length in advance.

PR #470, opened 2015-11-25
Add configuration option to set custom delimiter

JBodkin:
Various configuration files actually utilise the "{{ }}" delimiters and therefore it becomes more difficult to create the configuration automatically (e.g. RabbitMQ). This PR introduces an new flag that allows the user to select an custom delimiter.

Example Usage
```
./consul-template -delimiter "|| ||"
```

Example Template
```
Hello || env "NAME" ||
```

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -281,6 +281,12 @@ func (cli *CLI) parseFlags(args []string) (*Config, bool, bool, bool, error) {
 		return nil
 	}), "log-level", "")
 
+	flags.Var((funcVar)(func(s string) error {
+		config.Delimiter = strings.SplitN(s, " ", 2)
+		config.set("delimiter")
+		return nil
+        }), "delimiter", "")
+
 	flags.BoolVar(&once, "once", false, "")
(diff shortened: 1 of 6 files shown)


ryanuber:
Hi @JBodkin,

Thanks for the PR. At first glance my thoughts are that we should probably avoid doing this for a few reasons. Go's `text/template` does allow you to escape these, albeit in a not-so-pretty way, with `{{"{{"}}`. It's my feeling that while it may be more convenient to use an alternate delimiter for some config files, it places a burden on all other templates to use the same delimiter since this is a global configuration option. I think if we were to implement something like this, we would need it to be a per-template configuration option, but this also gets messy if you want to use alternate delimiters in many files and causes some configuration sprawl as the templates are now coupled to consul-template's config.

@sethvargo thoughts?

sethvargo:
I agree with @ryanuber. We also have another issue tracking a potential move to a more robust templating system, which might also solve this. As such, I am going to close this for now. Sorry! :frowning:

PR #471, opened 2015-11-25
Create bin directory during travisci build

JBodkin:
Hopefully fixes the build


sethvargo:
Hi @JBodkin 

Thank you for the PR. I'm actually talking with the Travis folks, and I think they are planning to fix this on their end /cc @meatballhat

Issue #455, opened 2015-11-06
How to determine that some node or service in maintenance state?


argusua:
Hi. We are using CT for generate configuration files for nagios3 and we want exclude nodes in maintenance state, but not failed nodes. I can't find any solution to get checks which available by consul api /v1/health/node/<node> in CT.

PR #477, opened 2015-12-08
Make node/service status "maintenance" when in maintenance mode

sethvargo:
Fixes GH-455

### DIFF
--- a/dependency/health_service.go
+++ b/dependency/health_service.go
@@ -26,7 +26,7 @@ const (
 	HealthMaint    = "maintenance"
 
 	NodeMaint    = "_node_maintenance"
-	ServiceMaint = "_service_maintenance:"
+	ServiceMaint = "_service_maintenance"
 )
 
 // HealthService is a service entry in Consul.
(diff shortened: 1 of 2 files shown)


dependency/health_service.go
```diff
@@ -236,8 +239,13 @@ func ParseHealthServices(s ...string) (*HealthServices, error) {
 // "warning", and finally "passing". If there are no checks, the service will be
 // marked as "passing".
 func statusFromChecks(checks []*api.HealthCheck) (string, error) {
-	var passing, warning, unknown, critical bool
+	var passing, warning, unknown, critical, maintenance bool
 	for _, check := range checks {
+		if check.CheckID == NodeMaint || check.CheckID == ServiceMaint {
```
slackpad: service is formatted as <prefix>:<service id> so the service one needs to be a `StartsWith()`

slackpad:
LGTM - we will want to note in rolling this out that `maint != critical` since people might pick up nodes they might not otherwise have picked up before.

PR #479, opened 2015-12-09
Adds child process reaping so CT can run as PID 1 in a Docker container.

### LINKED ISSUE #428: consul-template is not reaping children in container
We are running consul-template as a simple wrapper around HAproxy. Whenever the configuration changes, we perform a soft restart of haproxy. However, the exited HAProxy processes are not reaped and left as zombie processes.

Would you consider having  consul-template  take the PID 1 responsibility and handle the SIGCHLD events? Otherwise we would need to  complicate the container with an init system.

Cheers,

Mark

slackpad:
Not quite ready for merge - need to do a little more testing but wanted to push it up so you can take a look.

Fixes #428

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -86,11 +86,12 @@ func (cli *CLI) Run(args []string) int {
 	}
 
 	// If they configured a child process reaper, start that now
-	if config.Reap {
+	if config.Reap || (!config.WasSet("reap") && os.Getpid() == 1) {
 		if !reap.IsSupported() {
-			err := fmt.Errorf("[WARN] Child process reaping requested but not supported on this platform")
+			err := fmt.Errorf("[WARN] Child process reaping is not supported on this platform, please set the 'reap' option to false")
 			return cli.handleError(err, ExitCodeConfigError)
 		}
+		log.Printf("[DEBUG] Automatically reaping child processes")
 
 		pids := make(reap.PidCh, 1)
 		errors := make(reap.ErrorCh, 1)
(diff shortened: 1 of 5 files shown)


cli.go
```diff
@@ -84,6 +85,30 @@ func (cli *CLI) Run(args []string) int {
 		return ExitCodeOK
 	}
 
+	// If they configured a child process reaper, start that now
+	if config.Reap {
+		if !reap.IsSupported() {
+			err := fmt.Errorf("[WARN] Child process reaping requested but not supported on this platform")
```
sethvargo: Maybe we should say something about disabling it in the config.

cli_test.go
```diff
@@ -489,6 +489,20 @@ func TestParseFlags_dry(t *testing.T) {
...
+		"-reap",
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if config.Reap != true {
+		t.Errorf("expected reap to be true")
+	}
+}
```
sethvargo: I think you want to add this too:

```go
 	if !config.WasSet("reap") {
 		t.Errorf("expected reap to be set")
 	}
```

config.go
```diff
@@ -433,6 +440,7 @@ func DefaultConfig() *Config {
 		MaxStale:        1 * time.Second,
 		Wait:            &watch.Wait{},
 		LogLevel:        logLevel,
+		Reap:            false,
```
sethvargo: What do you think about defaulting this to true when running as PID 1 (or 0):

```go
Reap: os.Getpid() < 2
```

slackpad:
I think this is good to go - was able to watch it working running inside Docker using this test C program as the command run by CT:

```
#include <stdlib.h>
#include <sys/types.h>
#include <unistd.h>

int main()
{
  pid_t child_pid;

  sleep(5);
  child_pid = fork();
  if (child_pid > 0) {
    sleep(2);
  }
  exit(0);
}
```

Issue #482, opened 2015-12-16
Ability to iterate over checks for a given service on a given node


pepov:
I would like to iterate over checks in order to map them as individual services in nagios using https://github.com/ustream/nagios-consul-plugin. What are you're opinions about this? Shall I move on and create a pull request or am I missing something?

sethvargo:
Hi @pepov 

This is not something we currently expose, mostly because it has not been requested. I'm not sure if check changes affect blocking queries, so you might not get the behavior you expect. 

/cc @ryanuber for input

pepov:
It would be nice if you could point me to the right direction and I would be happy to implement it and send a PR

sethvargo:
Hi @pepov 

You'll want to look through the Consul API first: https://github.com/hashicorp/consul/tree/master/api to find which struct fields you need. Then you have to mirror them in Consul Template in the dependency/health_service.go.

PR #490, opened 2015-12-22
Make Checks available through HealthService

pepov:
First shot on implementing https://github.com/hashicorp/consul-template/issues/482


sethvargo:
Thanks!

Issue #507, opened 2016-01-12
Hot loop reaping child process 0, wait contention


slackpad:
Pulling this out from https://github.com/hashicorp/consul-template/issues/442#issuecomment-170895780 and https://github.com/hashicorp/consul-template/issues/442#issuecomment-170899841 since it's a separate issue.

It looks like some users are encountering a situation where the child process reaper is running at 100% claiming to be `Reaping child process 0`. Setting `-reap=false` will work around this, but we need a fix to prevent this.

There's also an issue where CT reports `[ERR] (runner) error running command: wait: no child processes` - this is happening when the reaper steals the wait result from CT.

Both of these issues affect Consul as well, so we will need to update there with a similar fix.

slackpad:
It looks like the zero return value should kick us back to `WAIT` instead of `POLL`:

https://www.mkssoftware.com/docs/man3/wait3.3.asp

> The wait3() and wait4() functions return 0 if WNOHANG is specified and there are no stopped or exited children, and return the process ID of the child process if they return due to a stopped or terminated child process. Otherwise, they return -1 and set errno to indicate the error.

geniousphp:
I encounter this issue when I use CP 0.12.0 with `-reap=true`. I'm working with a container docker containing CP and HAProxy, CP has PID 1. When CP restart HAProxy 2 (and exactly 2)  times, it exists with the following msg error `[ERR] (runner) error running command: wait: no child processes`

slackpad:
@geniousphp interesting - that looks like a separate issue where the process reaper is "stealing" the wait result from Consul Template itself. I've got the hot loop fixed in the PR above but I'll also take a look at this.

geniousphp:
Ouups, maybe I should have opened a new issue because I never encounter the issue `Reaping child process 0`. I'm not sure if my issue is related to #442. May be I should wait for a new CP version compiled with with the last GoLang version and see again. Meanwhile, I continue using CP 0.10.0

slackpad:
@geniousphp no worries - they are closely related so I added your issue to this one so we can track them together.

slackpad:
Fix is in for Consul Template as well as Consul.

PR #511, opened 2016-01-13
Adds reap lockout so we don't disturb sub process waiting.

slackpad:
Testing this now - should fix the rest of https://github.com/hashicorp/consul-template/issues/507.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -86,6 +86,7 @@ func (cli *CLI) Run(args []string) int {
 	}
 
 	// If they configured a child process reaper, start that now
+	var reapLock sync.RWMutex
 	if config.Reap || (!config.WasSet("reap") && os.Getpid() == 1) {
 		if !reap.IsSupported() {
 			err := fmt.Errorf("[WARN] Child process reaping is not supported on this platform, please set the 'reap' option to false")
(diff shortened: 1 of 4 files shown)


slackpad:
CI will be broken until we merge https://github.com/hashicorp/go-reap/pull/3. I could add a compatibility wrapper if you think that's cleaner (should only affect CT and Consul, which will rev shortly as well).

slackpad:
Tested this fix using a Dockerfile like this:

```
FROM ubuntu
COPY bin/consul-template /tmp
COPY foo.hcl /tmp
COPY foo.tmpl /tmp
COPY zombie /tmp
CMD ["/tmp/consul-template", "-config", "/tmp/foo.hcl"]
```

And this configuration:

```
consul = "demo.consul.io:80"
log_level = "debug"

template {
  source = "/tmp/foo.tmpl"
  destination = "/tmp/foo.out"
  command = "/tmp/zombie"
}
```

And this template:

```
{{key "global/time"}}
```

Using zombie.c:

```
#include <stdlib.h>
#include <sys/types.h>
#include <unistd.h>

int main()
{
  pid_t child_pid;

  sleep(5);
  child_pid = fork();
  if (child_pid > 0) {
    sleep(2);
  }
  exit(0);
}
```

Without the fix if you twiddle the KV using http://demo.consul.io/ui/#/nyc3/kv/global/time/edit you can get the race condition to happen and report `[ERR] (runner) error running command: wait: no child processes` within a few tries. After applying the fix I wasn't able to get this to occur, and I saw child process reaping messages in the debug logs for the intentional zombie processes.

Need to apply a similar fix to Consul as well.

PR #529, opened 2016-02-01
Add 'leave-on-failure' param to exit in case of error.

sandoracs:
https://groups.google.com/forum/m/#!topic/consul-tool/xo4VJLa7dNk

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,29 +23,30 @@ You can download a released `consul-template` artifact from [the Consul Template
 Usage
 -----
 ### Options
-|       Option      | Description |
-| ----------------- |------------ |
-| `auth`            | The basic authentication username (and optional password), separated by a colon. There is no default value.
-| `config`          | The path to a configuration file or directory of configuration files on disk, relative to the current working directory. Values specified on the CLI take precedence over values specified in the configuration file. There is no default value.
-| `consul`*         | The location of the Consul instance to query (may be an IP address or FQDN) with port.
-| `deduplicate`     | Enable de-duplication of template rendering. If many instances of consul-template render the same template this reduces the load on Consul. Please see the "De-Duplication Mode" section in caveats for more information.
-| `dry`             | Dump generated templates to the console. If specified, generated templates are not committed to disk and commands are not invoked. _(CLI-only)_
-| `log-level`       | The log level for output. This applies to the stdout/stderr logging as well as syslog logging (if enabled). Valid values are "debug", "info", "warn", and "err". The default value is "warn".
-| `max-stale`       | The maximum staleness of a query. If specified, Consul will distribute work among all servers instead of just the leader. The default value is 1s.
-| `once`            | Run Consul Template once and exit (as opposed to the default behavior of daemon). _(CLI-only)_
-| `pid-file`        | The path on disk to write Consul Template's PID file.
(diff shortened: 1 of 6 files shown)


pepov:
I'm referring back to the thread:
https://groups.google.com/forum/#!msg/consul-tool/xo4VJLa7dNk/k9l2LkyPGAAJ

However maybe it would be better to have a limit on the number of retry attempts - instead of just disabling retries - to be able to survive intermittent errors.

sethvargo:
Hi @sandoracs 

Thank you for the Pull Request. I'm very hesitant to add this behavior for a few reasons.

1. The term "error" is very generic here. Judging from the code, it seems like you want to exit whenever the Consul API request returns an error. Consul Template already exits in the event of a command failure or internal error, so I think the naming should somehow reflect this is only for Consul API queries.

2. This is opening the doors to really complex logic that I'd like to avoid, namely "I want to error if this fails, but not that". Of special note is the Consul K/V store. There exists a _very_ common pattern where users watch keys that only exist as feature flags or tunables. Sometimes these keys do not exist and can be used to control flow:

  ```ruby
  {{ if key "foo" }}
    # ...
  {{ else }}
    # ...
  {{ end }}
  ```

  or less commonly with range:

  ```ruby
  {{ range service "foo" }}
    # ...
  {{ else }}
    # ...
  {{ end }}
  ```

  Adding a configuration option like this breaks that assumption and could cause a very confusing UX, especially for users who are not very familiar with the Consul internals.

3. As already seen by just the one comment, it opens a massive rabbit hole for highly-specific use cases that only satisfy a small percentage of users while complicating the entire API. I think @pepov's point about having a certain number of errors with retries is a good idea too, but I don't think it belongs in the project.

Just to be clear, the code here looks really good (good job), but I'm more concerned with the project-level implications of adding this functionality and what it means for the API. I've read through the mailing list post, and I'm not sure I entirely understand the use case. From what I've read, you have CT processes that are "stale", but I don't quite know what that means. CT is designed to be incredibly fault tolerant in the face of change. Perhaps if you explain the use case a bit more, I can provide a more process-oriented solution :smile:. Thanks!

pepov:
Hey, thanks for the detailed answer on this! 

Our most important use case is this: we use lots of dependencies, that span all our datacenters, e.g.:
```
{{range datacenters}}
# Datacenter: {{.}}{{with $d := .}}
{{range nodes (print "@" $d)}}
local-data-ptr: "{{.Address}}  {{.Node}}.node.{{$d}}.consul."{{end}}
{{end}}{{end}}
```

If we lose an entire datacenter failing rpc queries to that datacenter will block the entire template to be up-to-date, thus it gets stale.

sethvargo:
Hi @pepov 

Thank you for your reply. Can you clarify what you mean by "lose a data center". If a datacenter goes entirely offline, I think you have slightly bigger problems than Consul Template :wink:. 

Consul Template queries the catalog endpoint to get the list of data centers. If the data center is removed from the catalog, it shouldn't block on the `nodes` query with the n-pass evaluation.

pepov:
We may have connectivity issues to satellite datacenters in which case we should keep the issue localized. If consul template would be able to update the template under such circumstances as well, it might be ok for us. Or maybe if I could have a simple health check API available in consul-template telling me about the staleness of the managed templates...

sethvargo:
Hi @pepov 

I think the latter option seems most appropriate (having a health check that tells consul template about staleness)

sandoracs:
Hi @sethvargo,

Thank you very much for your detailed answer. I will take a look at the code and try to find the proper solution. :)

Issue #534, opened 2016-02-08
Reloading configuration duplicates templates


wwalker:
I have a set of templates built from vault variables.  There are only 60 vault variables.  they have a minimum lease time of 30 seconds.  One would expect an average of 4 requests per second max (many of hte variables have longer lease times).

The vault audit log shows that it is handling over 40 requests per second from consul-template.

It seems to be throttled be vault performance (vault it using ~180 % CPU), so it appears that consul-template would be doing  even more if it could.

Vault -0.3.1.  the o.4.0 upgrade is planned for the next couple of weeks.
consul-template is 0.10.0
vault backend is consul 0.6.0

jefferai:
I'm not sure on the CT side but at this point I'd recommend planning for a Vault 0.5 upgrade instead, especially if it's still a few weeks out.

sethvargo:
Hi @wwalker 

As Jeff said, I would recommend going to Vault 0.5, since it will be released shortly.

I also believe there were known issues in Consul Template 0.10 that have been fixed in 0.12.1(and then 0.12.2 was a minor patch release for folks running in Docker containers). https://github.com/hashicorp/consul-template/blob/master/CHANGELOG.md#v0121-january-7-2016

Can you try upgrading and let me know if that fixes the issue?

wwalker:
Upgraded to 0.12.2.

Log shows that, during 1 second, it checked the same variable 6 times:

    2016/02/08 16:00:13 [DEBUG] ("vault(secret/ops/configs/properties/product_groups/indeedu-golf/redis.password)") vault returned the secret
    2016/02/08 16:00:13 [DEBUG] (view) "vault(secret/ops/configs/properties/product_groups/indeedu-golf/redis.password)" no new data (contents were the same)
    2016/02/08 16:00:13 [DEBUG] ("vault(secret/ops/configs/properties/product_groups/indeedu-golf/redis.password)") querying vault with &{AllowStale:true WaitIndex:1454968813 WaitTime:1m0s}
    2016/02/08 16:00:13 [DEBUG] ("vault(secret/ops/configs/properties/product_groups/indeedu-golf/redis.password)") pretending to long-poll for "15s"
    2016/02/08 16:00:13 [DEBUG] ("vault(secret/ops/configs/properties/product_groups/indeedu-golf/redis.password)") vault returned the secret
[...]

PR #537, opened 2016-02-09
Stop duplicating config templates in Merge() - 534

wwalker:
My shot at fixing #534


wwalker:
Well the CI says I've added  a race condition.  I'm new to go and not a big concurrent/parallel developer, so I assume that I have.

Issue #542, opened 2016-02-15
Vault .Data | toJSON | plugin intermittently does not pass data?


carlpett:
I'm trying to pass a vault credential into a plugin, without mixed success. This is the interesting part of the template: `{{ with vault "secret/my-creds" }}{{ .Data | toJSON | plugin "python" "my-plugin.py" }}{{ end }}`, 

Here are two different plugins, first a simple one:
```
import sys,json
print "input:{0}".format(sys.argv[1])
secret = json.loads(sys.argv[1])
print secret
```

Then one which is apparently not that simple:
```
import sys,json
print "input:{0}".format(sys.argv[1])
secret = json.loads(sys.argv[1])
print secret['username']
```

Output for the first example:
```
$ consul-template -config test.hcl -dry -once
> output
input:{"password":"bar","username":"foo"}
{u'username': u'foo', u'password': u'bar'}
```

But for the second:
```
$ consul-template -config test.hcl -dry -once
template: template: testdata:1:54: executing "testdata" at <plugin "python" "sim...>: error calling plugin: exec "python": exit status 1

stdout:

input:null


stderr:

Traceback (most recent call last):
  File "simple.py", line 4, in <module>
    print secret['username']
TypeError: 'NoneType' object has no attribute '__getitem__'
```

What's going on? It looks like when I try to explicitly read the data passed in, it is magically blanked. If I copy the json that is given as input and manually run python, all works fine:
```
$ python simple.py '{"password":"bar","username":"foo"}'
input:{"password":"bar","username":"foo"}
foo
```

The config file looks like this:
```
vault {
  address = "http://localhost:8200"
  token = "my-token"
}

template {
  source = "testdata"
  destination = "output"
}
```
(I've tried both on our production vault cluster as well as a local `-dev`)

I'm slowing going mad trying to understand what is happening here. I'm guessing it more or less has to be something in my code, but I can't figure it out.

I'm on consul-template 0.12.2, vault 0.5, windows 8.1.

sethvargo:
Hi @carlpett 

The first time the template runs, it will return the "empty" version of the data. Your plugin needs to exit early if it's given no arguments or empty arguments.

carlpett:
...
Haha. Ok, at least it turns out I'm not crazy, then :) This might deserve a mention in the plugin part of the documentation. I can find it in the documentation know that I know it is there, but I didn't see it before.

carlpett:
[...]

PR #543, opened 2016-02-16
Clarify plugin requirements wrt multi-phase execution

carlpett:
While mentioned in a few places already, the plugin section does not contain a reminder about handling cases before multi-phase execution is resolved (see #542).
Also added a note about providing non-template parameters to plugins.


sethvargo:
Thanks!

PR #546, opened 2016-02-18
Switches to the new pooled transport.

### LINKED ISSUE #545: Configure Consul API client with a pooled connection
The go-cleanhttp library was updated to make connection pooling and re-use explicit. Before, we used finalizers to close idle connections (which had a lot of problems on its own), but now we will make a new connection to Consul on every request.

https://github.com/hashicorp/consul/pull/1731

Since consul-template is in a long-term relationship with its Consul server, we should update our API client instance to use the new pooled transport. Here's an example from Vault that we can draft off of:

https://github.com/hashicorp/vault/pull/1090

A single connection is probably fine, and should get things back to the previous behavior.

slackpad:
This closes #545.

### DIFF
--- a/runner.go
+++ b/runner.go
@@ -22,6 +22,7 @@ import (
 	dep "github.com/hashicorp/consul-template/dependency"
 	"github.com/hashicorp/consul-template/watch"
 	consulapi "github.com/hashicorp/consul/api"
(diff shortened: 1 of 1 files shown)


slackpad:
Not sure if we want to apply this to Vault as well - seems like the churn should be way less there.

PR #555, opened 2016-02-26
Using Godep for vendoring

mssola:
Signed-off-by: Miquel Sabaté Solà <msabate@suse.com>


mssola:
For some reason `godep` was producing a wrong environment (maybe a bug on `godep` ?). `govendor` seems to work fine. Would that be an issue ?

jefferai:
@mssola The other HC tools have standardized on `godep`, so for consistency CT should as well. It would be useful if you can elaborate on the problems you saw, as we may know the way to fix it.

mssola:
@jefferai it was a mix of a messed up environment and me being frustrated with godep in other projects :P Now it should be fine.

jefferai:
Hi @mssola,

I'm not the maintainer of CT, so it's not up to me to merge this. I was simply informing you of the preference for `godep`.

sethvargo:
Hi @mssola 

Thank you for opening a PR. I have updated the build scripts to vendor all dependencies at release and tag. We are not going to pursue fully committing all the dependencies into the mainline source at this time, but you will be able to build off of any tag moving forward. Thanks! :smile:

PR #559, opened 2016-03-01
clarify wait is actually seconds

jippi:


sethvargo:
Hi @jippi 

It's not actually seconds, it's a number with a suffix. So you can say `100ms` or `10s` or `5m` and those are all valid inputs :smile:

jippi:
@sethvargo okay! :)

PR #576, opened 2016-03-13
Add CLI option to ignore commands execution

bhuisgen:
This patch adds a CLI option to ignore commands execution after templates rendering.

With my docker CT, I have two levels of configuration : 
- the first at init stage (container configuration) where consul-templates needs only to render templates and exit (once option enabled). No services are running at this stage.
- the second (services auto-reconfiguration) where services are running, consul-templates runs in daemon to render new configuration and execute commands to reload services if needed.

This patch allows me to have the same configuration files (with the commands) for the two configuration levels.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Usage
 | `consul`*         | The location of the Consul instance to query (may be an IP address or FQDN) with port.
 | `deduplicate`     | Enable de-duplication of template rendering. If many instances of consul-template render the same template this reduces the load on Consul. Please see the "De-Duplication Mode" section in caveats for more information.
 | `dry`             | Dump generated templates to the console. If specified, generated templates are not committed to disk and commands are not invoked. _(CLI-only)_
-| `ignore`          | Ignore commands. If specified, commands are not executed after templates rendering. _(CLI-only)_
+| `ignore-commands` | Ignore commands. If specified, commands are not executed after templates rendering. _(CLI-only)_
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -30,6 +30,7 @@ Usage
 | `consul`*         | The location of the Consul instance to query (may be an IP address or FQDN) with port.
 | `deduplicate`     | Enable de-duplication of template rendering. If many instances of consul-template render the same template this reduces the load on Consul. Please see the "De-Duplication Mode" section in caveats for more information.
 | `dry`             | Dump generated templates to the console. If specified, generated templates are not committed to disk and commands are not invoked. _(CLI-only)_
+| `ignore`          | Ignore commands. If specified, commands are not executed after templates rendering. _(CLI-only)_
```
jippi: `ignore-commands` maybe? 

`ignore` seem too broadly named to me

sethvargo:
Hi @bhuisgen 

Thank you for opening a Pull Request. I spent some time thinking about this and chatted internally with our team. We will not be merging this pull request, but I wanted to take a few minutes to explain why.

At it's core, Consul Template is a highly-fault tolerant system that's designed to be declarative. Having the behavior change conditionally, despite what's specified in the config, goes against how we've designed the tool. I also think this creates a very complex conditional logic that difficult to trace when debugging something.

Instead, I believe this kind of logic should live in the script itself. For example, you can use an environment variable or checking for the presence of a file as a pre-flight condition in your script before execution. Thanks! :smile:

PR #577, opened 2016-03-14
Working vault dump template

carlpett:
Now, while I agree that it should be discouraged, I actually had to dump my entire vault (long story, but we want to do a clean install, and have the possibility to re-add the secrets we had) today. The example given in the README doesn't work, however, mainly due to not taking recursion into consideration.

So, I took a swing at fixing it. This change describes a template that worked on my install at least, and should recurse properly. Now, whether this should be something which is advertised on the README is another question...

### DIFF
--- a/README.md
+++ b/README.md
@@ -329,13 +329,20 @@ The trailing slash is optional in the template, but the generated secret depende
 To iterate and list over every secret in the generic secret backend in Vault, for example, you would need to do something like this:
 
 ```liquid
-{{range secrets "secret/"}}
-{{with secret (printf "secret/%" .)}}
-{{range $k, $v := .Data}}
-{{$k}}: {{$v}}
-{{end}}
-{{end}}
-{{end}}
+{{- define "recursive-secret" }}
+  {{- $prefix := . }}
+  {{- range secrets (printf "secret/%s" .) }}
+    {{- if regexMatch ".+/$" . }}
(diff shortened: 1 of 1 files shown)


sethvargo:
Hi @carlpett 

Thank you for taking a stab at this. I just updated the readme so the template actually works, but I don't want to encourage this fully recursive behavior. The example in the readme is just to show folks how printf works, since it's a common question that arises. I'm going to close this issue, but it's googleable and searchable for anyone who might have this use case. Thanks! :smile:

carlpett:
I recently had to polish this abomination a bit more. So, if anyone else needs to recursively dump to JSON, here it is:

```
{{- define "recursive-secret" }}
  {{- $path := . }}
  {{- $key := . | regexReplaceAll ".+/(.+?)/?$" "$1" }}
  {{- if regexMatch ".+/$" . }}
    "{{ $key }}": {
    {{- range $idx, $secret := secrets . }}
      {{ if $idx }},{{ end }}{{- template "recursive-secret" (printf "%s%s" $path .) }}
    {{- end }}}
  {{- else }}
    {{ with secret . }}
      "{{ $key }}": {{ .Data | toJSON }}
    {{- end }}
  {{- end }}
{{- end }}
{ {{- template "recursive-secret" "secret/path/with/trailing/slash/" }} }
```

The output is not exactly well formatted. On the other hand, the template isn't exactly pretty either...

Issue #574, opened 2016-03-09
ctmpl Template auto-discovery from a folder


samifruit514:
Hello,

I need consul-template to find recursively all the ctmpl files in a folder and render at a destination where the folder structure is the same.

**Context:**
I have a web application with several consul template files at different folder levels. Example:
./app-tmpl/conf/pluginA/configs.php.ctmpl
./app-tmpl/conf/pluginA/include/confA.php.ctmpl
./app-tmpl/conf/pluginB/configuration.php.ctmpl
./app-tmpl/conf/pluginB/configuration.php.ctmpl

Our development team can add any ctmpl files at any time in any sub folders and they don't have access to the initial configuration.hcl

**The requirements:**
We need to render all templates without having to put specific template paths in the initial configuration.hcl. The result expected would be something like:

./app/conf/pluginA/configs.php
./app/conf/pluginA/include/confA.php
./app/conf/pluginB/configuration.php
./app/conf/pluginB/configuration.php

**Suggestion**
We could render the files at the same directory levels where the templates are **OR** We could have a config like this:
```
template {
    source      = "./app-tmpl/"
    destination = "./app/"
}
```
where consul-template could detect if source and destination are folders and proceed with rendering instead of giving the following error: "read ./app-tmpl/: is a directory"

This way, our DEV team could add any templates without changing the initial configuration.hcl file.

Do you think it could be a good idea? If yes, I could proceed with a PR.

samifruit514:
Ok I created a PR anyways: https://github.com/hashicorp/consul-template/pull/580

sethvargo:
Hi @samifruit514 

Thank you for opening an issue (and a PR). I'll respond here to keep everything in one place. 

In short, I don't think this is a good idea for Consul Template because it only gets you "halfway" there. One of the primary use cases for Consul Template is the arbitrary "command" parameter that executes in the event that a template changes. Perhaps you are not using that functionality in your use case, but it's definitely the more common functionality. I think the UX around pulling from multiple templates as a directory could be very confusing:

- Does a command render if all the templates are changed?
- Does a command render if one of the templates are changed?
- Does a command render if a template is added or removed?

Additionally, it will become inevitable that someone wants to control the command on a per-template basis, which is exactly why the command parameter is specified at a per-template level.

That being said, it's still possible to get the development behavior. Consul Template's `config` option _can_ be the path to a directory, and each file in that directory is loaded and merged into a single config. In other words, you could do this:

```hcl
[...]

PR #580, opened 2016-03-15
Adding support for template directory autodiscovery

samifruit514:
This feature finds templates under a specific directory to a specific directory.

For example:
```
template {
    source      = "./app-tmpl/"
    destination = "./app/"
}
```
will give a "template{}" entry for each templates found in source directory.

For more info:
https://github.com/hashicorp/consul-template/issues/574


sethvargo:
https://github.com/hashicorp/consul-template/issues/574#issuecomment-202130068

PR #589, opened 2016-03-22
Add template-specific wait timers

tgwizard:
Previously you could only configure the global `wait` parameter. Now you can set a `wait` parameter for each template. The template-specific wait takes precedence over the global wait.

This will make it simpler to run consul-template and render configurations where you want to limit the rate of reloads for some services. I guess you could do that with the global wait parameter, but then you'd need to run multiple consul-template processes.

There is one edge-case which this PR ignores. There could be multiple `wait`s for the same source template. This PR just takes the first one, and uses that. I'm not sure what the use-case for multiple identical source templates are, or what the expected result would be to specify different waits, or no waits.

What do you think?

### DIFF
--- a/runner.go
+++ b/runner.go
@@ -134,10 +134,21 @@ func (r *Runner) Start() {
 
 	for {
 		// Enable quiescence for all templates if we have specified wait intervals.
-		if r.config.Wait.Min != 0 && r.config.Wait.Max != 0 {
-			for _, t := range r.templates {
+		for _, t := range r.templates {
+			for _, ctemplate := range r.configTemplatesFor(t) {
+				if ctemplate.Wait.Min != 0 && ctemplate.Wait.Max != 0 {
+					if _, ok := r.quiescenceMap[t.Path]; !ok {
+						log.Printf("[DEBUG] (runner) enabling template-specific quiescence for %q", t.Path)
+						r.quiescenceMap[t.Path] = newQuiescence(
+							r.quiescenceCh, ctemplate.Wait.Min, ctemplate.Wait.Max, t)
+					}
+					break
+				}
+			}
+
+			if r.config.Wait.Min != 0 && r.config.Wait.Max != 0 {
 				if _, ok := r.quiescenceMap[t.Path]; !ok {
-					log.Printf("[DEBUG] (runner) enabling quiescence for %q", t.Path)
+					log.Printf("[DEBUG] (runner) enabling global quiescence for %q", t.Path)
 					r.quiescenceMap[t.Path] = newQuiescence(
 						r.quiescenceCh, r.config.Wait.Min, r.config.Wait.Max, t)
 				}
--- a/README.md
+++ b/README.md
@@ -273,6 +273,10 @@ template {
   // This is the maximum amount of time to wait for the optional command to
   // return. Default is 30s.
   command_timeout = "60s"
+  perms           = 0600
+  backup          = true
+  wait            = "2s:6s"
+}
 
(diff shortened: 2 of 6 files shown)


tgwizard:
Hmm, I'm not sure why the tests fail. The pass on my computer.

samifruit514:
I think their travis-ci check is not working properly. Their vault backend returns toomanyconnections.

tgwizard:
@sethvargo does this look good to you?

runner.go
```diff
@@ -140,10 +140,21 @@ func (r *Runner) Start() {
 
 	for {
 		// Enable quiescence for all templates if we have specified wait intervals.
-		if r.config.Wait.Min != 0 && r.config.Wait.Max != 0 {
-			for _, t := range r.templates {
+		for _, t := range r.templates {
```
ceh: Could we refactor this as an unexported `*Runner` method?

runner.go
```diff
@@ -140,10 +140,21 @@ func (r *Runner) Start() {
...
+			for _, ctemplate := range r.configTemplatesFor(t) {
+				if ctemplate.Wait.Min != 0 && ctemplate.Wait.Max != 0 {
```
ceh: What do you think about inverting the boolean statement and use `continue` if `Wait.Min` or `Wait.Max` is `0`, to reduce the indentation level.

runner.go
```diff
@@ -140,10 +140,21 @@ func (r *Runner) Start() {
...
+					if _, ok := r.quiescenceMap[t.Path]; !ok {
+						log.Printf("[DEBUG] (runner) enabling template-specific quiescence for %q", t.Path)
+						r.quiescenceMap[t.Path] = newQuiescence(
+							r.quiescenceCh, ctemplate.Wait.Min, ctemplate.Wait.Max, t)
+					}
+					break
+				}
+			}
+
+			if r.config.Wait.Min != 0 && r.config.Wait.Max != 0 {
```
ceh: Do you think it will read better if we invert the if statement here as well, in order to reduce indentation level?

runner.go
tgwizard: Perhaps. I left it in the style it was previously written.

tgwizard:
Ping @sethvargo et al., does this look good to you?

tgwizard:
Ping @sethvargo @slackpad @hashicorp, is it worth pursuing this PR, or should I attempt some other solution?

sethvargo:
Hi there,

Sorry for the delay in reviewing this. This PR is a fairly large core change, and I haven't had the chance to sit down and review it. Additionally, there are no tests, which makes it hard to merge as-is.

That being said, at a high level, I'm not sure it's worth adding template-level quiescence in CT. I think, if you want that level of granularity, you should just run multiple CT processes that specify the different wait intervals. What do you think?

tgwizard:
Thanks for the response @sethvargo. I agree that we could achieve similar results by running multiple CT processes. That would mean that our infrastructure configuration tooling (chef in this case) would need to support setting up two different services. Also, running two processes would increase the footprint slightly.

I do think that specifying quiescence like this on a template level should have a place in consul-template. If you agree I'd be happy to add some tests. I could probably copy/take inspiration from this test: https://github.com/hashicorp/consul-template/blob/8e35c8e991a46f31a6864ef5bcd595ab26a7a2ad/runner_test.go#L487-L539

slackpad:
Went ahead and rebased this one and added some tests under #618. Thanks @tgwizard!

Issue #591, opened 2016-03-23
Sending SIGHUP does causes consul-template to run out of file descriptors


benpillet:
I put a kill -HUP in a while loop like this:
`while [ 1 ]; do pkill -HUP -f consul-template; sleep 1; done`
and it hit an error with too many open files.  I suspect it's not letting go of something on the reconfig.  Version is 0.11.0.  Please let me know if you need more details.

jippi:
@benpillet can you try 0.14.0 ? :)

benpillet:
Just upgraded to 0.14.0 and seeing same issue.  Some more details.  The symptom doesn't show up if there aren't any templates.  I had it configured with no templates and it held steady at about 18 file descriptors.  When I added a template and sent SIGHUP, the file descriptors incremented by about 25 on each SIGHUP.

sethvargo:
Hi @benpillet 

Can you please share your CT config/command as well as the CT templates you're using?

benpillet:
Run command from puppet module: `/usr/local/bin/consul-template -config /etc/consul-template/config`

/etc/consul-template/config/config.json:
`consul = "localhost:8500"
token = ""
retry = "10s"

log_level = "debug"
template {
  source = "/etc/consul-template/clusterit.ctmpl"
  destination = "/etc/clusterit.consul"
  command = "true"
}
`

and template:
`# clusterit-consul.conf

# GROUPS

{{range services}}GROUP:role-{{.Name}}{{range service .Name}}
{{.Node}} # {{.Address}}{{end}}

{{end}}
`

sethvargo:
Hi @benpillet 

Can you share the output of running the command in debug mode as a GitHub gist after you send it sighup a few times?

benpillet:
https://gist.github.com/benpillet/c8520db0e3eb6a299ad0
The TCP connection lines from lsof are the ones that increase the file descriptor count.

PR #595, opened 2016-03-27
Close idle connections when stopping

sethvargo:
This fixes GH-591. By default (and correctly), the HTTP client maintains open connections to the Consul server. When reloading, we were not clearing those idle connections, thus they would accumulate  slowly over time, eventually using all the file descriptors.

This PR does a bit of refactoring so we have access to the underlying http.Transport so the watcher can close those connections during its shutdown procedure.


NZKoz:
This PR appears to break envconsul, attempting to build now I get:

```
➜  envconsul git:(master) ✗ go build
# github.com/hashicorp/envconsul
./runner.go:607: clients.Add undefined (type *dependency.ClientSet has no field or method Add)
./runner.go:615: clients.Add undefined (type *dependency.ClientSet has no field or method Add)
```

The two calls in question are [adding the consul client](https://github.com/hashicorp/envconsul/blob/master/runner.go#L585) and [adding the vault client](https://github.com/hashicorp/envconsul/blob/master/runner.go#L593)

Was the API change intended?  If so, any tips for a PR to send to envconsul?

Issue #572, opened 2016-03-08
0.14 regression with deduplicate enabled?


skippy:
in 0.13, this worked fine.  I migrated to 0.14 and did a rolling restart of haproxy containers, and they all went down.  after a wait, one of them came back up but none of the others did.

on the instances that didn't come up, I was seeing this:
```shell
PID   USER     TIME   COMMAND
    1 root       0:00 {run.sh} /bin/bash /run.sh
   14 root       0:00 /usr/local/bin/consul-template -config /consul-template/config.d -consul 10.0.0.68:8500
   26 root       0:00 sh
   65 root       0:00 ps
```

and consul-template was not starting haproxy.  I was expecting to see this:
```shell
PID   USER     TIME   COMMAND
    1 root       0:00 {run.sh} /bin/bash /run.sh
   14 root       0:06 /usr/local/bin/consul-template -config /consul-template/config.d -consul 10.0.0.28:8500
  956 haproxy    0:00 /usr/sbin/haproxy -D -p /var/run/haproxy.pid -f /etc/haproxy/haproxy.cfg -sf 951
  957 root       0:00 sh
  962 root       0:00 ps
```

the config file, with dedup enabled, is this:
```
template {
  source      = "/consul-template/template.d/haproxy.tmpl"
  destination = "/etc/haproxy/haproxy.cfg"
  command     = "touch /var/run/haproxy.pid; /usr/sbin/haproxy -D -p /var/run/haproxy.pid  -f /etc/haproxy/haproxy.cfg -sf $(cat /var/run/haproxy.pid) || true"
  backup      = true
}
deduplicate {
  enabled = true
  prefix = "consul-template/haproxy/"
}
```

with dedup on, for the servers which did not start haproxy, I see this (log-level INFO)
```
2016/03/08 22:44:57 [INFO] consul-template v0.14.0
2016/03/08 22:44:57 [INFO] (runner) creating new runner (dry: false, once: false)
2016/03/08 22:44:57 [INFO] (runner) creating consul/api client
2016/03/08 22:44:57 [INFO] (runner) creating vault/api client
2016/03/08 22:44:57 [INFO] (runner) creating Watcher
2016/03/08 22:44:57 [INFO] (runner) starting
2016/03/08 22:44:57 [INFO] (dedup) starting de-duplication manager
2016/03/08 22:44:57 [INFO] (runner) running
2016/03/08 22:44:57 [INFO] (dedup) attempting to create session
[...]

Issue #608, opened 2016-04-14
Resource utilization issues


bmonkman:
On certain pools of servers I have started seeing issues with consul-template since upgrading from 0.10 to 0.14 where it will start using about 75% of the total cpu on a 2 core vm, then start ramping up memory usage until the box runs out of memory. I'm looking at a consul-template process right now that's using over 5.5GB of memory.
I had to send it a HUP to turn on info level logging which seems to have calmed down the cpu usage a bit, but now it's in a state where it sits at next to no cpu usage for a while and then spikes up to 100% usage for a few seconds.

Some strangeness from the logs, after I switched to info logging, for about 10 minutes it spit about 100 log lines per second of this:

```
Apr 14 11:10:34 ### consul-template[11885]: (runner) quiescence maxTimer fired for /###/service-discovery-ingress-egress.conf
Apr 14 11:10:34 ### consul-template[11885]: (runner) quiescence maxTimer fired for /###/service-discovery-ingress-egress-backends.conf
Apr 14 11:10:34 ### consul-template[11885]: (runner) quiescence maxTimer fired for /###/service-discovery-ingress-egress-multi-dc-backends.conf
```

That has stopped and now it's just printing it's usual minTimer logging every 5 seconds as configured (though even that is more than I'd expect. It seems to write multiple logs about each configured template:
(This is all within 1 second)

```
Apr 14 11:45:09 ### consul-template[11885]: (runner) quiescence minTimer fired for /###/service-discovery-ingress-egress-backends.conf
Apr 14 11:45:09 ### consul-template[11885]: (runner) received template "/###/service-discovery-ingress-egress-backends.conf" from quiescence
Apr 14 11:45:09 ### consul-template[11885]: (runner) running
Apr 14 11:45:09 ### consul-template[11885]: (runner) diffing and updating dependencies
Apr 14 11:45:09 ### consul-template[11885]: (runner) watching 3 dependencies
Apr 14 11:45:09 ### consul-template[11885]: (runner) quiescence minTimer fired for /###/service-discovery-ingress-egress-multi-dc-backends.conf
Apr 14 11:45:09 ### consul-template[11885]: (runner) received template "/###/service-discovery-ingress-egress-multi-dc-backends.conf" from quiescence
Apr 14 11:45:09 ### consul-template[11885]: (runner) running
Apr 14 11:45:09 ### consul-template[11885]: (runner) quiescence minTimer fired for /###/service-discovery-ingress-egress.conf
[...]

PR #611, opened 2016-04-17
Do not share catalog node data

sethvargo:
Fixes https://github.com/hashicorp/consul-template/issues/572

/cc @slackpad @skippy

### DIFF
--- a/dependency/catalog_node.go
+++ b/dependency/catalog_node.go
@@ -128,7 +128,7 @@ func (d *CatalogNode) Fetch(clients *ClientSet, opts *QueryOptions) (interface{}
 
 // CanShare returns a boolean if this dependency is shareable.
 func (d *CatalogNode) CanShare() bool {
-	return true
+	return false
 }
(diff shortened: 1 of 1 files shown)


slackpad:
@sethvargo I'm a little confused by this one - why wouldn't we be able to share a Node instance, or does this somehow not point at a particular node and cause confusion?

sethvargo:
Hey @slackpad 

If I understand it correctly, the way "dedupe" works is that it renders on one node and pushes the result into the KV store (except for things that can't be shared). Then, each node only watches that one KV path and any unsharable deps instead. 

By caching the "node" (which is "self") of the first template that happens to write to the KV, we are writing the _wrong_ template on all other machines. Additionally, I think this is basically causing a hot loop while the template processes fight for a leader, since they will never agree. Does that make sense?

slackpad:
Ah ok I get it now - i'll look at this wrt to #608 then.

Issue #623, opened 2016-05-02
Parse file permissions as a String in JSON


Ginja:
9475d6f doesn't seem to fix #548. I'm still unable to specify `perms` as a String in JSON.

```bash
$ /opt/consul-template/0.14.0/consul-template --version
consul-template v0.14.0

$ cat /etc/consul-template/conf.d/someconfig.json
{
  "template": {
    "command": "somecommand",
    "destination": "/destination/path",
    "perms": "0600",
    "source": "/template/path"
  }
}

$ /opt/consul-template/0.14.0/consul-template -config /etc/consul-template/conf.d
Consul Template returned errors:
config: walk error: 1 error(s) occurred:

* 1 error(s) decoding:

* 'template[0].perms' expected type 'os.FileMode', got unconvertible type 'string'
```

If you need more information, or would like me to test anything please let me know :)

sethvargo:
Hey @Ginja 

I just put together #626 which has a bunch of test cases. Could you please try it out?

Ginja:
Hey @sethvargo,

I tried compiling consul-template from your PR branch, but ran into some issues. `govendor fetch -v +outside` seems to stall out after fetching a couple of packages. Tried working around it, but was unable to produce a binary with `make dev`. I'll test it out when a new release is published. 

Thanks!

sethvargo:
Hi @Ginja 

With the new vendored deps, just run `make dev` on Go 1.6+ and it'll "just work". No need to update deps.

PR #626, opened 2016-05-03
Use mapstructure decoder for os.FileMode

sethvargo:
This commit adds a new mapstructure function for automatically decoding into os.FileMode. The function expects the value to be a string in json or a string or int in hcl.

Fixes https://github.com/hashicorp/consul-template/issues/623

/cc @slackpad @cbednarski

### DIFF
--- a/mapstructure.go
+++ b/mapstructure.go
@@ -1,0 +1,33 @@
+package main
+
+import (
+	"os"
+	"reflect"
+	"strconv"
+
+	"github.com/mitchellh/mapstructure"
+)
+
+// StringToFileMode returns a function that converts strings to os.FileMode
+// value. This is designed to be used with mapstructure for parsing out a
+// filemode value.
+func StringToFileMode() mapstructure.DecodeHookFunc {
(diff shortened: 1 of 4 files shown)


slackpad:
LGTM

mapstructure.go
```diff
@@ -0,0 +1,45 @@
...
+			// Assume the user didn't set the special bit. Some users just put perms
+			// as "600", but they really meant "0600".
+			typed = string(append([]byte("0"), []byte(typed)...))
+		case 4:
+		default:
+			return data, fmt.Errorf("invalid filemode format: %q", typed)
+		}
+
+		// Convert it by parsing
+		v, err := strconv.ParseUint(typed, 8, 32)
```
sean-: Can you change this to `strconv.ParseUint(typed, 8, 12)` instead?  We don't need the other 20 bits when parsing the mode.  If you do that, you can also pass `data` directly to `ParseUint` without prepending a leading "0" because `ParseUint` will do the right thing.

PR #628, opened 2016-05-08
Remove the reaping functionality

sethvargo:
Just as we did with the official Consul container, we are going to rely on services like dumb-init to handle child process reaping, not the go process itself.

/cc @slackpad


slackpad:
LGTM - this will need some all caps in the change log :-)

Issue #635, opened 2016-05-13
Add support for maps with in/contains?


bradwilson:
We are attempting to use a key's value as JSON (via `parseJSON`), and want to be able to test for values inside the map, because we don't want to have to force all keys to have all legal property values in the JSON.

There appears to be no way to do this today, because `in` does not support maps (and the traditional method of testing for map membership in Go is not supported in consul-template). It would be nice if `in` (and thus `contains`) could implement support for maps, where the thing you're looking for in the map is the key name.

Example value:

```json
{ "tcp_port": 12345 }
```

We'd like to write something like:

```
{{ $value = .Value | parseJSON }}
{{ if in $value "tcp_port" }}
...
{{ end }}
```

Thoughts? Is there a better way to do this?

bradwilson:
BTW, I saw other issues related to people wanting to do things like this with tags, but I'd much prefer to keep these things in the KV system, as we're already using consul-template against KVs (not services or nodes).

sethvargo:
Hi @bradwilson 

Thank you for opening an issue. Please take a look at Go's [templating language](https://golang.org/pkg/text/template/). I believe you want to use the `index` function here. I just tested this:

```go
{{with key "foo" | parseJSON}}
{{index . "key"}}
{{end}}
```

on Consul Template 0.14. Is that not working for you?

bradwilson:
That does not appear to work. I tried:

```
{{ if index $value "tcp_port" }}
```

Then consul-template says:

```
executing "test.ctmpl" at <index $value "tcp_po...>: error calling index: cannot index slice/array with type string
```

Is that the intended usage?

sethvargo:
Hi @bradwilson 

Can you share your complete template? I ran that exact template I posted above and it worked without issue. What is `$value` in this context?

bradwilson:
@sethvargo:

> Can you share your complete template?

This is a simple repro template:

```
{{ range ls "published_services" }}
  {{ $serviceName := .Key }}
  {{ $value := .Value | parseJSON }}
  {{ if index $value "tcp_port" }}
    {{ $serviceName }} is on TCP port {{ $value.tcp_port }}
  {{ else }}
    {{ $serviceName }} is NOT using TCP
  {{ end }}
{{ end }}
```

To test, I have two keys in K/V:

* Key `published_services/foo` with a value of `{ "tcp_port": 12345 }`
* Key `published_services/bar` with an empty value

I expect to see:

```
foo is on TCP port 12345
[...]

PR #637, opened 2016-05-16
Return a map instead of an array for parseJSON

sethvargo:
The default function for parseJSON currently returns an array. This is problematic.

Fixes GH-635 because it makes using functions like `index` more difficult due to arrays being unable to have string indexes. This should be backwards compatible because one can still index a map using a string (thus treating it as an array).

/cc @slackpad


slackpad:
Code LGTM - there's one broken test. It would be good to also add an integer-indexing test as a sanity check.

Issue #660, opened 2016-06-21
Error using write based vault secret backends


willusher:
Commit GH-652 added support for write-based vault secret backends, but I am unable to get this to work. I am using the latest release (0.15.0).

When I use the example given in the documentation:
```liquid
{{ with secret "pki/issue/my-domain-dot-com" "common_name=foo.example.com" }}
{{ .Data.certificate }}
{{ end }}
```

I receive the error:
```
Consul Template returned errors:
template: template: vault.pem.ctmpl:1:7: executing "vault.pem.ctmpl" at <secret>: \
wrong number of args for secret: want 1 got 2
```

When I tried the original commit from @enkaskal [here](https://github.com/enkaskal/consul-template/commit/8e166e9a040252d6afb30839939fe0bec8a53fd9), it was successful. 

I suspect it is related to the modified function arguments [here](https://github.com/hashicorp/consul-template/pull/652/files#diff-6c8acac18c3cbac95194c2d445b084f0R162), compared to @enkaskal's commit using the original args and splitting [inside the function](https://github.com/enkaskal/consul-template/commit/8e166e9a040252d6afb30839939fe0bec8a53fd9#diff-6c8acac18c3cbac95194c2d445b084f0R164).

PR #662, opened 2016-06-21
Allow variadic template function for secret

sethvargo:
Fixes GH-660

/cc @slackpad

### DIFF
--- a/template_functions.go
+++ b/template_functions.go
@@ -210,15 +210,15 @@ func nodesFunc(brain *Brain,
 
 // secretFunc returns or accumulates secret dependencies from Vault.
 func secretFunc(brain *Brain,
-	used, missing map[string]dep.Dependency) func(string) (*dep.Secret, error) {
-	return func(s string) (*dep.Secret, error) {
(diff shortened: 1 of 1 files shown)


willusher:
@sethvargo - any reason you're holding off on merging this? Would be much appreciated, thanks.

sethvargo:
I'm waiting on a 👍 from @slackpad 😄

template_functions.go
```diff
@@ -210,15 +210,15 @@ func nodesFunc(brain *Brain,
 
 // secretFunc returns or accumulates secret dependencies from Vault.
 func secretFunc(brain *Brain,
```
slackpad: Do we need to do the same for `secretsFunc()` below? Not sure how that relates to this one.

slackpad:
Gah sorry I missed this. @sethvargo one question on here.

template_functions.go
sethvargo: No, `secrets` doesn't currently accept multiple arguments

slackpad:
👍

willusher:
Thank you @sethvargo and @slackpad !

PR #669, opened 2016-07-01
Allow commas in the prefix.

steveprentice:
Spring Cloud Consul has adopted a default prefix that includes a comma
to separate the application name and profile. Prefixes end up looking like
'config/application,profile/'. This commit adds support for these prefixes.

See: http://cloud.spring.io/spring-cloud-consul/spring-cloud-consul.html

```
config/testApp,dev/
config/testApp/
config/application,dev/
config/application/
```

### DIFF
--- a/dependency/store_key_prefix.go
+++ b/dependency/store_key_prefix.go
@@ -142,7 +142,7 @@ func (d *StoreKeyPrefix) Stop() {
 func ParseStoreKeyPrefix(s string) (*StoreKeyPrefix, error) {
 	// a(/b(/c))(@datacenter)
 	re := regexp.MustCompile(`\A` +
-		`(?P<prefix>[[:word:]\.\:\-\/]+)?` +
+		`(?P<prefix>[[:word:],\.\:\-\/]+)?` +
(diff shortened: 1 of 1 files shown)


sethvargo:
@slackpad can you comment on whether prefixes as commas are recommended? I know we strongly suggest users do not use dots, and I think commas might be the same.

steveprentice:
Should consul-template be opinionated about this? If consul accepts it as a folder name, as a user, I would expect that consul-template would accept it as well.

slackpad:
For the KV store it's probably ok since Consul makes minimal assumptions there are allows the separator to be passed in when querying. The comma isn't used for service names, right? I'd strongly recommend not using it there in order to not break DNS.

steveprentice:
Spring Cloud Consul doesn't use a comma in the application/service name – just as the default separator between the application name and the profile. I don't know their motivation behind using a comma, but my guess would be they needed a character that wasn't allowed in the application/service name but was allowed in the name of a folder in consul.

steveprentice:
Thanks!

PR #671, opened 2016-07-11
Add exec mode

sethvargo:
The README explains in more detail, but this PR adds support for what I'm calling "Exec Mode"™️ . In this mode, Consul Template behaves more akin to Envconsul in that it spawns and "manages" a subprocess as a faux-supervisor.

This enables more seamless integration with schedulers by proxying signals to the managed child process.

/cc @armon @dadgar


sethvargo:
The tests are probably going to fail because of missing deps, but I wanted to get review in first before adding 10000000000000000000000000000000000000000 lines of code.

README.md
```diff
@@ -246,6 +261,43 @@ deduplicate {
...
+  // or killing the child process. This can be used to prevent the thundering
+  // herd problem on applications that do not gracefully reload.
+  splay = "5s"
+
+  // This defines the signal that will be sent to the child process when a
+  // change occurs in a watched template. The signal will only be sent after
+  // the process is started, and the process will only be started after all
+  // dependent templates have been rendered at least once. The default value
+  // is "SIGHUP".
+  reload_signal = "SIGUSR1"
```
armon: How do you disable reload behavior? e.g. for apps that need to be restarted on change

sethvargo: I think you could just set it to a signal that the app doesn't respond to for now. I didn't implement forced-restart. Consul Template just sends the signal to the child process; it does not wait for the child process to respond or take action on that change.

README.md
```diff
@@ -1114,6 +1166,56 @@ func main() {
...
+- The child process must remain in the foreground. This is a requirement for Consul Template to manage the process and send signals.
+- The exec command will only start after _all_ templates have been rendered at least once. One may have multiple templates for a single Consul Template process, all of which must be rendered before the process starts. Consider something like an nginx or apache configuration where both the process configuration file and individual site configuration must be written in order for the service to successfully start.
+- After the child process is started, any change to any dependent template will cause the reload signal to be sent to the child process. This reload signal defaults to SIGHUP, but can be customizing via the CLI or configuration file.
+- When Consul Template is stopped gracefully, it will send the configurable kill signal to the child process. The default value is SIGTERM, but it can be customized via the CLI or configuration file.
+- It is not possible to have more than one exec command (although each template can still have its own reload command).
```
jippi: so using one CT instance with multiple `template {}` instances would need to move to one CT instance per `exec` 

it would be kinda nice to be able to have multiple `exec` per instance, and then attach an `exec` to a `template` a bit like puppet `notify`/`listen` behavior

sethvargo: Yes, and we discussed this a lot. It doesn't make sense to have Consul Template spawn multiple subprocesses; that's not the way any modern process supervisor works. If you want this behavior, you need a 1-1 mapping of Consul Template to a process, which is the Docker-recommended approach as well (one app per container). This is also not designed to replace the existing `command` arguments.

README.md
armon: I think we should support the "restart" value, since most apps can't do graceful reloads

signals.go
```diff
@@ -0,0 +1,30 @@
+package main
```
armon: This signals stuff is pretty useful. I wonder if its worth moving into a package in a shared repo?

runner.go
```diff
@@ -160,12 +172,37 @@ func (r *Runner) Start() {
...
+				// was spawned, so we need to watch a new exitCh.
+				if nexitCh != nil {
+					childExitCh = nexitCh
+				}
+			}
+		}
+
+		// If we are in exec mode and all templates have been rendered, spawn the
+		// child process.
+		if r.config.Exec.Command != "" && r.allTemplatesRendered() {
```
armon: This looks like a dangling block?

runner.go
```diff
@@ -741,6 +822,125 @@ func (r *Runner) deletePid() error {
...
+	if r.config.Exec.Splay > 0 {
+		ns := r.config.Exec.Splay.Nanoseconds()
+		offset := rand.Int63n(ns)
+
+		log.Printf("[DEBUG] (runner) waiting %.2fs for random splay",
+			time.Duration(offset).Seconds())
+
+		select {
+		case <-time.After(time.Duration(offset)):
+		case <-r.DoneCh:
```
armon: Should we return early if we see the done ch?

runner.go
```diff
@@ -741,6 +822,125 @@ func (r *Runner) deletePid() error {
...
```
armon: Same? Early exit?

runner.go
```diff
@@ -219,6 +256,11 @@ func (r *Runner) Start() {
 			log.Printf("[INFO] (runner) received template %q from quiescence", tmpl.Path)
 			delete(r.quiescenceMap, tmpl.Path)
 
+		case c := <-childExitCh:
+			log.Printf("[INFO] (runner) child process died")
+			r.ErrCh <- fmt.Errorf("child process died with exit code %d", c)
```
armon: Should we suppress error if `ExitCodeOK`?

signals.go
sethvargo: Thought about that. I want to get this out the door, but I have a TODO to make this more general and it's own package to share with envconsul and consul-replicate in the future.

README.md
```diff
@@ -1114,6 +1166,56 @@ func main() {
 
 Caveats
 -------
+### Exec Mode
+
+As of version 0.16.0, Consul Template has the ability to maintain an arbitrary child process (similar to [envconsul](https://github.com/hashicorp/envconsul)). This mode is most beneficial when running Consul Template in a container or cgroup, particularly on a scheduler like [Nomad](https://www.nomadproject.io) or Kubernetes. When activated, Consul Template will spawn and manage the lifecycle of the child process.
```
dadgar: I would remove the bit on `or cgroup` and replace with `or on a cluster scheduler like...`

README.md
```diff
@@ -1114,6 +1166,56 @@ func main() {
...
+```bash
+#!/usr/bin/env bash
+exec /bin/my-server -config /tmp/server.conf > /var/log/my-server.log
+```
+
+There are some additional caveats with Exec Mode, which should be considered carefully before use:
+
+- If the child process dies, the Consul Template process will also die. Consul Template **does not supervise the process!** This is generally the responsibility of the scheduler or init system.
+- The child process must remain in the foreground. This is a requirement for Consul Template to manage the process and send signals.
+- The exec command will only start after _all_ templates have been rendered at least once. One may have multiple templates for a single Consul Template process, all of which must be rendered before the process starts. Consider something like an nginx or apache configuration where both the process configuration file and individual site configuration must be written in order for the service to successfully start.
```
dadgar: did you mean to do _aLL_?

runner.go
```diff
@@ -741,6 +822,125 @@ func (r *Runner) deletePid() error {
...
+		log.Printf("[DEBUG] (runner) waiting %.2fs for random splay",
+			time.Duration(offset).Seconds())
+
+		select {
+		case <-time.After(time.Duration(offset)):
+		case <-r.DoneCh:
+		}
+	}
+
+	if err := r.cmd.Process.Signal(r.config.Exec.KillSignal); err == nil {
```
dadgar: So on windows the only supported signal you can send is a SIGKILL. Not sure if you want to guard against that. Probably when you support restart mode, only enable that on windows?

README.md
sethvargo: It's lowercase (copy-paste into an editor). GitHub's new font is just awful at italic apparently 😦

runner.go
sethvargo: I don't think so - the assumption is that the child process continues to run in the foreground. If the process stops (even with exit 0 and success), it's no longer under management and therefore Consul Template cannot send signals and manage that processes' lifecycle.

child/child.go
```diff
@@ -0,0 +1,345 @@
...
+
+	splay time.Duration
+
+	// cmd is the actual child process under management.
+	cmd *exec.Cmd
+
+	// exitCh is the channel where the processes exit will be returned.
+	exitCh chan int
+
+	// stopLock is the mutext to lock when stopping. stopCh is the circuit breaker
```
slackpad: _mutex_ :-)

child/child.go
```diff
@@ -0,0 +1,345 @@
...
+	// prevents multiple processes from all signaling at the same time. This value
+	// may be zero (which disables the splay entirely).
+	Splay time.Duration
+}
+
+// New creates a new child process for management with high-level APIs for
+// sending signals to the child process, restarting the child process, and
+// gracefully terminating the child process.
+func New(i *NewInput) (*Child, error) {
+	if i == nil {
```
slackpad: If they don't specify anything, the check that follows will always fail, so you can delete this or just make sure `i` isn't `nil`.

child/child.go
```diff
@@ -0,0 +1,345 @@
...
+		defer c.RUnlock()
+
+		return c.reload()
+	}
+}
+
+// Kill sends the kill signal to the child process and waits for successful
+// termination. If no kill signal is defined, the process is killed with the
+// most aggressive kill signal. If the process does not gracefully stop within
+// the provided KillTimeout, the process is force-killed. If a splay wa
```
slackpad: _was_

child/child.go
slackpad: Actually the way you have this is better than a `nil` check - nm.

runner_test.go
```diff
@@ -1353,3 +1353,185 @@ func TestRunner_dedup(t *testing.T) {
...
+	defer test.DeleteTempfile(tmpl, t)
+
+	// Create a tiny bash script for us to run as a "program"
+	script := test.CreateTempfile([]byte(strings.TrimSpace(fmt.Sprintf(`
+#!/usr/bin/env bash
+trap "echo 'one' >> %s" SIGUSR1
+trap "echo 'two' >> %s" SIGUSR2
+
+while true; do
+	: # Hang
```
slackpad: Might be good to put a short sleep in here so it doesn't burn CPU.

slackpad:
Took a look through and noted a few trivial things - LGTM!

Issue #656, opened 2016-06-09
toTOML helper function


siddharthist:
Since we already have `toYAML` and `toJSON`, I think this is a natural next step!

sethvargo:
Hi @siddharthist 

I would gladly review a pull request for this.

PR #675, opened 2016-07-26
template functions: toTOML

siddharthist:
Fits in nicely with `toJSON` and `toYAML`. My own motivation is that I would like to configure [Træfɪk](http://traefik.io/) entirely with Consul-Template, as part of [Mantl](http://mantl.io).

Unfortunately, Burntsushi/toml doesn't support the `Marshal` interface (see https://github.com/BurntSushi/toml/issues/76), so this is slightly more awkward than `toYAML` and `toJSON`. 

Fixes (the already closed) #656.

### DIFF
--- a/template_functions.go
+++ b/template_functions.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"os"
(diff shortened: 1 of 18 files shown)


siddharthist:
@sethvargo Thanks for merging!

Issue #683, opened 2016-08-08
Maintain template order


sethvargo:
We currently use a map, but that makes template order random. We should preserve template ordering.

n2taylor:
Hi, can you explain the expected behavior here when we pass in a directory for config, containing a series of config files? I'm still seeing occasional instances where our templates are rendered out of order (based on filename). e.g. passing in a directory containing `a.cfg`, `b.cfg`, `c.cfg` leading to rendered files with timestamps suggesting they were rendered in the order "b", "a", "c".

sethvargo:
Hi @n2taylor 

We generally do not triage old issues. If you think this is a bug, please open a new issue and supply your CT configuration and output please. Thanks! 😄

PR #684, opened 2016-08-08
Maintain template-specified order

sethvargo:
Previously order was not guaranteed because a map was used to ensure
uniqueness. This commit changes the behavior to maintain order, so the
order templates are specified in config are the order they are run.

Fixes GH-683


slackpad:
LGTM

PR #690, opened 2016-08-13
Add key_or_ignore

bennetimo:
This function works identically to `key`, except if the key is found no log will be generated.

The use case (for me) is templating config override files. In the case that a key is found, I would like to use it for setting a value. If the key is not found, I do not want to template anything (to fallback to the default), and wish to avoid logging a warning so as not to dilute warnings for mandatory keys.

With this function, I can achieve that with:
```
{{ with key_or_ignore "/mykey" }}{{printf "setting = %s" .}}{{end}}
``` 

This relates to GH-368


sethvargo:
Hi @bennetimo 

You can use `key_or_default` for this, and set the default value to the empty string or false.

bennetimo:
@sethvargo ah yes I have realised the same, I overlooked that key_or_default is not generating a warning :)

This suits fine:
```
{{ with key_or_default "/mykey" "" }}{{printf "setting = %s" .}}{{end}}
```

Thanks for your work on this project :)

PR #707, opened 2016-08-31
Add dockerized build environment

yanndegat:


sethvargo:
Hi there,

Thank you for the Pull Request. We do not put third-party build info in our repositories, so we will not be accepting this. You are more than welcome to use Docker as your build environment, but we will not be adding a Dockerfile or make target for this.

PR #710, opened 2016-09-15
NGINX example in Consul Template Readme

kunalvjti:
Adding NGINX as an example too in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -1323,7 +1323,7 @@ $ consul-template \
   -dry
 ```
 
-Depending on the state of the demo Consul instance and assuming that the only service registered with Consul here is the one which needs to be load balanced by NGINX, you could see the following output:
+Depending on the state of the demo Consul instance, you could see the following output:
 
 ```text
 upstream service {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1283,6 +1283,75 @@ This will still add the dependency to the list of watches, but Go will not evalu
...
+Save this file to disk as `nginx.ctmpl` and  run the `consul-template` daemon:
+
+```shell
+$ consul-template \
+  -consul demo.consul.io \
+  -template nginx.ctmpl:/etc/nginx/conf.d/default.conf
+  -dry
+```
+
+Depending on the state of the demo Consul instance and assuming that the only service registered with Consul here is the one which needs to be load balanced by NGINX, you could see the following output:
```
sethvargo: This part seems better suited for the Consul docs, since it doesn't involve Consul Template. Could you please update? Thanks!

kunalvjti: Done. Thanks

sethvargo:
Merged in 7886c36

PR #724, opened 2016-09-22
Refactor code such that consul-template can be imported by other tools

dadgar:


config/config_testing.go
```diff
@@ -0,0 +1,25 @@
+package config
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+func TestConfig(contents string, t *testing.T) *Config {
```
sethvargo: This is fine for now, but I'd like to refactor this at some point to be easier to create a config...or not need to write to the FS

PR #725, opened 2016-09-26
Add ability to watch for a template being rendered and all templates rendered

dadgar:
This PR adds two channels, one for when a template has been rendered and one for when all templates have been rendered.

### DIFF
--- a/manager/runner.go
+++ b/manager/runner.go
@@ -61,10 +61,12 @@ type Runner struct {
 	// templates is the list of calculated templates.
 	templates []*template.Template
 
-	// renderedTemplates is a map of templates we have successfully rendered to
-	// disk. It is used for once mode and internal tracking. The key is the Path
-	// of the template.
-	renderedTemplates map[string]struct{}
+	// renderedTimes is a mapping of the destination of a template to the time
+	// it was rendered.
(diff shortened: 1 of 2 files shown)


manager/runner.go
```diff
@@ -513,6 +534,12 @@ func (r *Runner) Run() error {
 				// Record that at least one template was rendered.
 				renderedAny = true
 
+				// Send the signal that the template got rendered
+				select {
+				case r.renderedCh <- tmpl.Path:
+				default:
```
dadgar: We don't want consul-template to be blocked by the listener. Though I had a decent idea. We should add a map[string]time.Time and expose it. So that if the caller doesn't keep up they can get the map and see what they have missed

sethvargo: Hmm yea - I'm just worried that there's a race where the channel is full and the listener never receives that message, even though it's reading from the channel.

dadgar: Yeah I think the access pattern will be select on channel, get map of updates. I am going to change the channel only to be one buffered and make it a chan struct{} so that is the only valid way to access it. That will solve this problem

dadgar: Done

sethvargo:
LGTM

PR #729, opened 2016-09-28
Embedded templates in config

dadgar:


README.md
```diff
@@ -301,14 +301,21 @@ exec {
...
   // This is the destination path on disk where the source template will render.
   // If the parent directories do not exist, Consul Template will attempt to
   // create them.
   destination = "/path/on/disk/where/template/will/render.txt"
 
+  // This option allows embedding a template in the configuration file rather
+  // then supplying the `source` path to the template file. This is useful for
+  // short templates. This option can is mutually exclusive with the `source`
+  // option.
+  embedded_template = "{{key_or_default \"service/redis/maxconns@east-aws\" \"5\"}}"
```
sethvargo: Kinda 🚲 🏠 but I wonder if there will be confusion that we call it "source" above and "template" here. I think the word "template" is overloaded in "consul template"

sethvargo: I wonder if "contents" might make more sense

template/template.go
```diff
@@ -87,14 +103,16 @@ func (t *Template) Execute(brain *Brain) ([]dep.Dependency, []dep.Dependency, []
 // init reads the template file and initializes required variables.
 func (t *Template) init() error {
 	// Render the template
-	contents, err := ioutil.ReadFile(t.Path)
-	if err != nil {
-		return err
+	if t.Contents == "" {
```
sethvargo: Nit-picky, but I'd prefer to reverse this logic:

```go
if t.Path != "" {
  // ...
}
```

template/template.go
```diff
@@ -30,9 +32,18 @@ type Template struct {
 // path. If the template does not exist, an error is returned. During
 // initialization, the template is read and is parsed for dependencies. Any
 // errors that occur are returned.
-func NewTemplate(path, leftDelim, rightDelim string) (*Template, error) {
+func NewTemplate(path, contents, leftDelim, rightDelim string) (*Template, error) {
```
sethvargo: I don't like this API anymore (I think it's confusing to have an input that is both a "path" and "contents"). I think NewTemplate should just take the contents as a string and we push the file reading up the stack. What do you think?

README.md
```diff
@@ -301,14 +301,21 @@ exec {
...
   source = "/path/on/disk/to/template.ctmpl"
 
   // This is the destination path on disk where the source template will render.
   // If the parent directories do not exist, Consul Template will attempt to
   // create them.
   destination = "/path/on/disk/where/template/will/render.txt"
 
+  // This option allows embedding a template in the configuration file rather
+  // then supplying the `source` path to the template file. This is useful for
+  // short templates. This option can is mutually exclusive with the `source`
```
nugend: This option <strike>can</strike> is

README.md
dadgar: I am fine with `contents` as the map structure name.

README.md
dadgar: Nice catch

template/template.go
dadgar: I would actually keep the path as the `ID` method relies on it to disambiguate two templates with the same content

dadgar: You kinda need the Path for de-dupping the same template in the `ID()` method as well

README.md
dadgar: Done

template/template.go
dadgar: Done

PR #730, opened 2016-09-29
Vault auth with certificate

carlpett:
This PR adds the possibility of authenticating with Vault using the TLS Certificate auth backend. It simply piggybacks on the existing SSL config block and tries to auth with the set certificate if there is no token set and a client certificate is configured.

One potential improvement would be to allow a configurable auth backend path (currently hardcoded to `auth/cert/login`). Would that be valuable?

### DIFF
--- a/dependency/client_set.go
+++ b/dependency/client_set.go
@@ -235,6 +235,24 @@ func (c *ClientSet) CreateVaultClient(i *CreateVaultClientInput) error {
 	if i.Token != "" {
 		log.Printf("[DEBUG] (clients) setting vault token")
 		client.SetToken(i.Token)
+	} else if i.SSLCert != "" {
+		log.Printf("[DEBUG] (clients) attempting to authenticate with certificate")
+
+		secret, err := client.Logical().Write("auth/cert/login", nil)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


sethvargo:
HI @carlpett 

Thank you for the Pull Request. The general sentiment with these tools is that all communication requires a VAULT_TOKEN, and you can request that token out of band. We are not yet ready to open the door for letting these tools perform actual authentication, because there are a lot of UX and consistency things we need to think through first.


Thanks and sorry!

carlpett:
Hm, I see. Where would be a good place to discuss this? We run consul-template as a systemd service on _a lot_ of machines, and getting that token into the environment for consul-template and especially making sure it never expires (and if so provision a new one) is really hard.

pvandervelde:
@carlpett Have you ever figured out how to solve getting the token into the environment. We too run Consul-Template as a systemd service and are experiencing the same issue you described :(

carlpett:
@pvandervelde I don't recall the specifics any more, but I _think_ we ended up building some other binary with "real" vault integration to wrap the call to consul-template. Not very pretty for sure.
Perhaps minds have changed in the ~3 years since this was closed and you could propose it again? I'm not working with these tools daily anymore, I'm afraid.

PR #734, opened 2016-10-03
Remove AllRendered and push the logic to the user and expose more detail

dadgar:
This PR address the fact that the destination is not always needed (in dry mode) and accounts for the fact that wouldRender is triggered even if the content is empty, this makes the old AllRendered implementation broken.

This current implementation gives the caller of the `Runner` all the information they need to track it themselves

### DIFF
--- a/manager/runner.go
+++ b/manager/runner.go
@@ -54,25 +54,18 @@ type Runner struct {
 	outStream, errStream io.Writer
 	inStream             io.Reader
 
-	// ctemplatesMap is a map of each template to the ConfigTemplates
+	// ctemplatesMap is a map of each template ID to the ConfigTemplates
 	// that made it.
 	ctemplatesMap map[string][]*config.ConfigTemplate
 
 	// templates is the list of calculated templates.
 	templates []*template.Template
 
-	// renderedTimes is a mapping of the destination of a template to the time
-	// it was rendered.
-	renderedTimes map[string]time.Time
+	// renderEvents is a mapping of a template ID to the render event.
+	renderEvents map[string]*RenderEvent
 
-	// renderedTimesLock protects access into the renderedTimes slice
-	renderedTimesLock sync.RWMutex
-
-	// allRenderedCh is used to signal that all templates have been rendered
-	allRenderedCh chan struct{}
-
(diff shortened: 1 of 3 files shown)


manager/runner.go
```diff
@@ -750,23 +733,55 @@ func (r *Runner) configTemplatesFor(tmpl *template.Template) []*config.ConfigTem
...
+// stores it as would have been rendered.
+func (r *Runner) markRenderTime(tmplID string, didRender bool) {
+	r.renderEventsLock.Lock()
+	event, ok := r.renderEvents[tmplID]
+	if !ok {
+		event = &RenderEvent{}
+		r.renderEvents[tmplID] = event
+	}
+	if didRender {
+		event.LastDidRender = time.Now()
```
sethvargo: This is super picky, but I think there's a tiny race here where these two `time.Now()`s are actually different times. Could we grab the current time right after we acquire the lock, just to be safe?

manager/runner.go
```diff
@@ -750,23 +733,55 @@ func (r *Runner) configTemplatesFor(tmpl *template.Template) []*config.ConfigTem
...
+	} else {
+		event.LastWouldRender = time.Now()
+	}
+	r.renderEventsLock.Unlock()
```
sethvargo: Is there a reason you did this here instead of using a `defer` like above? Just curious, but I'd prefer to be consistent.

manager/runner.go
```diff
@@ -330,26 +333,19 @@ func (r *Runner) Stop() {
...
-func (r *Runner) RenderedTimes() map[string]time.Time {
-	r.renderedTimesLock.RLock()
-	defer r.renderedTimesLock.RUnlock()
-	times := make(map[string]time.Time, len(r.renderedTimes))
-	for k, v := range r.renderedTimes {
+// RenderEvents returns the render events for each template was rendered. The
+// map is keyed by template ID.
+func (r *Runner) RenderEvents() map[string]*RenderEvent {
+	r.renderEventsLock.RLock()
+	defer r.renderEventsLock.RUnlock()
```
sethvargo: Newline after this plz

PR #735, opened 2016-10-05
Allow reading configs directly from memory

sethvargo:


jippi:
YAY, step closer to nomad awesome sauce? :)

PR #736, opened 2016-10-05
Issue a blocking query for "key"

sethvargo:
This fixes a commonly confusing issue where the `key` function does not actually issue a blocking query. This is because users have, in the past, used the API like this:

    {{ if key "foo" }}
      feature flag active
    {{ else }}
      default
    {{ end }}

In this model, the existence of a key is the check. With the introduction of `key_or_default`, it makes sense to return `key` to be a blocking query.

/cc @dadgar @slackpad


slackpad:
LGTM

PR #743, opened 2016-10-10
Adding first-class support for an explicit TLS server name for Vault/Consul

stevehorsfield:
Vault and Consul provide options to allow the specification of an expected server name in a certificate that is different to the provided address.

When using `https://vault.service.consul` and `https://consul.service.consul`, for example, (not to mention variants of these), using a fixed name creates issues with redirects (Vault), public certificate authority certificates (due to private naming) and private-IP-based certificate alternate names. Vault provides the `VAULT_TLS_SERVER_NAME` environment variable and Consul provides the `server_name` configuration option to address these scenarios.

Partial support for `VAULT_TLS_SERVER_NAME` was added implicitly by upgrading the Vault client. This PR extends support to allow values to be placed in configuration and to push `VAULT_TLS_SERVER_NAME` into child process environments.

We are currently working on implementing full TLS features in our own environment and so do not yet have a way to effectively test this PR ourselves. Please review, provide feedback and merge when possible.

Thanks,

Steve

### DIFF
--- a/dependency/client_set.go
+++ b/dependency/client_set.go
@@ -47,6 +47,7 @@ type CreateConsulClientInput struct {
 	SSLCert      string
 	SSLKey       string
 	SSLCACert    string
+	ServerName   string
 }
 
 // CreateVaultClientInput is used as input to the CreateVaultClient function.
@@ -59,6 +60,7 @@ type CreateVaultClientInput struct {
 	SSLCert     string
 	SSLKey      string
 	SSLCACert   string
+	ServerName  string
 }
 
 // NewClientSet creates a new client set that is ready to accept clients.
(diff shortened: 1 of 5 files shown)


dependency/client_set.go
```diff
@@ -135,7 +137,11 @@ func (c *ClientSet) CreateConsulClient(i *CreateConsulClientInput) error {
 		tlsConfig.BuildNameToCertificate()
 
 		// SSL verification
-		if !i.SSLVerify {
+		if i.ServerName != "" {
+			tlsConfig.ServerName = i.ServerName
+			tlsConfig.InsecureSkipVerify = false
+			log.Printf("[DEBUG] (clients) using explicit consul TLS server host name: %s", tlsConfig.ServerName)
+		} else if !i.SSLVerify {
```
sethvargo: I think this logic (and the logic on 222) could lead to some subtle challenges. Basically if you set a server name, you can _never_ disable SSL verification. In essence this breaks the ability to have a later configuration that is merged (remember CT can merge multiple configs) that would disable SSL verification.

A use case I'm thinking is that a company has a default "base" config, but then they disable SSL verification when running in local development. As written, this change would require them to both disable SSL verification AND set the server_name to `""`.

sethvargo:
808fbaf

stevehorsfield:
Thanks for pulling this through, @sethvargo . For others interested, see #746.

PR #746, opened 2016-10-10
Add first-class support for explicit TLS server name

sethvargo:


stevehorsfield:
@sethvargo - Thanks for working on this. I was expecting to come back to the review comments but had been pulled into other workstreams temporarily.

PR #764, opened 2016-10-24
Add CAPath config to SSL and use when creating Consul/Vault clients

dadgar:
This PR also fixes the Vault environment variables set during exec

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,6 +230,10 @@ ssl {
   // useful for self-signed certificates or for organizations using their own
   // internal certificate authority.
   ca_cert = "/path/to/ca"
+
+  // This is the path to a directory of PEM-encoded CA cert files. If both
+  // `ca_cert` and `ca_path` is specified, `ca_cert` is preferred.
+  ca_path = "path/to/certs/"
 }
 
 // This block defines the configuration for connecting to a syslog server for
--- a/dependency/client_set.go
+++ b/dependency/client_set.go
@@ -2,15 +2,14 @@ package dependency
 
 import (
 	"crypto/tls"
-	"crypto/x509"
 	"fmt"
-	"io/ioutil"
 	"log"
 	"net/http"
 	"sync"
 
 	consulapi "github.com/hashicorp/consul/api"
 	"github.com/hashicorp/go-cleanhttp"
+	rootcerts "github.com/hashicorp/go-rootcerts"
 	vaultapi "github.com/hashicorp/vault/api"
 )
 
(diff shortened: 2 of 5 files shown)


dependency/client_set.go
```diff
@@ -205,15 +205,14 @@ func (c *ClientSet) CreateVaultClient(i *CreateVaultClientInput) error {
...
-			if err != nil {
-				return fmt.Errorf("client set: vault: %s", err)
+		if i.SSLCACert != "" || i.SSLCAPath != "" {
+			rootConfig := &rootcerts.Config{
+				CAFile: i.SSLCACert,
+				CAPath: i.SSLCAPath,
+			}
+			if err := rootcerts.ConfigureTLS(&tlsConfig, rootConfig); err != nil {
+				return fmt.Errorf("client set: vault configuring TLS failed: %s", err)
 			}
```
dadgar: @sethvargo I am not sure if you want to change it but you can see how Nomad creates a Vault client. It just creates a default vault client config and then builds the TLS struct and calls this: https://github.com/hashicorp/vault/blob/master/api/client.go#L108

Makes it so that all the config parsing happens in Vault api/ and not the caller

README.md
```diff
@@ -230,6 +230,10 @@ ssl {
   // useful for self-signed certificates or for organizations using their own
   // internal certificate authority.
   ca_cert = "/path/to/ca"
+
+  // This is the path to a directory of PEM-encoded CA cert files. If both
+  // `ca_cert` and `ca_path` is specified, `ca_cert` is preferred.
+  ca_path = "path/to/certs/"
```
sethvargo: Since this is a directory, I would prefer (either):

- `cas_path` (but I don't like this because I think "check-and-set"
- `ca_dir`

"path" makes me thing "path to a file", whereas "cert" makes me think "path to a place"

However - could we just allow ca_cert to be a filepath or directory path? Stat it and, if it's a directory, treat it as such?

dependency/client_set.go
sethvargo: Yea - I'm okay to change that.

README.md
sethvargo: Hmm. I see some of this might be leaking from Vault actually. I'm not 100% sure we should copy that UX. It feels weird to me to have two variables that effectively do the same thing - one is just a directory while the other is a path to a specific file.

dadgar: Its not just Vault UX it is the TLSConfig struct too

dadgar:
Let me know the action items. Not going to change the Vault client creation in this PR btw

README.md
sethvargo: Okay - let's leave it as-is

Issue #765, opened 2016-10-25
Compilation instructions are incomplete


mcnels1:
### Consul Template version
consul-template v0.17.0dev (febd929a1cb04657993dc44332851e2965e1f482)

### Configuration
There is no configuration involved, just trying to follow the instructions in https://github.com/hashicorp/consul-template#contributing

### Command

```shell
make dev
```

### Debug output
<script src="https://gist.github.com/mcnels1/690d5d131d634fa70c98f1c5e64597db.js"></script>

### Expected behavior
consul-template is compiled

### Actual behavior
lots of errors

### Steps to reproduce
1. Install go (I used brew install go)
2. Clone consul-template to a random directory not under $GOPATH
3. make dev

sethvargo:
This is fixed now in 2155fc5

PR #766, opened 2016-10-25
provide more complete compilation instructions

mcnels1:
I have opened issue #765 

The two main missing elements from the instructions are where to clone to and the need to run make bootstrap to install gox.


sethvargo:
This is fixed with the new Docker-based build process

Issue #760, opened 2016-10-21
Add support for creating a struct in a template


bitsofinfo:
I have a need to capture the output of a {{template ..}} call to variable. If some functions were exposed via consul-template when the containing template is processed this might be possible. 

Can you guys please consider adding such a feature?

http://stackoverflow.com/questions/40164896/capture-or-assign-golang-template-output-to-variable

bitsofinfo:
Actually just support (exposed function) for creating a struct in the template and pass it around would be great

ewah:
Something that implements methods close to Java's Map interface should do the trick. 

Too bad there's no magic method todo $myStruct.input.

It has the potential of looking like this.
`
{{- define "deftmpl" -}}
	{{ .Remove "a" }}
	{{ .Put "one" 1 }}
	{{ .Put "two" 2 }}
{{- end -}}

{{- /********************************* */ -}}


Hello Template
{{ $myStruct := NewMap }}
{{/* output empty */}}
{{ $myStruct }}

{{ $myStruct.Put "a" "abc" }} {{ $myStruct.Put "p" "pq" }} {{ $myStruct.Put "x" "xyz" }}

{{/* output apx */}}
{{ $myStruct }}

{{ template "deftmpl" $myStruct}}

{{/* output px12 */}}
{{ $myStruct }}

{{ $tmp := $myStruct.Get "two" }}

{{ $yourStruct := NewMap }}
{{ $yourStruct.Put "z" "DEF" }}
{{ $yourStruct.ClearReturnBlank }}

{{ $yourStruct.Put "a" "ABC" }} {{ $yourStruct.Put "c" $tmp }}
{{/* output ac */}}
{{ $yourStruct }}

{{ $yourStruct.PutAllReturnBlank $myStruct }}

{{/* output acpx12 */}}
{{ $yourStruct }}
`

ewah:
quickly put together in https://github.com/ewah/consul-template

ewah:
with General Map structure for returning values from define. #768

nfedyk:
+1

bitsofinfo:
Just this simple thing will make this pretty versatile, a referencable struct

peloncano:
+1

bmudda:
+1

PR #768, opened 2016-10-27
General Map structure for returning values from define. 

ewah:
Add support for creating a struct in a template #760


bitsofinfo:
Nice!

bitsofinfo:
#780

ewah:
Wondering what the future of this PR will be? 

We're about to push this into an integration testing arena.  And would prefer to know if the direction will go this way or towards something like Scratch (#780).

From a quick look, Scratch adds an "Add()" method and does some stricter logging than my Map.  My Map has a few more method implemented (containsKey and containsValue I really would like to have in play)

jippi:
the indentation is way off? :)

bitsofinfo:
No discussion on this PR vs Scratch??

PR #781, opened 2016-11-06
Fix make dev target

alvaroaleman:
Not sure what the current code is/was supposed to do, but cut only accepts one ``-f`` argument.
Also the build only works if PROJECT is set to ``github.com/hashicorp/consul-template``

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ CURRENT_DIR := $(CURRENT_DIR:/=)
 # Get the project metadata
 GOVERSION := 1.7.3
 VERSION := 0.17.0
-PROJECT := $(shell echo $(CURRENT_DIR) | rev | cut -d'/' -f1 -f2 -f3 | rev)
+PROJECT := github.com/hashicorp/consul-template
 OWNER := $(dir $(PROJECT))
 OWNER := $(notdir $(OWNER:/=))
 NAME := $(notdir $(PROJECT))


sethvargo:
Hi @alvaroaleman 

My version of `cut` accepts multiple `-f` flags:

```shell
echo $(pwd) | rev | cut -d'/' -f1 -f2 -f3 | rev
github.com/hashicorp/consul-template
```

This is designed to make the Makefile portable. If you are in a proper GOPATH as described in the README, the result of that command will return `github.com/hashicorp/consul-template`.

Can you share more information please?

alvaroaleman:
That's interesting. My cut is from GNU coreutils 8.25 on Fedora 24. Fedora 25 Beta has the same coreutils version.

Quote from the ``cut`` man page: ``Use one, and only one of -b, -c or -f. ``

Aside from the cut issue:

Where does the README describe how to set up a GOPATH? I followed the instructions in the [contributing](https://github.com/hashicorp/consul-template#contributing) section, those only state to git clone and do a ``make dev``.

IMHO it is much simpler not having to set up a GOPATH if the goal is just to try out the current development version.  Does the hardcoded value have any disadvantages?

sethvargo:
Hi @alvaroaleman 

To work on Go projects, you need a $GOPATH (this is a requirement of Go, not a requirement of Consul Template). We generally assume that you are familiar with the Go toolchain if you are building the project.

sethvargo:
Hmm... I guess with the Docker build stuff, you don't technically need a GOPATH. I'll merge this for the next release. Thanks!

Issue #799, opened 2016-11-21
Programmatically access the dependencies a template is blocked on


dadgar:
It would be great to be able to access the dependencies a template is blocked on for to know the template hasn't been rendered because it is waiting for a key to exist.

Further, it would be nice to be able to be notified when the set changes.

sethvargo:
This is done!

PR #803, opened 2016-11-24
Add POC for a channel that holds template events

sethvargo:
Relates to #799 for early feedback from @dadgar 

This adds a new `TemplateEvents` channel which will push individual template events onto the consumer. The event will include the full list of used dependencies, missing dependencies, and unwatched dependencies, as well as the result of WouldRender and DidRender.

If you like this, I'm fairly certain we can remove the other RenderEvent stuff. Does this give you enough information to do what you need to do?

### DIFF
--- a/config/retry.go
+++ b/config/retry.go
@@ -25,8 +25,8 @@ type RetryConfig struct {
 	// the error fall through.
 	Attempts *int
 
-	// Backoff is the base of the exponentialbackoff. This number will be multipled
-	// by the next power of 2 on each iteration.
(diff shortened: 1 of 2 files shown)


dadgar:
I wonder if it is better to add the data to the `RenderEvent` and use the same methods to consume: https://github.com/hashicorp/consul-template/blob/a5dc2f9a0a8a51bd3f218ce0046e7a6c79191bdd/manager/runner.go#L421-L438

As a consumer it would be very hard to correlate these if they are separate

Issue #814, opened 2016-12-08
executeTemplate: allow optional argument


aferreira:
This is a request for an enhancement.

Extend "executeTemplate" with an optional argument. So while

    {{executeTemplate templateName}}

executes in the context of the parent,

    {{executeTemplate templateName data}}

executes in the context given by the argument "data"

sethvargo:
This was merged. Thanks!

PR #815, opened 2016-12-08
executeTemplate: allow optional argument

aferreira:
While

    {{executeTemplate templateName}}

executes in the context of the parent,

    {{executeTemplate templateName data}}

executes in the context given by the argument "data".

### DIFF
--- a/template/funcs.go
+++ b/template/funcs.go
@@ -47,11 +47,22 @@ func datacentersFunc(b *Brain, used, missing *dep.Set) func() ([]string, error) 
 }
 
 // executeTemplateFunc executes the given template in the context of the
-// parent. This can be used for nested template definitions.
-func executeTemplateFunc(t *template.Template) func(string) (string, error) {
-	return func(s string) (string, error) {
+// parent. If an argument is specified, it will be used as the context instead.
+// This can be used for nested template definitions.
+func executeTemplateFunc(t *template.Template) func(string, ...interface{}) (string, error) {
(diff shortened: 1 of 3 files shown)


aferreira:
This is a proposal to solve https://github.com/hashicorp/consul-template/issues/814

template/funcs.go
```diff
@@ -47,11 +47,22 @@ func datacentersFunc(b *Brain, used, missing *dep.Set) func() ([]string, error)
...
+	return func(s string, data ...interface{}) (string, error) {
+		var dot interface{}
+		switch len(data) {
+		case 0:
+			dot = nil
+		case 1:
+			dot = data[0]
+		default:
+			return "", fmt.Errorf("executeTemplate: wrong number of arguments, expected 1 or 2"+
+				", but got %d", len(data)+1)
```
sethvargo: I don't think you want the +1 here

aferreira: I think the +1 is correct because `len(data)` does not account for the previous (required) argument for the template name.

    {{ executeTemplate "t" }}         => the original usage
    {{ executeTemplate "t" .ctx }}  => the augmented case
    {{ executeTemplate "t" .ctx .something_else }}  
            => error "wrong number of arguments, expected 1 or 2, but got 3"

sethvargo: Ah, I'm reading it as "arguments to the template", but you're reading it as "arguments to the `executeTemplate` function". Interesting... I guess I don't care enough.

PR #840, opened 2017-01-11
Make master-only readme doc block more visually stand out

jippi:


sethvargo:
It's already in bold, but people still don't see it 😦 . Hopefully this'll help. Thanks! 😄

Issue #844, opened 2017-01-13
v0.18.0-rc2 doesn't have vault write operations wired up


nbrownus:
We were trialing v0.18.0-rc2 and noticed that we couldn't use it with pki mounts in `vault`

```
{{- with secret "pki-mount/issue/cert" "common_name=..." "alt_names=..." "ttl=..." "format=pem"}}
...
{{- end }}
```

Digging into it we noticed that only `dep.NewVaultReadQuery` was wired up. We did get it to work by porting over the code that was present in v0.16.0.

I am also curious why errors are no longer being reported.

Here is how we got it working

```
diff --git template/funcs.go template/funcs.go
index 08533c0..e678a0a 100644
--- template/funcs.go
+++ template/funcs.go
@@ -286,15 +286,37 @@ func nodesFunc(b *Brain, used, missing *dep.Set) func(...string) ([]*dep.Node, e
 }
 
 // secretFunc returns or accumulates secret dependencies from Vault.
-func secretFunc(b *Brain, used, missing *dep.Set) func(string) (*dep.Secret, error) {
-       return func(s string) (*dep.Secret, error) {
+func secretFunc(b *Brain, used, missing *dep.Set) func(...string) (*dep.Secret, error) {
+       return func(s ...string) (*dep.Secret, error) {
                result := &dep.Secret{}
 
                if len(s) == 0 {
                        return result, nil
                }
 
-               d, err := dep.NewVaultReadQuery(s)
+
+               path, rest := s[0], s[1:]
+
+               data := make(map[string]interface{})
+               for _, str := range rest {
+                       parts := strings.SplitN(str, "=", 2)
+                       if len(parts) != 2 {
+                               return result, nil
+                       }
+
+                       k, v := strings.TrimSpace(parts[0]), strings.TrimSpace(parts[1])
+                       data[k] = v
+               }
+
+               var d dep.Dependency
+               var err error
+
+               if len(rest) == 0 {
+                       d, err = dep.NewVaultReadQuery(path)
+               } else {
+                       d, err = dep.NewVaultWriteQuery(path, data)
+               }
+
                if err != nil {
                        return result, nil
                }
```

sethvargo:
Hey @nbrownus 

Thank you for reporting that. I'll get this fixed up ASAP. Can you explain this a bit more:

> I am also curious why errors are no longer being reported.

Thanks!

nbrownus:
[...]

PR #845, opened 2017-01-13
	Properly parse different vault dep types

sethvargo:
Properly parse different vault dep types

Fixes GH-844


nbrownus:
Awesome, thanks!

PR #848, opened 2017-01-14
Add render times to the RenderEvent

dadgar:
This PR brings back render times to the render event. While rebasing
Nomad's integration with the runner I realized there is not enough
information to decide whether to restart a task or even if all templates
have been rendered.

The root of the issue is timing. By the time a caller unblocks on the
render channel and retrieves the map of render events, the event
containing the boolean that it would have rendered or did render could
be replaced by a newer render event. This means that both booleans will
be false, and integrators will not know if the event ever rendered or
rendered again.

By just storing the timestamp, it is always possible to tell if the
template never would have rendered (ts = 0) and if there has been an
update since you have last seen it.

### DIFF
--- a/manager/runner.go
+++ b/manager/runner.go
@@ -109,12 +109,6 @@ type Runner struct {
 // RenderEvent captures the time and events that occurred for a template
 // rendering.
 type RenderEvent struct {
-	// DidRender determines if the Template was actually written to disk. In dry
-	// mode, this will always be false, since templates are not written to disk
-	// in dry mode. A template is only rendered to disk if all dependencies are
-	// satisfied and the template is not already in place with the same contents.
-	DidRender bool
-
 	// Missing is the list of dependencies that we do not yet have data for, but
 	// are contained in the watcher. This is different from unwatched dependencies,
 	// which includes dependencies the watcher has not yet started querying for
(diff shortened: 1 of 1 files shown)


sethvargo:
Hey @dadgar 

Is there a reason you can't correlate this using the `UpdatedAt` query?

dadgar:
@sethvargo Yeah. I think an example timeline would be telling:

1. template rendered 
2. Channel notified
3. run loop again - nothing done
4. event updates with new updated at, did render and would render are now false
5. Integrator calls RenderEvents() and it looks like nothing has happened

sethvargo:
Hey @dadgar 

Totally open to changing this, I'm just trying to understand a bit more. In the case you described, is any of the data useful? Since it's just a snapshot in time, you wouldn't be able to guarantee anything about the data. You mentioned wanting to be able to correlate missing dependencies to a particular template, but I don't think it's possible to do that if you're just grabbing a snapshot in time now.

dadgar:
I haven't yet got integrating with the dependencies, still trying to get it to run :) but I think since the dependencies are always set it should work: https://github.com/hashicorp/consul-template/blob/master/manager/runner.go#L578

Issue #847, opened 2017-01-14
High vault server CPU usage in 0.18.0-rc2


kak-tus:
### Consul Template version
0.18.0-rc2

### Configuration
Template and config same as in 0.16.
```javascript
max_stale = "2m"

template {
  source = "/root/dkimproxy.map.template"
  destination = "/etc/dkimproxy.map"
}

template {
  source = "/root/start_dkimproxy.sh.template"
  destination = "/usr/local/bin/start_dkimproxy.sh"
  perms = 0755
}

exec {
  command = "/usr/local/bin/start_dkimproxy.sh"
  splay = "60s"
}
```

```liquid
{{ range $host := secrets "secret/dkim" }}{{ with secret ( printf "secret/dkim/%s" $host ) }}
{{ $host }} dkim(s=mail,key=/etc/dkim/{{ $host }}.key){{ plugin "store.sh" ( printf "/etc/dkim/%s.key" $host ) ( printf "%s" .Data.value ) }}
{{ end }}{{ end }}
```

### Command

```shell
consul-template -config /etc/dkimproxy.hcl
```

### Debug output
None.

### Expected behavior
Normal processing.

### Actual behavior
High CPU usage at vault server when running consul-template 0.18-rc2. 0.16 version works fine.

### Steps to reproduce
1. Upgrade any consul-template to latest release.
2. Use it with vault.

### References
None.

sethvargo:
Hi @kak-tus 

Thank you for opening an issue. Could you please run with `log-level=trace` and gist the output?

kak-tus:
@sethvargo 
Sure. There is many duplicate lines in log. Consul-template tryied to get values frequently
https://gist.github.com/kak-tus/d724d5af437c5524293ff4f64569b077

sethvargo:
Hmmmm... "long polling for 300ns". That means your lease is "600ns", or at least that is the value being returned by Vault via the API. Can you share the Vault version and the output of reading that secret from vault using the `vault` CLI (you can oppress the secret, I just want to see the lease duration). What version of Vault?

kak-tus:
Vault 0.6.2 or 0.6.4 - both have same issue. Lease of this values - 86400.

```
$ vault read secret/dkim/value
Key             	Value
---             	-----
refresh_interval	24h0m0s
ttl             	86400
value           	-----BEGIN RSA PRIVATE KEY-----
...
-----END RSA PRIVATE KEY-----
```

PR #849, opened 2017-01-14
Vault list queries do not have a lease duration

sethvargo:
Fixes GH-847

### DIFF
--- a/dependency/vault_common.go
+++ b/dependency/vault_common.go
@@ -1,8 +1,10 @@
 package dependency
 
+import "time"
+
 var (
 	// VaultDefaultLeaseDuration is the default lease duration in seconds.
-	VaultDefaultLeaseDuration = 5 * 60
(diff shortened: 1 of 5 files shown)


dependency/vault_common.go
```diff
@@ -20,7 +22,7 @@ type Secret struct {
 // leaseDurationOrDefault returns a value or the default lease duration.
 func leaseDurationOrDefault(d int) int {
 	if d == 0 {
-		return VaultDefaultLeaseDuration
+		return int(VaultDefaultLeaseDuration.Nanoseconds() / 1000000000)
 	}
```
jefferai: Why not just do .Seconds(), or divide by time.Second?

sethvargo: @jefferai `.Seconds()` returns a float64 with decimals for anything less than a second. I could have divided by time.second though

jefferai: Sure but you're int()ing anyways :-D

PR #851, opened 2017-01-16
 Fix example for Scratch.Get

hazim1093:
Example syntax for `Scratch.Get` included `Scratch.Key` instead of `Scratch.Get`


sethvargo:
Ah yes. Copy-paste. Thanks!

PR #863, opened 2017-01-26
Add support for tagged addrs and node meta everywhere

sethvargo:


sethvargo:
Hmmm... these tests are all passing locally

MiLk:
What's the status on this? I would like to start using the node meta data from consul-template.

sethvargo:
Hi @MiLk 

As you can see, the tests are failing. I haven't had time to dig into why. If you want to use this, you can build from this branch, but it might have bugs

MiLk:
I've been able to build and run the tests on OSX with go 1.7.5 without any failure.
I will try again on one server with Ubuntu 16.04.
At first glance, I don't see why it's failing on travis.

sethvargo:
@MiLk same. Works fine locally, but sad on Travis. If you figure it out - let me know. I spent about 4 hours before I had to move on 😦

MiLk:
Tests are fine on my server with Ubuntu 16.04 using the same go version Travis is installing.
I will try to reproduce the Travis environment with http://stackoverflow.com/questions/29753560/how-to-reproduce-a-travis-ci-build-environment-for-debugging.

MiLk:
It seems Travis doesn't allow to listen on/close connections to ports fast enough for the tests to work.
See #876

MiLk:
You probably need to run only `make test-race` and not `make test test-race`, I think that's why the other tests are failing.

PR #876, opened 2017-02-03
Sethvargo/tagged addrs

MiLk:
Follows #863

The `-parallel=10` seems to be an issue for Travis and Consul is not always able to listen on the assigned ports, etc.


sethvargo:
Thanks! Limited the parallelism on the other PR. Thanks for helping out with this!

PR #878, opened 2017-02-04
If vault token is unspecified, try reading from ~/.vault-token

cblecker:
Similar to hashicorp/terraform#11529, if VAULT_TOKEN is unspecified in the ENV or manually, try to use the vault cli's token file, if present.

### DIFF
--- a/config/vault.go
+++ b/config/vault.go
@@ -2,9 +2,12 @@ package config
 
 import (
 	"fmt"
+	"io/ioutil"
 	"time"
+	"strings"
 
 	"github.com/hashicorp/vault/api"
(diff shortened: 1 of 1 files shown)


config/vault.go
```diff
@@ -178,7 +181,17 @@ func (c *VaultConfig) Finalize() {
 	c.SSL.Finalize()
 
 	if c.Token == nil {
-		c.Token = String("")
+		// Use the vault CLI's token, if present.
+		homePath, _ := homedir.Dir()
```
sethvargo: Hmm. I'm very hesitant to discard the errors like this. A better approach might be to do this in the CLI.

PR #894, opened 2017-02-14
Fix vault url in readme

gnur:
After a mass replace of '//' to '#' the vault url in the readme was
incorrect. This fixes that.


sethvargo:
// oops

I mean

#oops

Issue #862, opened 2017-01-25
no way to determine local datacenter


blalor:
consul-template 0.18.0 has `node`, but the resulting CatalogNode (?) instance doesn't expose the local datacenter.  I want this to exclude the local node when iterating over a service in all datacenters.

sethvargo:
Hi @blalor,

Thank you for opening an issue. This is because that API does not actually return the current data center, but I think that's a good enhancement. Could you please open an upsteam issue on Consul? Thanks!

sethvargo:
For reference: https://www.consul.io/docs/agent/http/catalog.html#catalog_node

PR #927, opened 2017-04-18
Add support for datacenters

sethvargo:
Fixes GH-862


sethvargo:
We'll need to wait for a release of Consul for the tests to pass

Issue #698, opened 2016-08-19
dedup slaves re-render templates and execute commands every 60 seconds


deltaroe:
Not sure if this falls under the bug or unexpected feature category. 

We are running consul-template 0.15.0 in dedup mode with a single template. Consul is running version 0.6.4. We tried to attach a command to the template to be executed on update and ran into an issue

The master node seems to be working fine, processing template changes and only executing the command when changes occur. The slave nodes however re-render the template and execute the attached command every 60 seconds regardless if there is a change or not. Updating the resulting file that often isn't really an issue but running the command once a minute causes issues in our environment

Looking through the code this appears to be because the template watch is doing a 60s blocking wait on the kv entry where the leader puts the template. When the 60s expires consul returns the kv entry as is, but consul-template doesn't check if the modify version changed or not. Instead consul-template assumes that a change occurred and starts up the runner to process the new template. 

For now I've wrapped the command in a if diff hack that keeps track of the last version of the template and only executes the real command if it has changed.

slackpad:
Interesting - I'll see if I can clean this up. The timeout definitely makes sense for why it's waking, but it should be able to tell that the contents are the same.

PR #933, opened 2017-05-09
Do not return if data is unchanged

sethvargo:
Fixes GH-698

### DIFF
--- a/manager/dedup.go
+++ b/manager/dedup.go
@@ -295,6 +295,9 @@ func (d *DedupManager) watchTemplate(client *consulapi.Client, t *template.Templ
 		WaitTime:   60 * time.Second,
 	}
 
+	var lastData []byte
+	var lastIndex uint64
+
 START:
 	// Stop listening if we're stopped
 	select {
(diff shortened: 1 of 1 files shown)


manager/dedup.go
```diff
@@ -342,12 +353,23 @@ START:
...
+	}
+
+	if meta.LastIndex < lastIndex {
+		log.Printf("[TRACE] (dedup) %s had a lower index, resetting", path)
+		lastIndex = 0
+		goto START
+	}
+	lastIndex = meta.LastIndex
+
+	if reflect.DeepEqual(pair, lastPair) {
```
slackpad: Did you find that you needed this? The index check should be sufficient and much cheaper.

sethvargo: I needed that when I wrote the poller on the original Consul Template. I found the index could change if someone updated the key to the same value.

sethvargo: Is that still the case?

slackpad: That's still true, though in that case the `DeepEqual` will see the index changes - you might want to just compare the values of the keys.

PR #939, opened 2017-05-20
RetryConfig: Add MaxBackoff option (fixes #938)

### LINKED ISSUE #938: Add a way to specify max wait time between retries
At the moment when specifying `retry { attempts = 0 }`, there is no upper limit for the wait time in between retries.

My proposal is to introduce an option to limit the maximum wait time.

Implementation options:
- Add a time based option like "MaxBackoff" (I'd prefer this)
- Add an option like "MaxBackoffExponent"

I understand that for normal users of consul_template this is not a problem because they will probably wrap in in a systemd unit with Restart=on-failure or applying similar techniques and thus getting the behaviour outlined.

When looking at the nomad code trying to fix https://github.com/hashicorp/nomad/issues/2623 I noticed that it would be easier if this feature was implemented in consul-template.

Would you be open for a patch?

### References

- nomad https://github.com/hashicorp/nomad/issues/2623

stefreak:
This is a work in progress Pull request so you can have a look already

Questions:
- Should this also be exposed to users via the config file?
- I did not find a test for the RetryFunc itself, did I just overlook it or should I add one?

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -232,6 +232,11 @@ func (cli *CLI) ParseFlags(args []string) (*config.Config, []string, bool, bool,
 		return nil
 	}), "consul-retry-backoff", "")
 
+	flags.Var((funcDurationVar)(func(d time.Duration) error {
+		c.Consul.Retry.MaxBackoff = config.TimeDuration(d)
+		return nil
+	}), "consul-retry-max-backoff", "")
+
 	flags.Var((funcBoolVar)(func(b bool) error {
 		c.Consul.SSL.Enabled = config.Bool(b)
 		return nil
--- a/cli.go
+++ b/cli.go
@@ -426,6 +426,11 @@ func (cli *CLI) ParseFlags(args []string) (*config.Config, []string, bool, bool,
 		return nil
 	}), "vault-retry-backoff", "")
 
+	flags.Var((funcDurationVar)(func(d time.Duration) error {
+		c.Vault.Retry.MaxBackoff = config.TimeDuration(d)
+		return nil
+	}), "vault-retry-max-backoff", "")
+
 	flags.Var((funcBoolVar)(func(b bool) error {
 		c.Vault.SSL.Enabled = config.Bool(b)
 		return nil
--- a/config/retry.go
+++ b/config/retry.go
@@ -13,6 +13,9 @@ const (
 	// DefaultRetryBackoff is the default base for the exponential backoff
 	// algorithm.
 	DefaultRetryBackoff = 250 * time.Millisecond
+
+	// DefaultRetryMaxBackoff is the default maximum of backoff time
+	DefaultRetryMaxBackoff = 0 * time.Millisecond
 )
 
 // RetryFunc is the signature of a function that supports retries.
@@ -23,11 +26,16 @@ type RetryFunc func(int) (bool, time.Duration)
 type RetryConfig struct {
 	// Attempts is the total number of maximum attempts to retry before letting
 	// the error fall through.
+	// 0 means unlimited.
 	Attempts *int
 
 	// Backoff is the base of the exponentialbackoff. This number will be
 	// multipled by the next power of 2 on each iteration.
 	Backoff *time.Duration
+
+	// MaxBackoff is an upper limit to the sleep time between retries
+	// A MaxBackoff of zero means there is no limit to the exponential growth of the backoff.
+	MaxBackoff *time.Duration
 
 	// Enabled signals if this retry is enabled.
 	Enabled *bool
@@ -50,6 +58,8 @@ func (c *RetryConfig) Copy() *RetryConfig {
 	o.Attempts = c.Attempts
 
 	o.Backoff = c.Backoff
+
+	o.MaxBackoff = c.MaxBackoff
 
 	o.Enabled = c.Enabled
 
@@ -82,6 +92,10 @@ func (c *RetryConfig) Merge(o *RetryConfig) *RetryConfig {
 		r.Backoff = o.Backoff
 	}
 
+	if o.MaxBackoff != nil {
+		r.MaxBackoff = o.MaxBackoff
+	}
+
 	if o.Enabled != nil {
 		r.Enabled = o.Enabled
 	}
@@ -103,6 +117,11 @@ func (c *RetryConfig) RetryFunc() RetryFunc {
 		base := math.Pow(2, float64(retry))
 		sleep := time.Duration(base) * TimeDurationVal(c.Backoff)
 
+		maxSleep := TimeDurationVal(c.MaxBackoff)
+		if maxSleep > 0 && maxSleep < sleep {
+			return true, maxSleep
+		}
+
 		return true, sleep
 	}
 }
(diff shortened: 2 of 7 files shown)


sethvargo:
Hi @stefreak 

Thank you for the Pull Request. I'm a bit confused by the expected behavior here. Which of the following are you trying to achieve:

1. At most backoff for a _total of X_
2. Never backoff for _more than X_

As it's written here, it says "never back off for _more than X_", but the behavior @dadgar described to me seemed more akin to the first option above (at most backoff for a _total of X_).

As the code is written, if the user chooses a backoff of 2s, attempts of 5, and max-backoff as 10s, the total sleep time will be:

| attempt | sleep | total sleep |
| :------: | :----: | :----------: |
|  1          |  2s     |    2s             |
|  2          |  4s     |    6s            |
|  3          |  16s   |    24s          |
| ...          | ...       | ...                |

In this case, as this code is written, attempt 3 will actually sleep for 10s (24s > 10s, so return 10s). This means the _total_ sleep time is actually 18s though.

Now, if implemented as option 2, the third backoff would sleep for 2s (because we've already slept for 8s, which leaves 2s remaining in the total sleep to reach "max backoff").

As a user, I can see both options as being valid. I never want my data to be more than Xs stale, but I also never want to wait more than Ys for a single backoff.

> Should this also be exposed to users via the config file?

Yes please

> I did not find a test for the RetryFunc itself, did I just overlook it or should I add one?

That'd be nice, but I think we need to agree on the behavior first 😄 

@dadgar this is what I was trying to explain in chat. It's very hard to describe this behavior in a way that makes sense to people. In my head "max_backoff" would mean "total maximum sleep time", but another person might thing "no single backoff should exceed this value".

stefreak:
@sethvargo i see where the confusion is coming from, naming things is hard right :)

The use case I have in mind is that you are in a situation where you don't want consul_template to ever terminate upon failure, but want it to retry forever.

What you are describing (limiting total sleep) is already kind of possible by limiting number of attempts.

When I set attempts to 0, attempts are unlimited and thus sleep will grow exponentially with every attempt, potentially way too large than necessary.

My idea was that this could simplify changing the nomad template behaviour to retry forever while keeping sane retry intervals facing longer outages.

Not sure what @dadgar has in mind though, I just wanted to contribute this as it was the most annoying part of nomad I discovered and I liked the rest very much so far.

config/retry_test.go
```diff
@@ -168,6 +203,7 @@ func TestRetryConfig_Finalize(t *testing.T) {
 			&RetryConfig{
 				Attempts: Int(DefaultRetryAttempts),
 				Backoff:  TimeDuration(DefaultRetryBackoff),
+				MaxBackoff:  TimeDuration(DefaultRetryMaxBackoff),
```
dadgar: Run go fmt?

config/retry.go
```diff
@@ -23,12 +26,18 @@ type RetryFunc func(int) (bool, time.Duration)
...
+        // 0 means unlimited.
 	Attempts *int
 
 	// Backoff is the base of the exponentialbackoff. This number will be
 	// multipled by the next power of 2 on each iteration.
 	Backoff *time.Duration
 
+        // MaxBackoff is an upper limit the result of the exponential backoff
+        // calculation
+        // 0 means unlimited.
```
dadgar: I would inline these with the previous line. I would also say `A MaxBackoff of zero means there is no limit to the exponential growth of the backoff.`

dadgar:
@stefreak Awesome work! I had this on my backlog to get done! So much appreciated!

@sethvargo The behavior would be something like this with your configuration (As the code is written, if the user chooses a backoff of 2s, attempts of 5, and max-backoff as 10s)

| attempt | sleep | total sleep |
| :------: | :----: | :----------: |
|  1          |  2s     |    2s             |
|  2          |  4s     |    6s            |
|  3          |  10s   |    16s          |
|  4          |  10s   |    26s          |
| ...          | ...       | ...                |

@stefreak explanation of why you would want this is spot on.

sethvargo:
Can we be sure to update the README as well?

config/retry.go
```diff
@@ -23,14 +26,19 @@ type RetryFunc func(int) (bool, time.Duration)
 type RetryConfig struct {
 	// Attempts is the total number of maximum attempts to retry before letting
 	// the error fall through.
-	Attempts *int
+	// 0 means unlimited.
+	Attempts *int `mapstructure:"attempts"`
```
sethvargo: These mapstructure values should all be the default except max_backoff

stefreak: now it should be right, right? ;)

stefreak:
@dadgar @sethvargo thank you for the feedback, now it should be done incl. unit test for RetryFunc

cli.go
```diff
@@ -232,6 +232,11 @@ func (cli *CLI) ParseFlags(args []string) (*config.Config, []string, bool, bool,
 		return nil
 	}), "consul-retry-backoff", "")
 
+	flags.Var((funcDurationVar)(func(d time.Duration) error {
+		c.Consul.Retry.MaxBackoff = config.TimeDuration(d)
+		return nil
+	}), "consul-retry-max-backoff", "")
```
dadgar: Same for vault?

stefreak: done

config/retry.go
```diff
@@ -13,6 +13,9 @@ const (
 	// DefaultRetryBackoff is the default base for the exponential backoff
 	// algorithm.
 	DefaultRetryBackoff = 250 * time.Millisecond
+
+	// DefaultRetryMaxBackoff is the default maximum of backoff time
+	DefaultRetryMaxBackoff = 0 * time.Millisecond
```
sethvargo: Let's merge this PR as-is, but I'd like to get both your thoughts on making this a non-zero default. Obviously this would be a "breaking" change, but I think setting this to a reasonably high value (like 30 minutes) would be a good idea.

dadgar: Yeah I would change all the defaults. They are too short currently in my opinion.

I would do:
```
DefaultRetryAttempts = 12
DefaultRetryBackoff = 250 * time.Milliseconds
DefaultRetryMaxBackoff = 1 * time.Minute
```

Leads to:

| attempt 	| sleep 	| total 	|
|---------	|-------	|-------	|
| 1       	| .25s  	| .25s  	|
| 2       	| .5s   	| .75s  	|
| 3       	| 1s    	| 1.75s 	|
| 4       	| 2s    	| ~4s   	|
| 5       	| 4s    	| ~8s   	|
| 6       	| 8s    	| ~16s  	|
| 7       	| 16s   	| ~32s  	|
| 8       	| 32s   	| ~1m   	|
| 9       	| ~1m   	| ~2m   	|
| 10      	| ~1m   	| ~3m   	|
| 11      	| ~1m   	| ~4m   	|
| 12      	| ~1m   	| ~5m   	|
| ...     	| ...   	| ...   	|

Gives you five minutes of retrying with a good back off ramp and fast early retries.

sethvargo:
@dadgar do you wanna submit a PR for those changes?

config/retry.go
stefreak: @dadgar I like that very much

stefreak:
@sethvargo @dadgar thank you guys, happy to have contributed this :)

Issue #972, opened 2017-07-04
Absent Vault fields render as "<no value>"


pschultz:
### Consul Template version

```
$ ./consul-template -v
consul-template v0.19.0 (33b34b3)
```

### Configuration

No config required. Template:

```liquid
{{ with secret "secret/my-secret" }}
password = "{{ .Data.b }}"
{{ end }}
```

### Command

```shell
vault server -dev -dev-root-token-id=root
```

```shell
export VAULT_TOKEN=root
export VAULT_ADDR=http://localhost:8200

vault write secret/my-secret a=a

./consul-template -once -dry \
    -vault-renew-token=false \
    -template test.tpl
```

### Debug output

https://gist.github.com/pschultz/62e2ce8bba3cc4b872e6e582f4aae97a

### Expected behavior

consul-template should fail or render `password = ""`, since the field `b` doesn't exist. This is what happens if secret/my-secret itself doesn't exist.

### Actual behavior

consul-template renders `password = "<no value>"`. This seems quite dangerous to me. It makes it very easy to setup predictable passwords, encryption keys, etc. by accident.

sethvargo:
Hi @pschultz 

Thank you for opening an issue. This is definitely an interesting one... I spent some time tracking this down. We use [Go's text/template](https://golang.org/pkg/text/template) implementation for template parsing. The templating language does support an [Option](https://golang.org/pkg/text/template/#Template.Option) to print out the zero value instead of "<no value>" (which is coming from Go's templating language, not Consul Template directly). However, I tried to switch to `missingkey=zero` and it doesn't work.

Vault's `Secret` struct returns a `Data` object of type `map[string]interface{}`, basically this:

```golang
type Secret struct {
  Data map[string]interface{}
  // other fields
}
```

Since this data is type `map[string]interface{}`, the keys have to be strings, but the values could be anything, including deeply nested types like other maps, lists, ints, floats, etc. Go's templating language relies on the type to figure out what the "zero" value is, but since `interface{}` has no zero value, the templating engine falls back to "<no value>". There's no way for it to know that it should return `""` vs `0` vs `0.0` vs `nil`, since the type `interface{}` satisfies all of those zero types.

The only other option is to use `missingkey=error`. This correctly errors, but it would introduce a breaking change for folks who are relying on the pass-through behavior as-is. As a work-around for now, you can guard against the empty value using another `with` clause:

```liquid
{{ with secret "secret/my-secret" }}{{ with .Data.b }}
[...]

PR #973, opened 2017-07-05
Add a configuration for erroring on missing keys

sethvargo:
Fixes GH-972

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,6 +434,12 @@ template {
   # return. Default is 30s.
   command_timeout = "60s"
 
+  # Exit with an error when accessing a struct or map field/key that does not
+  # exist. The default behavior will print "<no value>" when accessing a field
+  # that does not exist. It is highly recommended you set this to "true" when
+  # retrieving secrets from Vault.
+  error_on_missing_key = false
+
   # This is the permission to render the file. If this option is left
   # unspecified, Consul Template will attempt to match the permissions of the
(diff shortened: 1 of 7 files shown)


pschultz:
Hi Seth,

thanks for your quick work on this.

Unless I missed something there no way to set the new option on the command line, without a config file. We are running consul-template literally with -once -dry in CI, and creating a config file is inconvenient. It would also be nice to enforce this option in the build script by setting a flag.

By the way, your fork seems to be a private repo. Just in case that's not intended.

sethvargo:
Hi @pschultz 

Thank you for your comment. You are correct that this is not a command-line option. There's no easy way to add additional per-template options to the CLI. The `source:destination:command` syntax is very limiting and leaves out a ton of other available configuration options including:

- `backup`
- `contents` (raw contents instead of source file)
- `error_on_missing_key` (new)
- `perms`
- `wait`
- `left_delim` and `right_delim` (for customizing the in-template delimiters

I don't see a way to add this as a "flag", since it's a per-template configuration option. In large organizations, one template might want the error-out behavior while another is happy with the "<no value>" semantics. It's not safe to make this a global configuration option, unfortunately.

I don't have a fork - I'm working off of a branch called "sethvargo/exit_error" directly on this repo.

pschultz:
Yes, I didn't consider multiple templates. We'll make it work as-is then. Thanks again.

PR #976, opened 2017-07-10
Added log-utc option to use local timezone

bogdanov1609:
This PR adds new option -log-utc. 
By default it's enabled and logger writes logs in UTC
If set -log-utc to false logger will write logs with server default timezone.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -359,6 +359,11 @@ func (cli *CLI) ParseFlags(args []string) (*config.Config, []string, bool, bool,
 		return nil
 	}), "log-level", "")
 
+	flags.Var((funcBoolVar)(func(b bool) error {
+		c.LogUTC = config.Bool(b)
+		return nil
+	}), "log-utc", "")
+
 	flags.Var((funcDurationVar)(func(d time.Duration) error {
 		c.MaxStale = config.TimeDuration(d)
 		return nil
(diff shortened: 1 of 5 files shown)


bogdanov1609:
Hi, any comments here?

jefferai:
What is the use case and why is this enabled by default?

bogdanov1609:
By default consul-template writes the stdout logs at UTC time zone. 
This pr adds option to disable UTC time zone and write stdout logs with default system datetime.

In our case we have two datacenters, both at UTC+3 timezone. So all our applications write logs in UTC+3. 
And when we try to debugging consul-template it's very hard to compare logs time. 
We have tried to use syslog but it seems the native stdout method is more correct. And we don't use syslog for applications logs at all.

dormidon:
The feature will be very useful for us too!

bogdanov1609:
Any updates here? Maybe i should rename option or try another way to solve our cases?

sethvargo:
Hi @bogdanov1609 

Thank you for the PR. Unfortunately we are not going to be merging this. We are working on standardizing all our logging into a single package https://github.com/hashicorp/go-hclog, and therefore are not accepting changes to the existing logging solution at this time. Thanks and have a great day! 😄

PR #978, opened 2017-07-12
Use new renewer API client

sethvargo:


config/vault.go
```diff
@@ -11,7 +11,7 @@ const (
 	// DefaultVaultGrace is the default grace period before which to read a new
 	// secret from Vault. If a lease is due to expire in 5 minutes, Consul
 	// Template will read a new secret at that time minus this value.
-	DefaultVaultGrace = 15 * time.Second
+	DefaultVaultGrace = 5 * time.Second
```
LinusU: The commit message says 5 minutes, but this looks like 5 seconds, typo? @sethvargo

PR #1003, opened 2017-08-31
Add note about generate_lease in the pki example

alvaroaleman:


examples/vault-pki.md
```diff
@@ -8,7 +8,11 @@ When generating PKI certificates with Vault, the certificate, private key, and
 any intermediate certs are all returned as part of the same API call. Most
 software requires these files be placed in separate files on the system.
 
+**Note:** [`generate_lease`][generate_lease] must be set to `true` (non-default) on the
```
sethvargo: I intentionally didn't document this because other versions of Vault automatically generated leases 😦

alvaroaleman: Well, in Vault 0.8.1 it is turned off by default and results in certificates with no associated lease. After spending an evening trying to find out why consul-template ignores the ttl on the pki role, I thought this might help others.

sfnelson: Anecdotally, this documentation would have saved me a couple of hours debugging too

Issue #1011, opened 2017-09-14
Feature request: Notify systemd when consul-template has completed its first run


sfnelson:
On systems that uses systemd for managing processes, it would be helpful if consul-template could inform systemd when it had finished processing templates so that systemd can delay initialising other services until their configurations have been generated.

Specifically, I want to use vault's PKI infrastructure to issue a certificate to identify a machine. I can configure consul-template to create the certificate file and notify the services that use it when the certificate changes (via exec), but I can't delay the services from starting on boot until the certificate has been created. My workaround is to run consul-template -once with `ExecStartPre`, then run it as a daemon with `ExecStart`, but this causes the certificate to be issued twice because consul-template doesn't retain information about leases between invocations. This is racy enough that sometimes services will fail to initialise because consul-template is concurrently writing the second certificate/key pair while they are initialising.

If consul-template supported `sd_notify` (consul does) then I could create a `Type=notify` service for consul-template and systemd would correctly delay services that depend on consul-template until after it has completed its first run.

Related: https://github.com/hashicorp/consul/issues/2121

sethvargo:
Hi @sfnelson 

Is there a reason you can't signal systemd using the arbitrary command that runs after all templates have been rendered, something like:

```
$ consul-template -template:foo:bar:/my-script.sh
```

```text
# my-script.sh
do_stuff
sudo systemctl start foo
```

sfnelson:
Hi @sethvargo. Thanks for your reply.

I don't understand what you mean. Perhaps you could explain further?

Here is a simplified version of my problem:

```
consul-template.service
 - nginx.hcl: /etc/nginx/nginx.conf.tpl => /etc/nginx/nginx.conf
 - forwarder.hcl: /etc/forwarder.conf.tpl => /etc/forwarder.conf
 - certificates.hcl: PKI => /etc/ssl/private/localhost.{cert,key}
nginx.service
 - needs /etc/nginx/nginx.conf (nginx.hcl)
 - needs /etc/ssl/private/localhost.{cert,key} (certificate.hcl)
log-forwarder.service
 - needs /etc/forwarder.conf (forwarder.hcl)
 - needs /etc/ssl/private/localhost.{cert,key} (certificate.hcl)
```

I could use `command` in one of the templates to start/restart the services (or a systemd target), but which one should I add it to? In both cases, there are two templates that need to be processed before the service can start.

[...]

PR #1012, opened 2017-09-15
runner: notify systemd after all templates have been rendered

sfnelson:
Add support for notifying systemd after consul-template has finished
processing all templates for the first time, so that systemd can delay
starting other units that depend on consul-template until after their
templates have been processed.

Systemd notifications use a socket to avoid creating a library dependency.

See https://github.com/hashicorp/consul-template/issues/1011


sethvargo:
https://github.com/hashicorp/consul-template/issues/1011#issuecomment-329820154

PR #1020, opened 2017-10-03
Remove unnecessary check against vault grace period in renewLease. Le…

### LINKED ISSUE #1019: Dynamic secrets are renewed all the time, rather than according to their lease duration
Reproduced this on consul-template master using Vault's AWS secret backend with a lease period of 2 minutes and the following test template that read accesskeys and stored them in a file

```
production:{{with secret "aws/creds/deploy" }}
  lease_id: {{.Data.lease_id}}
  lease_duration: {{.Data.lease_duration}}
  lease_renewable: {{.Data.lease_renewable}}
  access_key: {{.Data.access_key}}
  secret_key: {{.Data.secret_key}}
{{end}}
```

This causes consul template to constantly calling vault to renew the secret instead of every ~30 seconds (because the code uses 1/3 the lease period as the renew interval). 

I found the root cause, will have a PR for it shortly. 

Found this when investigating https://github.com/hashicorp/nomad/issues/3197

preetapan:
Leases should be renewed based on their TTL returned by the response, and this check was causing all dynamic secrets to be renewed constantly. 

This fixes #1019

### DIFF
--- a/vendor/github.com/hashicorp/vault/api/renewer.go
+++ b/vendor/github.com/hashicorp/vault/api/renewer.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"errors"
+	"log"
 	"math/rand"
 	"sync"
 	"time"
@@ -164,7 +165,6 @@ func (r *Renewer) Renew() {
 	} else {
 		result = r.renewLease()
 	}
-
 	select {
 	case r.doneCh <- result:
 	case <-r.stopCh:
@@ -269,11 +269,7 @@ func (r *Renewer) renewLease() error {
 		// Grab the lease duration and sleep duration
 		leaseDuration := time.Duration(renewal.LeaseDuration) * time.Second
 		sleepDuration := r.sleepDuration(leaseDuration)
-
-		// If we are within grace, return now.
-		if leaseDuration <= r.grace || sleepDuration <= r.grace {
-			return nil
-		}
+		log.Printf("[TRACE] Renewer: Sleeping for %v before attempting to renew lease", sleepDuration)
 
 		select {
 		case <-r.stopCh:


vendor/github.com/hashicorp/vault/api/renewer.go
```diff
@@ -269,11 +269,7 @@ func (r *Renewer) renewLease() error {
 		// Grab the lease duration and sleep duration
 		leaseDuration := time.Duration(renewal.LeaseDuration) * time.Second
 		sleepDuration := r.sleepDuration(leaseDuration)
-
-		// If we are within grace, return now.
-		if leaseDuration <= r.grace || sleepDuration <= r.grace {
-			return nil
-		}
```
preetapan: I removed this entire if clause, It would cause the renew method to return instantly because it was checking against the vault grace period. With this removed, it correctly sleeps for sleepDuration before returning. The return is interpreted as a write to the DoneCh which causes the loop in vault_read.go to go renew the token again. After I removed this and tested the same template, it correctly only renews according to leaseDuration.

sethvargo:
Hi @preetapan 

You're editing code in vendor  - the renewer lives in Vault.

Not all leases have a TTL - I've discussed this with Alex multiple times. Only dynamic secrets have TTLs. Static secrets, such as those in the kv backend, only provide a refresh interval and do not count down. In those situations, this breaks.

There is no answer here, but I think removing this grace check breaks the promise of the grace period. If the remaining time on a renewal is <= to grace, CT should stop and request a new credential. If users are setting grace periods which are outside of the random splay, this is going to continue to happen. The best solution is to come up with a more sane default for the default grace, or disable the default grace entirely and make it opt-in instead.

preetapan:
whoa, totally spaced on the fact that this was vendored, sorry! 
> If the remaining time on a renewal is <= to grace, CT should stop and request a new credential. 
Maybe I don't understand what this grace period is about, but that statement implies that all TTLs for leases must be greater than the grace period, otherwise CT will *constantly* renew.  In my example the grace period was 5 minutes, and the ttl was 3 minutes. Should the grace period be set as low as possible then? I guess I don't see how having two different ways to decide when to renew a token helps. 

What is "random splay" ?

preetapan:
>Not all leases have a TTL

Again, all new code to me so I could be wrong, but the method I edited pretty much just uses the [lease duration](https://github.com/hashicorp/vault/blob/4dd846657ce34e3529dd1386ff0780321b8bcd6a/api/renewer.go#L270), if that's optional then why isn't is checking anything about the value

preetapan:
I am closing this PR. Let us discuss if you want to make nomad set a more reasonable grace period, or have consul template's default grace period become 15 seconds (that matches Vault's default).

sethvargo:
> Maybe I don't understand what this grace period is about, but that statement implies that all TTLs for leases must be greater than the grace period, otherwise CT will constantly renew

It's a confusing concept. In short, it's a minimum amount of time where the client should not renew and just return. Here's why: let's say I have a lease of 30min. The algorithm sleeps for ~10m, wakes up, sees ~20min remaining, sleeps ~10min, wakes up, sleeps ~5min, wakes up, sleeps ~2.5m, wakes up.... at some point, it will be sleeping for nanoseconds. We need a minimum value which says "when the calculated remaining lease duration is <= to this value, just return now". Otherwise, you'll effectively send thousands of requests to Vault each time a secret is nearing it's maximum expiration time.

Additionally, in the case of dynamic secrets (like a database credential), you usually don't want to wait until the credential is fully-expired before trying to renew. If my app needs to talk to a database, I want to give my application time to gracefully stop, get new creds, and start with those new creds. Waiting until Vault completely revokes that database credential will most likely incur downtime, especially when all the services are started at the same time.

In the case of a static secret (a secret without a lease), the lease duration comes back as a refresh interval, but never ticks down. In these cases, you'll never reach the grace, since there's no clock counting down.

> In my example the grace period was 5 minutes, and the ttl was 3 minutes.

_Which_ ttl was 3 minutes? There are two TTLs - the original TTL and the remaining TTL after renewal. In either event, your example tells Consul Template "any credential which is going to expire within 5 minutes (+/- some randomness), just return and give me a new one so my app has time to deal with it". The behavior you've described _is_ the intended behavior for those values. This is documented in the README:

> # Note: If you set this to a value that is higher than your default TTL or
> # max TTL, Consul Template will always read a new secret!

I'm all for looking into ways to make that clearly, but it's called out explicitly in the documentation that this is the behavior. Again, as I mentioned above, without this behavior, your apps will DOS Vault.

If your credential had a TTL of 30min with a Max TTL of 1d, this would have worked as expected:

```text
- renew (30min remaining)
- sleep for 30/3+rand (~10min)
- repeat 
- ...
- hour 23:30 comes along
- renew, vault returns 30, sleep ~10min
- hour 23:40 comes along
- renew, vault returns **20**, sleep ~7min
- hour 23:47 comes along
- renew, vault returns **13**, sleep  ~4min
- return, 4min < 5min, reached our grace
```

> What is "random splay"?

The heuristic is about 1/3 lease + some random number. This reduces the changes of a thundering herd problem, should many applications request credentials from Vault simultaneously. Without this value, all those apps would request renewal at the same time, every time, and they would also hit their expiration at the same time, likely resulting in downtime. The random splay (it's later down in the vendored code) prevents this from happening.

> Should the grace period be set as low as possible then?

The grace period should be set lower than the TTL (not max TTL, regular TTL). The average Vault TTL is ~30min, and the default is 30days, so a sensible default of 5min didn't seem outrageous at the time. It's entirely configurable, but if your credential's TTL is < grace, the client is going to renew continuously by the definition of the algorithm. Users which have very low TTLs should 

> I guess I don't see how having two different ways to decide when to renew a token helps.

Totally understand this is new code, but it's a really complex topic that Jeff and I have worked to build the correct behavior. I pinged you on Slack if you want to talk about it face-to-face. I'm happy to explain it to ya 😄 .

Lastly, I'm not discounting there's a bug here. Anything is possible, but there's been some disagreements about the behavior here, so I'm up for discussion.

PR #1021, opened 2017-10-03
Make default grace period match Vault's default, this makes it so 

preetapan:
users are not surprised by continuous renewals on dynamic secrets with TTLs lower than 5 minutes (the previous default).

### DIFF
--- a/config/vault.go
+++ b/config/vault.go
@@ -11,7 +11,7 @@ const (
 	// DefaultVaultGrace is the default grace period before which to read a new
 	// secret from Vault. If a lease is due to expire in 5 minutes, Consul
 	// Template will read a new secret at that time minus this value.
-	DefaultVaultGrace = 5 * time.Minute
+	DefaultVaultGrace = 15 * time.Second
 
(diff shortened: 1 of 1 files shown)


config/vault.go
```diff
@@ -11,7 +11,7 @@ const (
 	// DefaultVaultGrace is the default grace period before which to read a new
 	// secret from Vault. If a lease is due to expire in 5 minutes, Consul
 	// Template will read a new secret at that time minus this value.
-	DefaultVaultGrace = 5 * time.Minute
+	DefaultVaultGrace = 15 * time.Second
```
preetapan: I don't have a strong opinion on this, because I don't have context for where 5 minutes originally came from. 

IMO changing this to match the default in vault allows for better user experience for CT out of the box, but this can very well be managed in the caller (e.g nomad) by setting a smaller grace period. I will say that without knowing implementation details and digging through the code, it was not obvious that the default grace period affected renewal of dynamic token leases in such a drastic way.

PR #1026, opened 2017-10-05
Allow colons in the template command on the command line

tanner-bruce:
This allows you to write something like:

```shell
consul-template -template "in.ctmpl:out:curl https://google.com:443"
```

### DIFF
--- a/config/template.go
+++ b/config/template.go
@@ -399,7 +399,8 @@ func ParseTemplateConfig(s string) (*TemplateConfig, error) {
 	case 3:
 		source, destination, command = parts[0], parts[1], parts[2]
 	default:
-		return nil, ErrTemplateInvalidFormat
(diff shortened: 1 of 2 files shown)


sethvargo:
Can you add a test for this? You can also see that it's failing some windows-specific tests (which is why the behavior is defined like it is).

tanner-bruce:
@sethvargo working on it, had trouble running tests locally so I wanted to get the travis output. This is my `make test` output if you are willing to take a look: https://gist.github.com/tanner-bruce/43a5b168315a34f4f5eddd2fce9caeb7

sethvargo:
Do you have consul installed and compiled on your path?

tanner-bruce:
@sethvargo thanks! That was it.

PR #1032, opened 2017-10-23
Renew the unwrapped token

martinssipenko:
In cases where both `renew_token=true` and `unwrap_token=true` the watcher is trying to renew the wrapping token and not the wrapped token, this change fixes it. It will create a watcher by getting the vault token from Vault client, which will be the unwrapped token in case `unwrap_token=true` is used.


martinssipenko:
Any chance this fix can be released soon?

Issue #1024, opened 2017-10-05
RFC: Indent template helper


tanner-bruce:
I'd like to add an `indent` function to consul-template, along the lines of

```liquid
{{ "foo" | indent 2 }}
```

It can be somewhat accomplished via 
```liquid
{{ "foo" | regexReplaceAll "^|\n" "\n<number of spaces to indent>" }}
```

however this adds extra spacing at the start (`{{-` fixes that), and is pretty ugly to write.

Of course, it can also be accomplished by a `plugin`, but that is also not very user-friendly.

sethvargo:
What's the use case here?

tanner-bruce:
We have some keys that contain multiline blocks of YAML code. When we are including these blocks in the template, occasionally these blocks need to be at a different level of indentation. 

If there is a better solution that I don't know about, I'm all ears :)

sethvargo:
The yaml help just dumps from `yaml.Marshal`, so it'd be impossible to get a real indentation, especially if the yaml itself contains newline characters, etc

tanner-bruce:
The YAML is besides the point, it could be any arbitrary block that I want to indent. In this case, I am not even using the `toYAML` helper, because it doesn't fit quite right with what we are doing.

sethvargo:
Okay, but I think you're missing my point. Given a key in consul like this:

```
This is
my value
```

indent(2) could either return:

```
  This is
my value
```

or

```
  This is 
  my value
```

especially when you consider `\r\n` vs `\n`, etc.

tanner-bruce:
What I am looking for is the second one - all I want is a way to indent my multiline block of code from a Consul key. If `indent(2)` is not the way to do it, and writing the `regexReplaceAll` feels hacky, what would you suggest?

tanner-bruce:
> especially when you consider \r\n vs \n, etc.

I do agree with this, but it's pretty trivial to deal with in code. The contract for indent(#) would simply be that it will split a string on \n or \r\n, insert the number of spaces at the start and combine the string.

Since 
```
  This is
my value
```
is trivially accomplished by adding 2 spaces before the opening delimiter, I would disagree that this is a problem.

sethvargo:
https://github.com/hashicorp/consul-template/pull/1034

Issue #327, opened 2015-06-23
toYAML starting indent


gmr:
I'd like to use ``toYAML`` to define snippets in a YAML template that is being built, but run into an issue where due to the starting indent at 0. Is it possible to expand toYAML to include a starting indention level? If not, do you have any suggestions to get around this issue?

Ideally, I could do something like line #8 (stuff provided for context):
```
Foo:{{with $hostname := env "HOSTNAME"}}{{with $c := printf "services/foo/node_assignments/%s" ($hostname) }}{{range key $c | parseJSON}}
  {{.}}:
    connections: {{range service "bar"}}
      - {{.Node}}{{end}}
    qty: {{with $k := printf "services/foo/%s/qty" (.)}}{{ key $k }}{{end}}
    config: {{with $config := printf "services/foo/%s/config/" (.)}}
{{ tree $config | explode | toYAML(6)}}{{end}}
{{end}}{{end}}{{end}}
```

Also FWIW, I am using the following code to work around it at the moment since YAML does parse JSON properly:
```
config: {{with $config := printf "services/foo/%s/config/" (.)}}{{ tree $config | explode | toJSON}}{{end}}
```

sethvargo:
Hi @gmr 

You could write a plugin that adds indentation like this:

```liquid
{{ tree "config" | explode | toYAML | plugin "indent" 6 }}
```

Then write a plugin that splits on the newline character an indents by the give parameter.

Does that answer your question?

gmr:
Fair enough, thanks.

PR #1034, opened 2017-10-24
Add indent function to template

gmr:
This should address #1024 and #327.


sethvargo:
Hi @gmr 

Thank you for the PR. As discussed [here](https://github.com/hashicorp/consul-template/issues/1024#issuecomment-334558251), we'll need to account for all possible newline characters (`\n`, `\r\n`, and `\r`), and then re-join on the appropriate line ending (CRLF, etc).

gmr:
Will have a revision shortly.

PR #1035, opened 2017-10-24
Add a datacenter helper function

gmr:
This is primarily a convenience function that makes it easier to get at the local datacenter. While this doesn't add functionality that can be achieved with scratch:

```liquid
{with node}}{{scratch.Set "datacenter" .Node.Datacenter}}{{end}}
```

And then ``{{scratch.Get "datacenter"}}`` to render the value, I figured simply being able to do ``{{datacenter}}`` was easier and more direct.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ BUG FIXES:
 IMPROVEMENTS:
 
 * The Vault grace period in the config is now set to 15 seconds as the default. This matches Vault's default configuration for consistency.
+* Add `datacenter` function
 
 ## v0.19.3 (September 11, 2017)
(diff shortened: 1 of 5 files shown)


sethvargo:
Hi @gmr 

Thank you for the PR. I'm worried this is going to confuse users with the `datacenters` function which lists all datacenters. I also think adding convenience functions like these open flood gates. Why not add a helper for every possible value on a particular node?

The node and service datacenter is available on almost all response objects, as in your example with scratch. I would recommend using scratch for this or, if you _really_ don't want to use scratch, use a plugin that extracts the datacenter from the json node. Thanks! 😄

PR #1059, opened 2018-01-05
Add documentation for a default env variable.

### LINKED ISSUE #829: Add a envOrDefault helper
## Feature Request
For a new project we use a bench of `{{ if env "TIMEOUT_CONNECT" }}{{ env "TIMEOUT_CONNECT" }}{{ else }}5s{{ end }}` which is a bit verbose and redundant.

It would be very conveniant to use a syntax like `{{envOrDefault "TIMEOUT_CONNECT" "5s"}}`

It should be trivial to implement, I can make a pull request if you are ok with it.

dfredell:
Fixes #829 
Give an example of how to default an environment variable if it doesn't exist.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1254,6 +1254,12 @@ This function can be chained to manipulate the output:
 {{ env "CLUSTER_ID" | toLower }}
 ```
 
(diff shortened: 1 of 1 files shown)


slackpad:
Thanks!

Issue #1065, opened 2018-02-01
When watching all services, consul-template is DOSing the Consul agent


pierresouchay:
### Consul Template version

```
consul-template v0.19.4 (68b1da2)
```

### Expected behavior

Consul template should rate limit its calls to avoid DOSing the Consul Servers.

### Actual behavior

When using consul-template to watch all services on a large cluster (1k+ kind of services, more than 4500 nodes, Consul Indexes almost always change, especially on /catalog/services), consul-template very quickly DOS the Consul Servers by requesting far too fast to its local agent data and its local agent fowarding it very quickly to Consul Servers. In one of our configuration, we then see more than 500Mb/s data between Consul-Agent and Consul-Server.

slackpad:
Hi @pierresouchay we recommend [deduplication mode](https://github.com/hashicorp/consul-template#de-duplication-mode) if you have many instances all rendering the same template. Also, there's an issue with the extra writes to the catalog that's about to get a fix in https://github.com/hashicorp/consul/pull/3845.

pierresouchay:
Hello @slackpad ,

We already applied those patches https://github.com/hashicorp/consul/pull/3845 and https://github.com/hashicorp/consul/pull/3642

Basically, when watching many services (for a load balancer), consul template consumes more than 400 Mbits/sec downloading data from its local agent and the local consul agent consumes around 500 Mbits/sec (continuously) to its Consul servers.

We see this when watching /catalog/services and iterating over the results (event with filtering results based on tags / other parameters). I suspect that many services are queried continuously, this is especially visible when Consul-template uses its local agent since the latency between both is close to 0.

I am preparing a patch to add a delay in case of a successful query, but I think we are not the only ones having this issue. While deduplication mode might help, unfortunately, it is limited by the limitations of size of the KV and looks to me like a poor workaround.

I am pretty sure large clusters owners would like to have a way to limit delay before watching again a service that had just been modified. This is especially true if you want to wait for instance at least one second before regenerating your file.

I think we might keep this issue open, don't you think ?

Regards

slackpad:
I'll reopen this to take a deeper look.

> Basically, when watching many services (for a load balancer), consul template consumes more than 400 Mbits/sec downloading data from its local agent and the local consul agent consumes around 500 Mbits/sec (continuously) to its Consul servers.

That does seem like a huge amount of churn. Have you looked into consul-template's quiescence timers at all? They are designed to let you have a hold off period to rate limit updates.

[...]

Issue #1043, opened 2017-11-13
With `wait` consul-template never starts the program


iksaif:
### Consul Template version

0.19.4 (works with 0.19.0)

### Configuration

```hcl
log_level = "debug"

# Keep prometheus up even if consul cluster does not have a leader
max_stale = "24h"

syslog {
  enabled = true
  facility = "LOCAL5"
}

exec {
  reload_signal = "SIGHUP"
  kill_signal = "SIGKILL"
  kill_timeout = "30s"
  # Try to not resart all instances at the same time.
  splay = "60s"
}

# Add wait time before reloading config when multiple files are modified
# Do not put it too high as it could break marathon startup health check.
wait {
  min = "5s"
  max = "10s"
}

template {
  source = "alertmanager.yml.tmpl"
  destination = "alertmanager.yml"
}

template {
  source = "alertmanager.tmpl.tmpl"
  destination = "alertmanager.tmpl"
}

```

```liquid
{{$environ := env "ENV"}}
{{$app_id := env "PROMETHEUS_APP_ID"}}
{{$perimeter := env "PROMETHEUS_PERIMETER"}}
{{with (printf "/prometheus/%s/%s/config_%s" $perimeter $app_id $environ)}}
{{if keyExists .}}
# =======================================================================
# Updated from consul KV (key = {{.}}) at {{timestamp}}
# =======================================================================
{{key .}}
{{else}}
# =======================================================================
# Updated from consul KV (key = {{.}}) at {{timestamp}}
# =======================================================================

{{key (printf "/prometheus/%s/%s/config" $perimeter $app_id)}}
{{end}}
{{end}}
```

### Command

```shell
../consul-template \
    --config=alertmanager.hcl \
    --exec 'bash -c "echo start; sleep 10; echo stop"'
```

### Debug output

0.19.4
```
2017/11/13 12:41:20 [DEBUG] (logging) enabling syslog on LOCAL5
2017/11/13 11:41:20.648079 [INFO] consul-template v0.19.4 (68b1da2)
2017/11/13 11:41:20.648114 [INFO] (runner) creating new runner (dry: false, once: false)
[...]

PR #1066, opened 2018-02-06
Added rate_limit support to avoid sending too many requests to Consul Agent

pierresouchay:
This PR adds support for:
* waiting at least <duration> between 2 downloads when watching a file controlled by `rate_limit.min_delay_between_updates`
* add a random delay between fetching data with param `rate_limit.random_backoff`

This should avoid allow limiting the bandwidth used by consul-template on large clusters with lots of services when watching many services (and X-Consul-Index being updated very frequently)

cf https://github.com/hashicorp/consul-template/issues/1065

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,7 @@ consul {
   token = "abcd1234"
 
   rate_limit {
-    # Enable rate limit with throttle max requests/sec per endpoint.
+    # Enable rate limit with throttle max requests/sec to consul agent.
     enabled = true
 
     # Adds a random delay between consecutive calls
@@ -160,9 +160,9 @@ consul {
 
     # When watching a resource, if the resource is fetched before that amount
     # of time, do not try to fetch the resource again until this timeout occurs.
-    # Setting this value to 30 minute, means you will never be notified of changes
-    # faster than every 30 seconds between 2 updates.
-    min_delay_between_updates = "30s"
+    # Setting this value to 10 seconds means you will never be notified of changes
+    # faster than every 10 seconds between 2 updates.
+    min_delay_between_updates = "10s"
   }
   # This controls the retry behavior when an error is returned from Consul.
   # Consul Template is highly fault tolerant, meaning it does not exit in the
--- a/README.md
+++ b/README.md
@@ -151,6 +151,19 @@ consul {
   # This option is also available via the environment variable CONSUL_TOKEN.
   token = "abcd1234"
 
+  rate_limit {
+    # Enable rate limit with throttle max requests/sec per endpoint.
+    enabled = true
+
+    # Adds a random delay between consecutive calls
+    random_backoff = "33ms"
+
+    # When watching a resource, if the resource is fetched before that amount
+    # of time, do not try to fetch the resource again until this timeout occurs.
+    # Setting this value to 30 minute, means you will never be notified of changes
+    # faster than every 30 seconds between 2 updates.
+    min_delay_between_updates = "30s"
+  }
   # This controls the retry behavior when an error is returned from Consul.
   # Consul Template is highly fault tolerant, meaning it does not exit in the
   # face of failure. Instead, it uses exponential back-off and retry functions
--- a/config/config.go
+++ b/config/config.go
@@ -213,6 +213,7 @@ func Parse(s string) (*Config, error) {
 		"auth",
 		"consul",
 		"consul.auth",
+		"consul.rate_limit",
 		"consul.retry",
 		"consul.ssl",
 		"consul.transport",
--- a/config/consul.go
+++ b/config/consul.go
@@ -11,6 +11,9 @@ type ConsulConfig struct {
 	// Auth is the HTTP basic authentication for communicating with Consul.
 	Auth *AuthConfig `mapstructure:"auth"`
 
+	// RateLimit is the configuration for re-query on success.
+	RateLimit *RateLimitConfig `mapstructure:"rate_limit"`
+
 	// Retry is the configuration for specifying how to behave on failure.
 	Retry *RetryConfig `mapstructure:"retry"`
 
@@ -30,6 +33,7 @@ type ConsulConfig struct {
 func DefaultConsulConfig() *ConsulConfig {
 	return &ConsulConfig{
 		Auth:      DefaultAuthConfig(),
+		RateLimit: DefaultRateLimitConfig(),
 		Retry:     DefaultRetryConfig(),
 		SSL:       DefaultSSLConfig(),
 		Transport: DefaultTransportConfig(),
@@ -48,6 +52,10 @@ func (c *ConsulConfig) Copy() *ConsulConfig {
 
 	if c.Auth != nil {
 		o.Auth = c.Auth.Copy()
+	}
+
+	if c.RateLimit != nil {
+		o.RateLimit = c.RateLimit.Copy()
 	}
 
 	if c.Retry != nil {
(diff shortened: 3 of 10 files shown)


pierresouchay:
Any news on this ?
Since https://github.com/hashicorp/consul/issues/3890 has now been fixed, it should greatly improve performance for large clusters

pearkes:
Given consul-template is built on watches, we wonder if the suggested model should be to run it with `-once` on a schedule or based on an external signal. consul-template was designed to be a thin and concise wrapper around watches, and are concerned that adding complexity between the two will cause issues for other use cases.

Are you relying on de-duplication mode, which is disabled under `-once`? Or is there any other reason you couldn't use `-once` and accomplish the same thing as rate limiting?

pierresouchay:
@pearkes -once would have a worse behaviour on large clusters for some templates, because each time you would run -once, you would have to re-download everything

If we want to build a large template (for instance, watch all services having tag "http", for each, create a list of endpoint to create a HaProxy / nginx configuration file) => -once is not an option => well it is, but then having the full file rendered would take a lot of time and would not allow to:
 - have babysitting of processes
 - allow to send reload signals

I agree this is not optimal, but it simply renders consul-template not usable for large clusters. On our, when using it, the DL speed between consul-template and its Consul Agent was around 800Mb/sec (and around the same between Consul Agent and Consul Server)

Since https://github.com/hashicorp/consul/issues/3890 has now been merged, the situation is probably far better (except for /catalog/services and friends).

With this implementation https://github.com/criteo/consul-templaterb/ , for the same cluster, we succeeded to have around 5Mbits/sec instead of 800Mbits/sec, after optimization https://github.com/hashicorp/consul/pull/3934 => we are around 2Mbits/sec while watching at all services.

I definitely think we could improve a bit the current implementation.

I'll try with consul-template to see what is the current bandwidth after https://github.com/hashicorp/consul/pull/3934 has been applied

guidoiaquinti:
Little bump on this as https://github.com/hashicorp/consul/pull/3934 has been closed

sodabrew:
I am also interested in this work, given that `wait` has been broken since 0.19.1 per my analysis here: https://github.com/hashicorp/consul-template/issues/1043#issuecomment-435334094

Without the `wait` block, my site's Consul servers were badly pummeled when I tried updating to consul-template 0.19.5 from our previous 0.16 version (I kept the dedup stanza, but removed the wait values -- bad plan).

pierresouchay:
@sodabrew be my guest ;)

That's the main reason why I developed https://github.com/criteo/consul-templaterb it reduces the load on servers and bandwidth with orders of magnitude on large clusters while allowing far more features: real language, loops, parametrization, json, XML and even a fully customizable US UI with great performance on very large clusters: https://github.com/criteo/consul-templaterb/tree/master/samples/consul-ui

eikenb:
What's the downside to having this enabled?

I ask as given the description and general sentiment makes it sounds like this is a simple win. Why make it configurable? Why not just add a sensible limit all the time? I don't believe in making things configurable just because... there needs to be real, tangible trade-offs that people need to manage for something to justify the complexity of being configurable.

Thanks.

pierresouchay:
I would say none. The benefits have been reduced since we made patches in Consul itself, however, it is still a win for pathological cases

eikenb:
@sodabrew, @guidoiaquinti, @gmr, and anyone else who cares. Would you agree with @pierresouchay that there is no downside to this? If you agree, I propose we just make this the new behavior and skip the config options.

gmr:
I'd suggest that changing the behavior as non-optional should be made in major version changes, not minor or revision.  Don't get me wrong, I like the new options, but someone might be relying on current behavior and taken by surprise if it were to change in a point release.

pierresouchay:
@gmr the default behavior is probably not very wise (and has been changed in Consul in version 1.0.7+ for services )

gmr:
I guess if it's documented and clear that it's a change in behavior (for the better), then go for it?

pierresouchay:
Ark... now conflicting... :(

eikenb:
I'm still leaning towards just making this the default behavior and the fact that this has conflicts just reinforces that as it highlights just how much code this has to touch to make it configurable. If not configurable it looks like it'd only mean changing 1 file (view.go). 

What I'm thinking basically boils down to taking the logic from RateLimitFunc, put it a function in view.go and call it in the same place as in the patch. I'd guess with the current default values in place (set as vars to allow changing it for tests).

The current default values are: 

- rate_limit.min_delay_between_updates: 100 milliseconds
- rate_limit.random_backoff: 33 milliseconds

Are these good default values?

pierresouchay:
I would be glad to implement this, but won't have the time to in next 5 weeks, so if you need it sooner, feel free to implement.

Regards

eikenb:
I have a simplified version of this implemented that only has the little bit of code that actually does the rate limiting in watch/view.go. The current implementation has the base rate limiting value (MinDelayBetweenUpdates in the patch) as a module level variable so it can be 0'd out for tests as needed.

I was looking over the use/implementation in consul-templateb and it really looks like there might be some variety of values that are useful for the delay... so I was re-thinking the possibility of a simple command-line only option to set just the delay. This would let you disable it (by setting it to 0) as well as setting it more based on need. I'd keep it on by default, with the default value as is (100ms). Having it cli only would keep the changes down as well.

Any thoughts on this?

pierresouchay:
@eikenb At the time consul-templaterb was written, the index (especially on services, prior to version 1.0.6 IIRC) was not well behaving (so watches were always waken up when any change in raft did occur). In consul-templaterb, there is an automatic penalty for endpoints being notified too often (see https://github.com/criteo/consul-templaterb/blob/master/lib/consul/async/consul_endpoint.rb#L237 ), meaning that endpoint not changing will have a bigger delay than endpoints really being modified. One of our last PR to improve things (look at most or my PR and those from @aestek) is this one: https://github.com/hashicorp/consul/pull/4810 ... IIRC, this one is very helpful (starting with Consul 1.4.1) since if you watch a service being non-existent, it will avoid you to hammer the cluster for changes on a large cluster, hence all of the optimizations pushed in consul-templaterb.

Most of those optimizations are not really necessary anymore with recent versions since we worked a lot on reducing updates on anti-entropy changes (we have at least 5 PR merged regarding this).

Furthermore, all endpoints in of consul in consul-templaterb are handled the same way, including when there is no watch (eg: https://github.com/criteo/consul-templaterb/blob/master/bin/consul-templaterb#L69 ), i used the same mechanism to configure refresh on endpoint changing value all the time (metrics, /agent/self...) and for which I don't want being notified too often (eg: datacenters that sorts according RTT beween DCs)

This allowed us at the time, on large clusters, to reduce bandwidth from 800Mb/s to 8kb/s.

This is probably not that important today and you probably don't need that kind of granularity anymore since many optimization have been bought to Consul itself.

In the real life, most templating systems don't need to refresh data too often anyway (I mean 640ms is probably enough for anybody 😄  ), so having a safeguard to avoid refreshing every ms the cluster is a good thing and will avoid to DDOS the cluster.

In any case, I think this is a very good thing to integrate this finally in consul-template, thank you :-)

eikenb:
@pierresouchay .. Thanks so much for the information dump. Really helped me understand the background of this and why it probably doesn't need to be configurable at all. So that's what I'll start with.

I'll push my version of it up as a new PR soon.

c4milo:
@pearkes, how come this PR was not accepted? 😱 This is a must have for big deployments.

PR #1073, opened 2018-02-15
Obtain current hostname of machine/container where consul-template runs.

tears-of-noobs:
This PR provides template function - `{{ hostname }}` to set a hostname of machine or container where consul-template executes. I know that hostname can be obtained by ENV variables but in different Linux distributions, these variables are not the same, for example, in Ubuntu hostname stored in $HOSTNAME, but in ArchLinux in $HOST, so, I think the most common way to obtain hostname is read this file -  `/proc/sys/kernel/hostname`.

### DIFF
--- a/template/funcs.go
+++ b/template/funcs.go
@@ -58,6 +58,21 @@ func datacentersFunc(b *Brain, used, missing *dep.Set) func(ignore ...bool) ([]s
 	}
 }
 
+// getHostnameFunc returns a current hostname from /proc/sys/kernel/hostname
+func getHostnameFunc() func() (string, error) {
+	return func() (string, error) {
+		hostname, err := os.Hostname()
+		if err != nil {
+			return "", fmt.Errorf(
+				"hostname: can't get current hostname, reason^ %s",
+				err.Error(),
+			)
+		}
+
+		return hostname, nil
+	}
+}
+
 // envFunc returns a function which checks the value of an environment variable.
 // Invokers can specify their own environment, which takes precedences over any
 // real environment variables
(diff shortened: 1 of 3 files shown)


pearkes:
Is this something that can't be achieved by conditionally reading the various environment variables and preferring the one that works for each platform? We'd rather recommend that approach instead of adding this additional complexity.

tears-of-noobs:
Hi, thanks for reply. 

At first, I thought the same, read environment variables, but this way have some disadvantages:

- In first my message I made the mistake when I told that only $HOST env variable available in ArchLinux, $HOST and $HOSTNAME they are both available and they are part of shell $HOSTNAME sets by `bash` $HOST by `zsh`, so when I'll write template I must know about available shell on machine/container and under what kind of shell `consul-template` will be run - it inconveniently and not obvious.

- When hostname changed you must reload your shell to get the new name through env variable, otherwise, env variables will contain old data.

- Env variables `$HOST` and `$HOSTNAME` are not reliable sources because linux system not guarantees that they are available and contains something.  View the POSIX standard about expected env variables in a linux system - http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html

So, these facts speak that `$HOST` or `$HOSTNAME` unreliable and implicit data source and I wouldn't use them to determine hostname in my templates.

pearkes:
@tears-of-noobs understood appreciate that.  Maybe you can describe your use-case a bit more specifically?

We don't currently return system information via helpers that I know of, other than (maybe) `timestamp`. This would be the first instance of that and open the door to more platform specific helpers (though, understood Go does most of the hard work there).

Can I suggest a plugin approach? You can see an example from the community [here](https://gist.github.com/robbiet480/a87fe549883f2c1138a5) and the docs [here](https://github.com/hashicorp/consul-template#plugins).

tears-of-noobs:
@pearkes Hmmm, plugins look like a good solution. Thanks for the advice, I'll make a plugin to obtain any system information.

niko:
```{{ plugin "hostname" }}``` will do. Or ```{{ plugin "hostname" "-i" }}```.

Guys, i love the plugin Feature.

PR #1079, opened 2018-02-28
Add note about vault.grace period behaviour

bullno1:
Before there is a better solution to this https://github.com/hashicorp/vault/issues/3414, I think it should at least be documented.


pearkes:
Thanks! We think there are some upstream changes coming that should make this less surprising, but in the meantime we feel this is a good note.

Issue #1008, opened 2017-09-07
Add --left-delimiter and --right-delimiter command-line options


WolverineFan:
I'd like to be able to pass in the left/right delimiter on the command-line.

### Consul Template version
0.18.0 - 0.19.2

### Configuration
I can override the template delimiters in a config file using 
```hcl
template {
  ...
  left_delimiter  = "{:"
  right_delimiter = ":}"
  ...
}
```

But I have some validation code that renders the template to a temp file:
```bash
consul-template -template "$1:/tmp/consul-template-rendered" -once
```
This doesn't work if I override the delimiter.  If I could pass the delimiters on the command-line it would.  Another option would be to allow overriding the delimiters globally in addition to in the `template` block.  Then I believe I could have consul-template read a config file with just global configuration while still passing `-template ... -once` for my validation.

sethvargo:
Hi @WolverineFan 

Unfortunately it's not possible to do this via the CLI because it's a _template-specific_ option, not a global option. Template-specific options require a configuration file.

WolverineFan:
@sethvargo What about my suggestion of making it a global option?  Honestly either one would solve the problem...

WolverineFan:
Let me rephrase...  Allow setting a global `default` value and/or a CLI option to set the global default.

WolverineFan:
@sethvargo I noticed last night that `wait` is already an option you can configure a global default for.  I'm think delimiters are a reasonable thing to treat the same way.  I tried digging in to the source but I can't quite figure out how the `wait` default values get populated and used by the `template` block.  I'm happy to do the work (or at least some of it) but I'd need some tips to get started.

sethvargo:
Hi @WolverineFan 

`wait` is a legacy option, so we keep it for backwards compatibility. It predates template-specific configurations.

WolverineFan:
@sethvargo So you see no value in allowing a global default for left/right delimiter?  I'm surprised I'm the only person interested...

dd-caleb:
I also found this super-confusing, didn't realize the template option is for a single template and not global options for all possible `-template` on the command line. Because of this there doesn't appear to be an easy way to change the delimiters across the board.

PR #1081, opened 2018-03-02
allow overriding the default delimiter

dd-caleb:
For: https://github.com/hashicorp/consul-template/issues/1008

This implements a `defaults` section that allows you to overwrite the default delimiters used for all templates.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -310,6 +310,16 @@ func (cli *CLI) ParseFlags(args []string) (*config.Config, []string, bool, bool,
 		return nil
 	}), "dedup", "")
 
+	flags.Var((funcVar)(func(s string) error {
+		c.Defaults.LeftDelim = config.String(s)
+		return nil
+	}), "defaults-left-delimiter", "")
+
+	flags.Var((funcVar)(func(s string) error {
+		c.Defaults.RightDelim = config.String(s)
+		return nil
+	}), "defaults-right-delimiter", "")
+
 	flags.BoolVar(&dry, "dry", false, "")
 
 	flags.Var((funcVar)(func(s string) error {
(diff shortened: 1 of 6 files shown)


pearkes:
We don't really have much precedence for setting a global option like this, so would hesitate to accept this. Can you provide some more background (similar to what you discussed in #1008) as to why you feel this is necessary / important from a UX point of view?

dd-caleb:
Sure. 

Currently there is no way to globally change the default delimiters for consul-template. We are using helm in addition to consul-template, so we have config maps that contain embedded templates:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: example
data:
  somefile.txt: |
    {{"{{"}} key "example" {{"}}"}}
```

We would like to be able to change the delimiters so we can make embedding cleaner. For example:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: example
data:
  somefile.txt: |
   {% key "example" %}
```

The current customization options have to be applied in a config file and only support rendering a single template, but we pass multiple templates on the command line:

```yaml
command:
- consul-template
- -once
- -template=tpl1:out1
- -template=tpl2:out2
```

As far as we could there was no other way to achieve our goal.

pearkes:
Thanks @dd-caleb, appreciate the detail there. We'll try to take a look at this soon.

jboero:
Can we get an update on this?  IHAC who's trying to workflow Helm templates with consul-template and they are both hard-coded to the same delimiters.  We need this to be merged for that.
Thx

JulienBalestra:
Would love to get any news around this 👍

eikenb:
Hey @dd-caleb, since you submitted your pull request we've implemented a policy of requiring a signed CLA before merging. Would you please consider signing the CLA? Thanks.

lwille:
@dd-caleb sign the CLA, pleeeeease!

eikenb:
@dd-caleb ping? CLA?

One last try. I'm making a sweep of PRs to determine which to merge/discard and would like to consider this one for merging but can't without the signed CLA and will have to discard it. As this one does seems pretty popular, if he doesn't respond does someone else want to take a crack at it?

mdgreenfield:
@dd-caleb will not be able to sign that CLA so I've opened a new PR with the changes on behalf of Datadog.

eikenb:
New version of this in #1290

Issue #1080, opened 2018-03-01
Log level info is too noisy


jarro2783:
I think that log level `INFO` is too noisy. For example, if I want to know when Consul Template actually updates a file, I need to see all the other `INFO` messages that I don't really care about. So this message is useful:
`log.Printf("[INFO] (runner) executing command %q from %s", command, t.Display())`

but one of these every 5 seconds is not:
`[INFO] (runner) initiating run`

Can the log levels of some of the messages be adjusted so that there is a more useful distinction between info and debugging noise. Really I feel `initiating run` is debugging, and I would be happy enough if that was just moved to `DEBUG`.

### Consul Template version

`v0.19.4`

jvrplmlmn:
@jarro2783 I've opened https://github.com/hashicorp/consul-template/pull/1088 to address this issue.

pearkes:
Should be fixed in #1088

PR #1088, opened 2018-03-24
Use DEBUG instead of INFO for the noisy '(runner) initiating run' log

jvrplmlmn:
This PR addresses https://github.com/hashicorp/consul-template/issues/1080


pearkes:
Thanks @jvrplmlmn! I agree with this change, but I'm wondering if you looked for other instances of `INFO` being used too frequently that you'd suggest we change to `DEBUG`? Just want to be sure if we're dialing back the verbosity we're doing it somewhat consistently. 😄

pearkes:
We'll take this regardless of my comment, as I agree it addresses that open issue.

Issue #1095, opened 2018-04-19
High CPU Utilization


aaronbbrown:
Hello Hashicorp friends!

Over at GitHub, we are using `consul-template` to drive some of our edge and internal load balancing tier.  Recently, we have been seeing the `consul-template` process alternating between consuming a 150% CPU and 10-30% CPU.  It looks something like this, which is a graph of `consul-template` CPU utilization per host for that particular load balancing tier:

![image](https://user-images.githubusercontent.com/469234/39016810-93a3d012-43ef-11e8-928a-dad4879bcb54.png)

From ☝️, it's clear that the distribution of CPU utilization is bimodal during periods of 📈 high CPU utilization.  We have multiple separate templates and one `consul-template` process running per host in dedup mode.

Here's what we have learned:

* 📈 During CPU spikes, the template leaders are **spread out across multiple hosts**
* 📈 During CPU spikes, The higher CPU tier are *consumers of deduplicated templates* and the lower CPU tier are the *template leaders*. 
* 📉 During the low CPU periods, the template leaders are **consolidated all on the same host**

To clarify, we have 10 templates: When all 10 template leaders are consolidated to the same `consul-template` process, CPU utilization is low 📉.  During the spikes above, 2 leaders were on one host, and 6 on another and the CPU utilization was high 📈.  We can reproduce this by stopping all `consul-template` processes and bringing them back up serially.  It also happens naturally sometimes when we deploy new template files.

During the 📈 high-CPU periods, the template leaders are regenerating and writing to the dedup key several times per second every wait interval.  This causes the dedup consumers to deserialize the template from consul sometimes hundreds of times.  TCP dumps on both side validate that the rate of GETs (consumers) and PUTs (leaders) to the `consul-template/$md5/dedup` keys is dramatically different when the template leaders are consolidated vs not.  The spikes of dedup writes to Consul align with the wait interval for each template (e.g. a template with a wait interval of 30s will see a spike of writes every 30s, followed by a few seconds of waiting).  `tcpdump`s from during the CPU spikes 📈 nearly identically resemble https://github.com/hashicorp/consul-template/issues/677#issuecomment-236547098.

As an example, from the logs on one of the 2 leaders at the time (which held leadership for 8 templates) during a 📈 CPU spike.
```
$ sudo awk '/2018-04-18T14:2.*updated de-duplicate/ {print $1 }' /var/log/consul-template.log  | sort | uniq -c
...
[...]

PR #1100, opened 2018-05-03
Fix #1099: de-dupe state hashes are now deterministic

### LINKED ISSUE #1099: De-dupe calculates hash of changes based on non-deterministic encoding.
In https://github.com/hashicorp/consul-template/blob/d0ee7a7a959b1106932665677d8259974907741e/manager/dedup.go#L206-L221

The de-dupe manager tries to decide if it needs to update the template based on the current dependency state. It does this by calculating a hash over the encoded data and comparing it to previous one.

Problem is that even if that data doesn't change, there is a very high probability (which increases exponentially with number of "views" being watched) that the map keys in the gob encoding will be re-ordered differently thanks to Go's lack of guarantee on map key iteration order. (proof of concept for that here: https://gist.github.com/banks/f957165227ac0dbaa3c084960613ca51

That means that with more than a few views watched (which is the main reason to use de-dupe in the first place) the leader for a template will see the values as "changed" almost every time it checks even when there is no actual change in the dependency. How often it checks is nominally controlled by the `wait { min = X }` config. Every time it sees the "changed" hash it will re-render template, update the key in KV and cause all other watchers to update and re-render their own templates too.

We need to find an alternate diff mechanism or switch to an order-independent encoding which might not be trivial given that we allow arbitrary go responses in that value now.

mitchellh/hashstructure is probably a good approach to solving this.

banks:
See #1099 for description of the prolem.

This was a quick fix as part of a larger investigation.

I tried to add a test case that would capture the issue but our existing tests for `UpdateDeps` actually assert almost nothing about it's behaviour so writing a test case that actually captures the previous bug is a lot harder than the fix.

Wanted to get this up anyway. We can decide if the effort of writing a more comprehensive test case is required immediately vs something for a later cleanup pass (perhaps when the root cause of #1095 is identified). 

Testing locally against the repro case from #1095 with only one CT instance running I now correctly see:

```
2018/05/03 16:50:23.889480 [INFO] (dedup) de-duplicate data 'consul-template/dedup/d67f95f5efb6f8ea8a1c9755b8f7e2f9/data' already current
2018/05/03 16:50:23.889488 [DEBUG] (runner) checking template 98948e90481ed803df93f99547b9ea0c
2018/05/03 16:50:23.891222 [INFO] (dedup) de-duplicate data 'consul-template/dedup/98948e90481ed803df93f99547b9ea0c/data' already current
2018/05/03 16:50:23.891234 [DEBUG] (runner) checking template 395c159b525b3d96201f5c97c51a48f8
2018/05/03 16:50:23.893860 [INFO] (dedup) de-duplicate data 'consul-template/dedup/395c159b525b3d96201f5c97c51a48f8/data' already current
2018/05/03 16:50:23.893893 [DEBUG] (runner) checking template c486fb172507c94f000f4983c92e68a7
2018/05/03 16:50:23.895776 [INFO] (dedup) de-duplicate data 'consul-template/dedup/c486fb172507c94f000f4983c92e68a7/data' already current
2018/05/03 16:50:23.895787 [DEBUG] (runner) checking template a9498a02103215bfab188e5a3a3c7cf7
2018/05/03 16:50:23.897412 [INFO] (dedup) de-duplicate data 'consul-template/dedup/a9498a02103215bfab188e5a3a3c7cf7/data' already current
```

Instead of previously updating the template KV every time even with no actual service or template output difference.


banks:
As far as I can tell, this actually was the root cause of #1095 too and appears to fix it. More details in  https://github.com/hashicorp/consul-template/issues/1095#issuecomment-386729243.

manager/dedup.go
```diff
@@ -220,26 +221,32 @@ func (d *DedupManager) UpdateDeps(t *template.Template, deps []dep.Dependency) e
...
-	compress := lzw.NewWriter(&buf, lzw.LSB, 8)
-	enc := gob.NewEncoder(compress)
-	if err := enc.Encode(&td); err != nil {
-		return fmt.Errorf("encode failed: %v", err)
+	// Compute stable hash of the data. Note we don't compute this over the actual
+	// encoded value since gob encoding does not guarantee stable ordering for
+	// maps so spuriously returns a different hash most times. See
+	// https://github.com/hashicorp/consul-template/issues/1099.
+	hash, err := hashstructure.Hash(td, nil)
+	if err != nil {
```
preetapan: not a blocker, but curious under what sort of inputs does hashstructure even ever return an error

vendor/vendor.json
```diff
@@ -0,0 +1,13 @@
+{
```
preetapan: This is not necessarily this PR's fault but strange that this project never had a vendor.json. Now you have added one but its contents only contain one dependency and not the others that this actually uses. Would suggest removing this file for now and we should file an issue and create a correct vendor.json with all dependencies.

manager/dedup.go
banks: mostly stuff like IO failures to write value to buffer and calculate hash over it but also if any reflect issues come across types it doesn't know how to handle etc. e.g. `fmt.Errorf("unknown kind to hash: %s", k)` (https://github.com/mitchellh/hashstructure/blob/master/hashstructure.go)

PR #1109, opened 2018-06-07
Obtain services filtered by their name and tag or multiple tags

tears-of-noobs:
Hi

This PR provides two template functions - `withTag` and `withTags` which helps to filter services with name and with tag or with multiple tags by way:
```
{{ $filtered := service "webapp" | withTag "supertag" }}

or you can use pipe to combine multiple tags
{{ $filtered := service "webapp" | withTag "supertag" | withTag "uniquetag" }}

or you can create required tag and filter by them
{{ $requiredTags := "tagone tagtwo tagthree" | slpit " " }}
{{ $filtered := service "webapp" | withTags $requiredTags }}
```

So, the main problem which I try to solve - it's an impossibility to determine required services without `range` directive and `if` condition on .Tags inside `range`. I want to know exists services or not by required name and tags before I do `range` on them.

Where it may help. I use `consul-template` with argument `-once` for the initial render of templates (nginx upstream and/or haproxy configs) when a container starts and if the required services are not yet available in the consul catalog I want to add stubs in nginx or haproxy configuration files and start the service. With current template functions, this is not possible or I can't find the solution because to determine required service I must do `range` and inside `range` I must use `if` condition on `.Tags` and Go templates don't support `break` statement to stop `range` loop over services if I can't find required service and add stub in `{{ else }} condition.

With that solution I can do that:
```
{{ $filtered := service "webapp" | withTag "supertag" }}

{{ if eq (len $filtered) 0 }}
add stub here
{{ else }}
    {{ range $service := $filtered }}
        range over available and non empty services
    {{ end }}
{{ end }}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,6 @@ Icon
 Network Trash Folder
 Temporary Items
 .apdisk
+
+tags
+consul-template
--- a/template/funcs.go
+++ b/template/funcs.go
@@ -539,6 +539,108 @@ func byTag(in interface{}) (map[string][]interface{}, error) {
 	return m, nil
 }
 
+// withTag is a template func that takes services and filter them
+// by tag.
+//
+// {{ $s := service "service-name" | withTag "tagone" | withTag "tagtwo" }}
+//
+func withTag(tag string, in interface{}) ([]interface{}, error) {
+	var (
+		m   []interface{}
+		err error
+	)
+
+	switch typed := in.(type) {
+	case nil:
+	case []*dep.CatalogSnippet:
+		for _, snippet := range typed {
+			if isTagExists(snippet.Tags, tag) {
+				m = append(m, snippet)
+			}
+		}
+
+	case []*dep.CatalogService:
+		for _, service := range typed {
+			if isTagExists(service.ServiceTags, tag) {
+				m = append(m, service)
+			}
+		}
+
+	case []*dep.HealthService:
+		for _, service := range typed {
+			if isTagExists(service.Tags, tag) {
+				m = append(m, service)
+			}
+		}
+
+	case []interface{}:
+		m, err = filterServicesByTag(
+			typed,
+			tag,
+		)
+
+		if err != nil {
+			return nil, fmt.Errorf(
+				"withTag: can't filter services by tag %s, reason: %s",
+				tag,
+				err.Error(),
+			)
+		}
+
+	default:
+		return nil, fmt.Errorf(
+			"withTag: wrong argument type %T",
+			in,
+		)
+	}
+
+	return m, nil
+}
+
+// withTags is a template func that takes services and filter them
+// by tags.
+//
+// {{ $s := service "service-name" | withTags $requiredTags }}
+//
+func withTags(tags []string, in interface{}) ([]interface{}, error) {
+	var (
+		m []interface{}
+	)
+
+	switch typed := in.(type) {
+
+	case nil:
+	case []*dep.CatalogSnippet:
+		for _, snippet := range typed {
+			if isTagsExists(snippet.Tags, tags) {
+				m = append(m, snippet)
+			}
+		}
+
+	case []*dep.CatalogService:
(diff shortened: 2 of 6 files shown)


angrycub:
I was looking at your PR and I'm curious if `byTag` can't/doesn't provide the same behavior.

```
{{ $nomad := service "nomad" | byTag }}
{{ if eq (len $nomad.http) 0 }}
no services
{{ else }}
    {{ range $service := $nomad.http }}
       {{ $service.Address }}
    {{ end }}
{{ end }}
```

tears-of-noobs:
I checked `byTag` function again. So the main point which I described in PR it's:

- More clear and straightforward work with `tags`
- Remove limitation of numbers of filtering `tag` for one `service`. If I understood correctly `byTag` returns `map[string][]interface` where a key of the map is only one tag, so how I can do filter by multiple tags in this behavior?

Your example looks like pretty well but as I was saying - I wanna do filter by multiple tags. In my setups where `consul-template` runs generally uses more than one tag per service and tags either start with `.` or contains it so `{{ if eq (len $nomad.mytag.with.dot) 0 }}` doesn't works and I should do - `{{ if eq (len (index $nomad ".mytag.with.dot)) 0 }}` which looks not so good in terms of simplicity.

tears-of-noobs:
Guys, will you merge this PR? It exists so long and I'm don't understand whats the problem with merging it? If I should provide any additional information - tell me or tell why merge is impossible now.

bofm:
Consul API supports querying by multiple tags https://www.consul.io/api/health.html#tag. No need to fetch all services to the Consul-template side over the network. There may be thousands of services registered. Wouldn't it be better to support multi-tag queries instead of filtering?

tears-of-noobs:
@bofm thanks for your reply.  If I right understand the workflow of `consul-template` - it doesn't fetch and watches for all exists services, only for defined by `service` built-in function in your templates, and yes there may be a thousand nodes of one service, but I don't sure it can be a problem for network or performance problem for `consul-template` during filtering thousands of elements from slice. Otherwise, I'll take a look on built-in consul ability to fetch service by multiple tags and maybe I'll fix this PR to use that ability.

bofm:
Filtering is not a problem. What can be a problem is when there are thousands of services with the same name (for example, "redis") and consul-template watches all of them. The payload of the Consul HTTP response for all the services can be megabytes in size. And it will be sent to consul-template every time ANY of the services changes. And the changes can occur very often (hundred times per second). Especially when health-checks are used.

tears-of-noobs:
@bofm Yep, I understand but in fact, fetching from consul API by multiple tags can also lead to the same problem if a result of my query is thousand of nodes and megabytes traffic. So as I was saying - I'll look on the Consul API ability and change the behavior of this PR if it makes sense.

eikenb:
Hey @tears-of-noobs, thanks for your patience.

I'm reviewing all of the PRs and here your discussion ended with you looking into multi-tag queries and then you pushed code after that. Did you make any related changes or was that just a rebase or something? Thanks.

eikenb:
Hey @tears-of-noobs, since you submitted your pull request we've implemented a policy of requiring a signed CLA before merging. Would you please consider signing the CLA? Thanks.

tears-of-noobs:
Hi @eikenb I've already signed the CLA and I don't understand why it still `pending` status.
About PR, I've researched another way to determine services by multiple tags on the `Consul` side and found [ServiceMultipleTags](https://godoc.org/github.com/hashicorp/consul/api#Health.ServiceMultipleTags) function, so I think that this an idiomatic way to fetch services by multiple tags, and it remains only to implement it in the [Fetch](https://github.com/hashicorp/consul-template/blob/master/dependency/health_service.go#L136) function. So, now you can close my PR, I'll implement this feature ASAP and create a new PR.

eikenb:
@tears-of-noobs ... closing this PR as requested. If/when you file the new PR as you mentioned, please reference/mention this one to help maintain context. Thanks.

PR #1111, opened 2018-06-11
Consume CT_LOCAL_CONFIG in Go

sethvargo:
Instead of relying on the entrypoint to do this, we can just read this in Go, which will make it work in the scratch containers or containers without a shell.

/cc @pearkes


sethvargo:
Ping @banks

PR #1115, opened 2018-06-29
Implement support for K/V version 2 mounts

philpennock:
K/V version 2 mounts change the URL path to read and how data is returned.  Within consul-template, it's just harder to get values, but in envconsul it manifests as panics; see hashicorp/envconsul#175

Because I suspect that a dominant use for consul-template and envconsul is reading from secrets mounts, the "correct" behavior seems to be to try to abstract away the presence of version 2 mounts, mutating both the query and how data is returned.  Until now, the Secret object has been strictly agnostic as to the backend, but we got away with that because assuming that `map[string]interface{} == map[string]string` worked for existing backends.  This doesn't hold when K/V v2 puts the data key=value items as keys in a sub-dict of the response.

End-users reading data from Vault shouldn't need to know or care if the mount-point is V1 or V2.

Handling this is complicated by:

1. The `/sys/internal/ui/mounts` being marked internal-only; well, it's the only source of data and is used by the Vault CLI tool.  Unless and until something better comes along, it's our only choice.  Vault was released with v2 mounts by default, forcing the issue for us.
2. consul-template's vendored vault server is old, so doesn't have the sys/internal mount needed, so tests won't pass.  Trying to update the vendored copy leads to a slew of complications.

This commit skips the update of the vendored library, so tests are broken.  Fixing that too was too much yak-shaving for this external contributor when unbreaking the tools.

This commit does not change the vault write logic, only the vault read logic.  That's TBD.

---

## PR notes

Before this can be merged upstream, I think that upstream will need to update their in-repo vendored copy of the Vault library and fix all the resolution issues (protobuf mismatches, etc).  Fixing that too was too much yak-shaving for this external contributor to unbreak the tool.

Once the vendored library has updated, this PR should be mergeable, with the caveat of writes.  I can't really add meaningful tests of v1 vs v2 unless/until there's an updated Vault library to test against, initialized with a v1 and a v2 mount of the secrets backend.

Writes: there's a complete second copy of the read logic in the write path, which seems to be primarily intended for transit backends.  They could be used with K/V backends too, though.  Is this intentional and a desired use-case?  I can update the PR with handling for writes, once I know if they should be rejected or mutated, and if the approach used for the reads is acceptable.

### DIFF
--- a/dependency/vault_common.go
+++ b/dependency/vault_common.go
@@ -24,7 +24,17 @@ type Secret struct {
 
 	// Data is the actual contents of the secret. The format of the data
 	// is arbitrary and up to the secret backend.
+	// For a K/V mount of version 2, Data will hold the "data" key of the data
+	// response, unwrapping into the actual secrets, thus "do the right thing"
+	// for that type, rather than adhere strictly to agnosticism.
 	Data map[string]interface{}
+
+	// Metadata is set for K/V version 2 mounts, and holds the metadata from the
+	// "contents" of the secret, at the top-level.  It's the sibling to the data
+	// which we include in Data for this special-case.
+	// If not a K/V version 2 mount, or something else which bundles the concept
+	// of metadata, then this will be left as a nil map, rather than an empty map.
+	Metadata map[string]interface{}
 
 	// Warnings contains any warnings related to the operation. These
 	// are not issues that caused the command to fail, but that the
@@ -107,16 +117,16 @@ func vaultSecretRenewable(s *Secret) bool {
 // transformSecret transforms an api secret into our secret. This does not deep
 // copy underlying deep data structures, so it's not safe to modify the vault
 // secret as that may modify the data in the transformed secret.
-func transformSecret(theirs *api.Secret) *Secret {
+func transformSecret(theirs *api.Secret, isKV2 bool) *Secret {
 	var ours Secret
-	updateSecret(&ours, theirs)
+	updateSecret(&ours, theirs, isKV2)
 	return &ours
 }
 
 // updateSecret updates our secret with the new data from the api, careful to
 // not overwrite missing data. Renewals don't include the original secret, and
 // we don't want to delete that data accidentally.
-func updateSecret(ours *Secret, theirs *api.Secret) {
+func updateSecret(ours *Secret, theirs *api.Secret, isKV2 bool) {
 	if theirs.RequestID != "" {
 		ours.RequestID = theirs.RequestID
 	}
@@ -134,7 +144,12 @@ func updateSecret(ours *Secret, theirs *api.Secret) {
 	}
 
 	if len(theirs.Data) != 0 {
-		ours.Data = theirs.Data
+		if isKV2 {
+			ours.Data = theirs.Data["data"].(map[string]interface{})
+			ours.Metadata = theirs.Data["metadata"].(map[string]interface{})
+		} else {
+			ours.Data = theirs.Data
+		}
 	}
 
 	if len(theirs.Warnings) != 0 {
--- a/dependency/vault_read.go
+++ b/dependency/vault_read.go
@@ -3,7 +3,6 @@ package dependency
 import (
 	"fmt"
 	"log"
-	"net/url"
 	"strings"
 	"time"
 
@@ -25,6 +24,11 @@ type VaultReadQuery struct {
 
 	// vaultSecret is the actual Vault secret which we are renewing
 	vaultSecret *api.Secret
+
+	// isKV2 is possibly set non-false on the first read and indicates that a
+	// transformation will be needed to the path and to reading the secrets.
+	isKV2     bool
+	mountPath string
 }
 
 // NewVaultReadQuery creates a new datacenter dependency.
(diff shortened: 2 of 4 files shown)


gmr:
*Bump* this has been idle for four months and is an issue downstream in envconsul when using kv v2. Any chance for a merge or feedback on what you guys want to do differently?

soapergem:
I also came here looking for an update. Was trying to use envconsul with Vault (utilizing K/V v2). Was disappointed to see that envconsul is pretty much dead in the water with that version of the secrets backend, and that I need to downgrade. Any updates, Hashicorp?

catsby:
Hello! I'd like to help here, but first I need a bit more information. I'm a little new to consul-template so please bear with me 😄 

Some questions: 

- When would you say this problem manifests? Is it an issue when upgrading version of Vault in use? 
- For the patch itself, what do you think about forgoing using `sys/internal/..`and the check for kv2 version, and instead simply checking for the presence of `theirs.Data["metadata"]` in the `updateSecret` function? The metadata is only include in the KV2 response body, I feel it's safe to infer versions that way. 
- Could you pleas share an example template and `consul-template` invocation that demonstrates the issue(s) you’re hitting? Seeing the template files and command invocation should help me better understand the problems here.
- What updates do you anticipate being needed for writes? 

Some comments: 

> envconsul it manifests as panics

I’ll be honest, I’ve not looked at the linked issue where `envconsul` panics, so for now I’ll focus on just consul-template as it’s a dependency of `envconsul`. If we can sort out what needs to happen _here_, that will help me make a better choice for what needs to happen _there_.

> End-users reading data from Vault shouldn't need to know or care if the mount-point is V1 or V2

Can we clarify who we mean by “end-users” here? If the meaning is people who consume the generated template (with secrets in it, et. al.) then I completely agree. If instead we’re talking about the people who are writing the templates themselves, which may or may not be the same set of people, then I’m not sure I agree. KV1 and KV2 have different features (namely versioning support, but possibly other implications) and I’m not sure we should try to abstract away their differences. The output format of a READ is documented clearly on their respective doc pages:

- https://www.vaultproject.io/api/secret/kv/kv-v1.html#sample-response
- https://www.vaultproject.io/api/secret/kv/kv-v2.html#sample-response

———

Just talking about consul-template, I used the below template file and both kv versions:

```hcl
# example.tpl
kv1 results: {{ with secret "kv/kv1thing" }}
= {{ .Data.version}}{{ end }}

kv2 results: {{ with secret "secret/data/kv2thing" }}
= {{ .Data.data.version}}{{ end }}
```

The results I get:

```
kv1 results: 
= kv1

kv2 results: 
= kv2
```

Raw CLI:

```
$ vault secrets list --detailed 
Path          Type         Accessor              Plugin    Default TTL    Max TTL    Force No Cache    Replication    Seal Wrap    Options           Description
----          ----         --------              ------    -----------    -------    --------------    -----------    ---------    -------           -----------
cubbyhole/    cubbyhole    cubbyhole_1da77c9e    n/a       n/a            n/a        false             local          false        map[]             per-token private secret storage
identity/     identity     identity_d6981abf     n/a       system         system     false             replicated     false        map[]             identity store
kv/           kv           kv_04ee66d3           n/a       system         system     false             replicated     false        map[version:1]    n/a
secret/       kv           kv_8801a879           n/a       system         system     false             replicated     false        map[version:2]    key/value secret storage
sys/          system       system_4ddde6c1       n/a       n/a            n/a        false             replicated     false        map[]             system endpoints used for control, policy and debugging


$ vault kv get kv/kv1thing
===== Data =====
Key        Value
---        -----
version    kv1

$ vault kv get secret/kv2thing
====== Metadata ======
Key              Value
---              -----
created_time     2018-11-05T21:25:12.986988Z
deletion_time    n/a
destroyed        false
version          1

===== Data =====
Key        Value
---        -----
version    kv2

$ curl -s --header "X-Vault-Token: root" http://127.0.0.1:8200/v1/kv/kv1thing | jq .
{
  "request_id": "aa9ffe68-ce69-c796-19e5-003208ca7ff9",
  "lease_id": "",
  "renewable": false,
  "lease_duration": 2764800,
  "data": {
    "version": "kv1"
  },
  "wrap_info": null,
  "warnings": null,
  "auth": null
}

$ curl -s --header "X-Vault-Token: root" http://127.0.0.1:8200/v1/secret/data/kv2thing | jq .
{
  "request_id": "5e1e4b3e-84b1-1602-d814-f99988c7ab3f",
  "lease_id": "",
  "renewable": false,
  "lease_duration": 0,
  "data": {
    "data": {
      "version": "kv2"
    },
    "metadata": {
      "created_time": "2018-11-05T21:25:12.986988Z",
      "deletion_time": "",
      "destroyed": false,
      "version": 1
    }
  },
  "wrap_info": null,
  "warnings": null,
  "auth": null
}
```

Thank you for your patience here, I apologize for the delay in any kind of response. Hopefully I've asked good enough questions to get clarity and we can find a path forward. Thanks!

gmr:
The downstream expectation of the API as implemented and used in envconsul, which is the real issue, is that v1 and v2 behave the same and when you try and use kv2 with envconsul, it blows up, primarily because of its reliance on consul-template's API for getting at the vault data.

My primary guess is both because of the pathing differences and the ``.Data.version`` and ``.Data.data.version`` differences. The v2 change to ``.Data.data.KEY`` is unexpected and not clearly documented in consul-template.

catsby:
Thanks @gmr 

Given that, I've opened https://github.com/hashicorp/envconsul/pull/186 to address this issue over in envconsul. 

At this time I don't believe consul-template needs to change; I think it's reasonable for users to know both the path of the secret and the structure of the return response for use in their templates.

jippi:
@catsby not having _something_ like this, make it troublesome for nomad users to find a simple  upgrade path that doesn't involve double write and new KV paths.

kyhavlov:
Hey @philpennock, thanks for your contribution and apologies that we couldn't get to this in time. We ended up doing an updated pass at Vault compatibility this week and this functionality is now in v0.20.0 of consul-template. We ended up using some code from the Vault CLI and api client to do the job, and this PR helped inspire the end result of that work. Thanks again for your work here.

PR #1118, opened 2018-07-01
support for service meta 

### LINKED ISSUE #1113: Support for ServiceMeta
Support newest Version of Consul including ServiceMetas

Wing924:
fix #1113 and update vendor

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -2,7 +2,7 @@ package version
 
 import "fmt"
 
-const Version = "0.19.5"
+const Version = "0.19.6-dev"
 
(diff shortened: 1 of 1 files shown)


banks:
@Wing924 Thanks for this!

The vendor changes to this are huge which makes it seem pretty risky - can you paste what you did to update vendor? I understand Consul api needs to be updated but it's nicer to only touch the minimal set of dependencies we need to rather than pull latest version of everything going. It might be this really is minimal but it seems larger than I expect.

PR #1120, opened 2018-07-12
support for service meta

### LINKED ISSUE #1113: Support for ServiceMeta

Wing924:
fix #1113 and update vendor

what I did to update vendor:
```
dep ensure -update github.com/hashicorp/consul
```


Wing924:
@banks I deleted old branch by mistake. I re-send the PR here. #1118

banks:
😨 

Now there a over 4k files changed not "just" 800...

Something is funky with `go dep` here and I don't know what but landing 4000 + files change when we aren't sure what's going on with vendoring seems risky!

If you want to look into it feel free, I'll ask some others more familiar with `go dep` to look too.

banks:
I was wrong - `dep` doesn't allow updating submodules directly. Sorry I've not used `dep` much yet so I have to relearn all the vendoring tool footguns again!

It also looks like we previously managed deps via the make file:

```
# deps updates all dependencies for this project.
deps:
	@echo "==> Updating deps for ${PROJECT}"
	@dep ensure -update
	@dep prune
.PHONY: deps
```

Note that we prune junk out which you didn't do manually which _might_ be why the diff is so huge. On master running that gives "only" ~600 files changed.

```
git diff --shortstat
 684 files changed, 60673 insertions(+), 113275 deletions(-)
```

New versions of `dep` actually make this obsolete though (always prune) so we should remove that.

**But** if I checkout master and run the same command you pasted, check out the diff:
```
dep ensure -update github.com/hashicorp/consul
Warning: Gopkg.lock is out of sync with Gopkg.toml or the project's imports.
~/s/g/s/g/h/consul-template (master ⚡☡=) git diff --shortstat
 30 files changed, 1652 insertions(+), 216 deletions(-)
```

30 files changed sounds about right... not sure how you ended up with 4k+

That is with latest version of `dep` `0.4.1` so it is automatically pruning. Are you using an older version by any chance?

Wing924:
@banks 
I use dep 0.4.1.
```
-> % dep version
dep:
 version     : v0.4.1
 build date  : 2018-01-27
 git hash    : 37d9ea0
 go version  : go1.9.3
 go compiler : gc
 platform    : darwin/amd64
```

After I execute `dep prune`, the count of changed files become ~40.

banks:
> After I execute dep prune, the count of changed files become ~40.

Interesting... As far as I understand 0.4.1 should have automatically run prune as part of the `ensure`. Still that looks a lot better now 😄!

Gopkg.lock
```diff
@@ -94,8 +117,13 @@
...
   name = "github.com/hashicorp/consul"
-  packages = ["api","lib/freeport","testutil","testutil/retry"]
-  revision = "610f3c86a089817b5bd5729a3b8c2db33a9ae2b0"
+  packages = [
+    "api",
+    "lib/freeport",
+    "testutil",
+    "testutil/retry"
+  ]
+  revision = "73e3252076f69a06386a98a528bb79fa43bd538e"
```
banks: I verified this is the only dep that actually changed version - all the other changes here are due to `go dep` formatting differences with the current version!

Gopkg.lock
```diff
@@ -259,6 +363,12 @@
   packages = ["."]
   revision = "a61a99592b77c9ba629d254a693acffaeb4b7e28"
 
+[[projects]]
+  branch = "master"
+  name = "github.com/mitchellh/hashstructure"
+  packages = ["."]
+  revision = "2bca23e0e452137f789efbc8610126fd8b94f73b"
+
```
banks: I have a separate PR for adding this - it was my mistake it got left out, but I'll just merge yours I think.

Wing924:
@banks 
I've added service meta in other models and test cases.
Could you review it again?

banks:
Great job on adding those tests everywhere!

Thanks for this PR, it's much appreciated.

aburmeis:
Thanks for the great contribution! Mayme I missed something but I still (0.20.0) cannot get any service meta. I tried with {{ range service "test" }} and using .ServiceMeta (as found in health_service.go) but usage make the template fail. Any idea what I am doing wrong?

bofm:
Try to dump it as a json using `plugin echo` and see that's going on.

aburmeis:
good tip! So the result of {{ range service .Name }}# {{ . | toJSON }}{{ end }} looks like
```
{
  "Node": "pc1.loc",
  "NodeID": "9cb70cad-4587-fbf1-5ea4-2e7e033a2fe8",
  "NodeAddress": "127.0.0.1",
  "NodeTaggedAddresses": {
    "lan": "127.0.0.1",
    "wan": "127.0.0.1"
  },
  "NodeMeta": {
    "consul-network-segment": ""
  },
  "Address": "127.0.0.1",
  "ID": "TEST",
  "Name": "TEST",
  "Tags": [
    "myTag"
  ],
  "Checks": [...],
  "Status": "passing",
  "Port": 9008
}
```

So only NodeMeta is filled, nothing about ServiceMeta

banks:
@aburmeis what versions are your Consul clients and servers. They all need to be at least 1.2.3 to support Service Meta. Also the endpoint you output has the Service fields at the top level without `Service` prefix so I'd expect the key to be `Meta` there not `ServiceMeta` if any is present.

sodabrew:
I'm using consul-template 0.25.1. I've been digging from PR to PR for hours and I can't figure this one out. It seems like there should be a Meta field on a Service, but no:

```
{{range services}}{
    {{if in .Tags "public"}} <-- has worked for years
      {{index .Meta "apibasepath"}}  <-- nope
      {{index .ServiceMeta "apibasepath"}} <-- also nope
  {{end}}
{{end}}
```

Neither of these lines work, they both fail like this:

```
execute: template: :134:9: executing "" at <.ServiceMeta>: can't evaluate field ServiceMeta in type *dependency.CatalogSnippet
```

Does anybody have an example of a working service meta template?

eikenb:
@sodabrew .. `ServiceMeta` was added to the output of the `service` template function, not `services` (note the plural `s`). Seems like you could add an inner loop to that over the `service` and get the `ServiceMeta` data.

sodabrew:
To complete the loop for the next person, here's an excellent example in this comment: https://github.com/hashicorp/consul-template/issues/1348#issuecomment-675783081

PR #1128, opened 2018-07-28
Allow Vault token file to be used to provide token value.

patdowney:


jefferai:
@patdowney thanks for submitting!

Consul team, please do not merge this yet, as I need to discuss some workflow items with you regarding CT and Vault Agent.

patdowney:
👍Thanks Jeff. The use case for us is to avoid having to stick the token in the config file, so if there's a more preferred way of doing that then great. Also thanks for the heads up on the hot off the press `vault agent`, look forward to it being out of beta and replacing what we're currently doing.

jefferai:
@patdowney Just FYI, it's only in "beta" because in case any issues come up from people using it as it currently exists we want to be able to change the structure/config file format/etc. without having to officially break anything.

It works fine though, and there's not really anything slated to change between now and 0.11, so I highly suggest using it if you're interested in it, because this is the time during which it's easiest for any ideas/suggestions/etc. that you come up with to get incorporated into the first "non-beta" release.

jefferai:
Just an update with some thinking...

One thing CT does right now is if it gets a Vault token it tries to keep it refreshed. We actually don't want that when the token is coming from Agent because Agent is already doing that. OTOH if CT reads in a file from Vault's token-helper we have no idea which behavior is right, plus that file is really meant for user-interactive scenarios and CT isn't.

I think CT will basically need some configuration (e.g. path to file) to be able to read a token output by VA, where it knows, if this flag is set, that it should not attempt to manage lifecycle, and it should keep watch fo new values.

The question then is what, if anything, CT should do with a token-helper file. I'm leaning towards it not being officially supported. It seems odd and potentially confusing to have still another flag to control whether or not *this* file should be renewed by CT, especially since that flag would then have to be ingored if the VA file flag was set, which means still another flag to control a non-VA token file. And I think that would be necessary because I don't think reading it by default is the right approach...the tools are meant for different workloads and it seems too easy to accidentally end up using the wrong set of credentials if you really should have given CT a non-user token.

If someone really wanted to use the token helper file they could cat it into an env var and use it that way, or if they just want to be using Vault's CLI for authing they could use the available flags to not write the token helper file in the first place and simply pipe it to an env var for CT.

KJTsanaktsidis:
It seems to me that this patch really needs to _watch_ the vault agent token file somehow. That would support the following workflow:

* There's something like an EC2 metadata endpoint, or a Kubernetes service account token, that is constantly being updated
* Vault-agent creates a token from the EC2 metadata endpoint or Kubernetes service account token, and writes that to a file
* When the token's TTL expires, Vault-agent can generate _another_ token, from a new value of the EC2/K8S thing, and write it to the file
* Consul-template is re-reading the token from the file so it seamlessly picks up the new value when the old token expires.

Perhaps consul-template can re-read the token file when it would otherwise have attempted to renew it based on token TTL remaining?

MrBlaise:
Any news on this? @jefferai 

As I mentioned here: https://github.com/hashicorp/consul-template/pull/1153#issuecomment-453563241 This feature is crucial if we'd like to use vault agent.

jefferai:
I don't really have an update; I don't maintain C-T and ther are two different patches competing to do the same thing.

eikenb:
Hey @patdowney, since you submitted your pull request we've implemented a policy of requiring a signed CLA before merging. Would you please consider signing the CLA? Thanks.

patdowney:
Hi all, thanks for your help and sorry I haven't been able to pay this more attention. I'm closing this in favour of the work done by @izhukov which I think does the same job more completely and has already been merged.

eikenb:
Thanks @patdowney.

For the record, the PR referenced is #1153.

Issue #1131, opened 2018-08-06
consul-template render empty data during consul server raft election/change 


vaLski:
During:

- consul servers raft leader re-election caused by cluster outage/routine consul servers restart/etc
- consul-template that is using ```ls /path/to/keyprefix``` inside ctmpl
- is rendering **blank** data
- instead of rendering the **keys*** under the specific key prefix
- issue happens when using staled queries

I  know that according to the docs consul-template usually should not render anything if it receive error from the consul server. In that case however it seems that it is not receiving error but rather valid **empty** response from the servers which leads to rendering blank data  in the template destination.

I also experienced very similar and very bitter issue, with **consul-replicate** + stale queries. Under similar scenario (raft leader change/sync) the following data center received **blank** data from the parent leading to all KVs to be **erased** on the follower. The issue is not  fixed yet and I mitigated it by re-configuring consul-replicate to never use staled queries. Detailed report is available https://github.com/hashicorp/consul-replicate/issues/82

Recently the same issue, also happened with **consul-template** when used with staled queries. It is  only happening during raft re-election/outage recovery etec. I suspect that it is more related to consul which is sending **valid** but **blank** answers to the long pooling KV queries during servers outage/raft change. I also reported it here to the consul project but they closed it as not consul related https://github.com/hashicorp/consul/issues/3975 

Not sure if this is a strictly consul server bug or issue with consul-template and consul-replicate that are sharing the same presumably buggy code especially while using stale queries. Unfortunately this issue is really, really bad, leading to data loss or incorrect / blank configuration files rendered to template destination. Any suggestions will be greatly appreciated.

### Consul Template version
```
consul-template v0.18.1 (9c62737)
consul-template v0.19.5 (57b6c71)
```

### Configuration
```
max_stale = "5m"
```

### part of ctmpl file
```
allow_from = {{range $index, $kv := ls "/pub/server/mastermachines/pub/addr"}}{{if ne $index 0}} {{end}}{{$kv.Key}}{{end}}
```

### Command
```
consul-template -log-level info \
  -kill-signal SIGTERM -reload-signal SIGHUP \
  -vault-renew-token=false \
  -max-stale 10m \
  -config=/etc/consul-template/configs \
  -config=/etc/consul-template/templates.hcl
```
### Expected behavior

It should render like like that

```allow_from = 1.1.1.1 2.2.2.2 3.3.3.3```

### Actual behavior

```allow_from = ```

### Steps to reproduce

1. create several records
```
[...]

PR #1132, opened 2018-08-15
Introducing safels and safetree commands to the consul-template.

vaLski:
This pull request is based on rough idea dropped by Aaron Hurt.

safels and safetree behave exactly like the native ls and tree with one exception. They will *refuse* to render template, if KV prefix query return blank/empty data.

This is especially useful for rendering mission critical files that do not tolerate ls/tree KV queries to return blank data.

safels and safetree work in stale mode just as their ancestors but we get extra safety on top.

safels and safetree commands were born as an attempt to mitigate issues described here:

  https://github.com/hashicorp/consul-template/issues/1131
  https://github.com/hashicorp/consul/issues/3975
  https://github.com/hashicorp/consul-replicate/issues/82

### DIFF
--- a/README.md
+++ b/README.md
@@ -680,6 +680,23 @@ maxconns:15
 minconns:5
 ```
 
+##### `safels`
+
+Same as [ls](#ls), but refuse to render template, if the KV prefix query return blank/empty data.
+
+This is especially useful, for rendering mission critical files, that are being populated by consul-template.
+
+For example:
+
+```text
+/root/.ssh/authorized_keys
+/etc/sysconfig/iptables
+```
+
+Using `safels` on empty prefixes will result in template output not being rendered at all.
+
+To learn how `safels` was born see [CT-1131](https://github.com/hashicorp/consul-template/issues/1131) [C-3975](https://github.com/hashicorp/consul/issues/3975) and [CR-82](https://github.com/hashicorp/consul-replicate/issues/82).
+
 ##### `node`
 
 Query [Consul][consul] for a node in the catalog.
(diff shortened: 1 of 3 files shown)


pierresouchay:
Might not be needed since root cause is fIxed by https://github.com/hashicorp/consul/pull/4554

eikenb:
Hey @vaLski, thanks for the PR.

In light of https://github.com/hashicorp/consul/pull/4554 do you still think this is a good addition? 

@mitaka, @dkanchev. You both :+1:'d this... what do you think?

Thanks.

vaLski:
@eikenb Thank you for getting back to me.

Indeed the root cause of the issue I reported was fixed.

However if you think that the proposed changes are good enough and might help people preventing a disaster feel free to merge them or ask for further improvements.

While evaluating, kindly note the following:

* If an operator is rendering mission critical data (like authorized_keys, iptables firewalls etc.), he might consider using safels/safetree to ensure that he will not end up with an empty file. Due to bug in consul like the one described or accidental / operator error that will `consul kv delete -recurse /authkeysprefix/` 
* Besides https://github.com/hashicorp/consul/pull/4554 I can recall another discussion I had around (alas can't find it) where operator reported that consul was returning 404 for KV lookups, instead of 5XX during an outage. In his case I can recall that the following DC could not fetch its ACLs from the main dc due to network error or something, consul started answering with 404 for kv lookups, consul-template decided that those KVs were erased and result was case similar to mine - data deleted in following dc KV store. I think that was fixed too in consul master but I can't fix the exact discussion / PR etc.
* While digging for that discussion mentioned above I found yet one more case similar to those two described https://github.com/hashicorp/consul/issues/2637. So historically there is a trend those things happen and I assume that they will continue to via one way or another.

Feel free to comment.

eikenb:
I think I'll accept this as new template functions don't really complicate the code base much as they are pretty well self contained. I have been asking other new function writer to provide function tests... but I'll let that slide here as this is implemented more as a tweak to an existing function  vs. a brand new one. But if you feel like it.... feel free. :)

PR #1134, opened 2018-08-29
Fix race condition during file backup.

nahratzah:
If the command script moves the file away, and it is timed such that
this happens between `os.Stat()` and `copyFile()`, the `copyFile()`
operation fails and consequently consul-template will terminate.

By having the `copyFile()` operation indicate failure in an ignorable
way, we prevent this race from becoming a termination-cause.

The reason this can happen, is because consul-template does not wait
for the executed command to complete, before any new template rendering
operations take place.

### DIFF
--- a/manager/renderer.go
+++ b/manager/renderer.go
@@ -158,10 +158,8 @@ func AtomicWrite(path string, createDestDirs bool, contents []byte, perms os.Fil
 	// If we got this far, it means we are about to save the file. Copy the
 	// current contents of the file onto disk (if it exists) so we have a backup.
 	if backup {
-		if _, err := os.Stat(path); !os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


eikenb:
Hey @nahratzah, since you submitted your pull request we've implemented a policy of requiring a signed CLA before merging. Would you please consider signing the CLA? Thanks.

eikenb:
This doesn't really/completely fix the race but helps as it moves an extra check inside the copyFile() function call which would eliminate one goroutine/scheduler lock release and make the race a bit harder to hit. I think the correct behavior here is to have it just be more resilient to failure of the backup and log an error instead of crashing. I'll file an issue to fix this as that is not what this PR does (nor is it acceptable due to the CLA).

Thanks for filing it anyways as this is something that can be improved.

PR #1153, opened 2018-11-01
Add vault-agent-token-file configuration parameter for Vault Agent token

izhukov:
If this parameter is specified:
- CT will not renew token
- will monitor file for changes and reload it

### DIFF
--- a/config/vault.go
+++ b/config/vault.go
@@ -243,7 +243,7 @@ func (c *VaultConfig) Finalize() {
 			}, "")
 		}
 	}
-	if stringFromEnv([]string{"VAULT_TOKEN"}, "") != "" {
+	if StringVal(stringFromEnv([]string{"VAULT_TOKEN"}, "")) != "" {
 		c.Token = stringFromEnv([]string{
 			"VAULT_TOKEN",
 		}, "")
--- a/config/vault.go
+++ b/config/vault.go
@@ -53,8 +53,14 @@ type VaultConfig struct {
 
 	// Token is the Vault token to communicate with for requests. It may be
 	// a wrapped token or a real token. This can also be set via the VAULT_TOKEN
-	// environment variable.
+	// environment variable, or via configuration setting `vault-agent-token-file`.
 	Token *string `mapstructure:"token" json:"-"`
+
+	// VautlAgentTokenFile is the path of file that contains a Vault Agent token.
+	// If vault_agent_token_file specified:
+	//   - Consul Template will not try to renew token
+	//   - Consul Template will periodically stat token file and reload it if changed
+	VaultAgentTokenFile *string `mapstructure:"vault_agent_token_file" json:"-"`
 
 	// Transport configures the low-level network connection details.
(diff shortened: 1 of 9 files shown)


izhukov:
@jefferai Could you please take a look on it.

izhukov:
@pearkes @mkeeler please check this out.
Is it worth to merge?

config/vault.go
```diff
@@ -219,6 +231,11 @@ func (c *VaultConfig) Finalize() {
 	}
 	c.SSL.Finalize()
 
+	// Order of precedence
+	// 1. `token` configuration value
+	// 2. `VAULT_TOKEN` environment variable
+	// 3. `~/.vault-token` file
+	// 4. `vault_agent_token_file` configuration value
```
jefferai: I think the order here needs to be 4, 2, 1, 3.

MrBlaise:
This feature is a must for us. Can you please merge this?

Without this, the tutorial here: https://learn.hashicorp.com/vault/identity-access-management/vault-agent-k8s doesn't make sense

Problems:
 - Vault agent might not create the token in time and consul-template does not watch the file. Causing it to fail.
- The token is created in time but when it has expired it won't be able to function anymore. (It doesn't matter whether auth-agent renews it consul template won't be notified)

iusergii:
Also looking forward for this.

config/vault.go
izhukov: Thanks, @jefferai for pointing this out. I update PR accordingly.

izhukov:
@i0rek @mkeeler @banks @pearkes 
Could you please specify requirements that this PR should have to be allegeable to merge.
As you can see these changes needed for several peoples. And it would be valuable to see any feedback from the maintainers.

alexanderilyin:
@jefferai is the anything we can do to move forward with PR?

eikenb:
Hey @izhukov ... I know this is an old PR, but I had a question about it and this seemed the best place to ask.

Why did you have it set renew to false when the vault_agent_token_file is set? Why not use the existing renew_token field, setting it to false when you don't want it to renew the token (and allowing others to set it to true when they do want it renewed). 

I'm asking as I'm going to undo that change in an upcoming release and wanted to give you a chance to speak up in case there was a good reason for it.

izhukov:
Hi @eikenb.
We use consul-template and vault agent as a sidecars to deliver secrets from the vault to our software. Vault agent is responsible for authentication (it may be AppRole or Kubernetes auth) and reissuing a 
token. Tokens from the vault in our case not renewable, they are short-lived disposable credentials. Consul-template is responsible for actual interaction with the vault and relies on token from vault agent.
In this use case, consul-template didn't need to renew token(which it did by default), but it must periodically reread the file to peek up a new token. This was the main intention of these changes.

eikenb:
Thanks @izhukov!

So it sounds like our proposed changes will have no impact on you, which is what I expected but wanted to be sure. Thanks again.

PR #1168, opened 2019-01-21
change dedup to use one key and do auto-cleanup of the dedup kv store

### LINKED ISSUE #1158: Dedup data is not cleaned up in Consul KV storage
### Consul Template version

v0.19.5 and master (fd266d9)

### Expected behavior

Consul kv dedup storage should be maintained by Consul-Template. Stale `/consul-template/dedup/<md5>/{data,lock}` keys should be cleaned up automatically.

### Actual behavior

Many unused old dedup `data` and `lock` keys are left in the Consul kv dedup storage.

What's strange is that `/lock` keys are **not** cleaned up by Consul's Session TTL mechanism when consul-template instances are shut down gracefully. But the `/lock` keys **are** cleaned up when consul-template instances are killed.
The `/data` keys are not cleaned up in any case.

### Steps to reproduce

A script to run the testcase and it's output:

https://gist.github.com/bofm/677de0b543a071bab68bf77d070d6c3a

**Run it in an empty directory** on a host with consul and consul-template installed. It might be required to run several times to reproduce the problem.

### Questions

Does consul-template really need two keys per template hash (both `/data` and `/lock`) for deduplication? I believe one key would be sufficient. Consul-template could acquire a session lock on `/data` key. To determine if the template is already rendered a custom flag on the key could be used.

bofm:
Changed deduplication to achieve the following:
- use one kv key for both lock and data in `consul-template/dedup/<md5>/data`
- automatically cleanup deduplication data in Consul kv store using Consul Sessions

fixes #1158

### DIFF
--- a/manager/dedup.go
+++ b/manager/dedup.go
@@ -30,9 +30,7 @@ const (
 	// listRetry is the interval on which we retry listing a data path
 	listRetry = 10 * time.Second
 
-	// templateDataFlag is added as a flag to the shared data values
-	// so that we can use it as a sanity check
-	templateDataFlag = 0x22b9a127a2c03520
+	templateNoDataStr = "__NO_DATA__"
 )
 
 // templateData is GOB encoded share the dependency values
@@ -44,6 +42,10 @@ type templateData struct {
 
 	// Data is the actual template data.
 	Data map[string]interface{}
+}
+
+func templateNoData() []byte {
+	return []byte(templateNoDataStr)
 }
 
 // DedupManager is used to de-duplicate which instance of Consul-Template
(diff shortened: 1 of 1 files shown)


pearkes:
@bofm Thanks for the contribution. We'd like to land it with a few minor changes, but are you able to sign the CLA?

bofm:
I don't want to sign it. 

What would you like to change in the implementation?

manager/dedup.go
```diff
@@ -32,7 +32,8 @@ const (
 
 	// templateDataFlag is added as a flag to the shared data values
 	// so that we can use it as a sanity check
-	templateDataFlag = 0x22b9a127a2c03520
+	templateDataFlag  = 0x22b9a127a2c03520
```
kyhavlov: This can get removed now since it's not used anywhere.

bofm:
Can you please re-run the tests

bofm:
All signed. Can it be merged?

eikenb:
Merged and will be part of upcoming release. Thanks!

PR #1178, opened 2019-02-07
Clarify that `plugin` does not JSON encode its args

### LINKED ISSUE #1177: Plugin documentation is misleading
### Consul Template version

`consul-template v0.19.5 (57b6c71)`

### Configuration

```liquid
{{ with $globals := file "../globals/globals.json" | parseJSON }}
{{ with $locals := file "locals.json" | parseJSON }}
{{ plugin "merge" $globals $locals | toJSON }}
{{ end }}{{ end }}
```

### Command

```shell
consul-template -dry -once template.ct:config.json
```

### Debug output

https://gist.github.com/gregsymons/52e9b91a92da275b512697712525d6ee

### Expected behavior

The plugin should have been called with the JSON rendering of the `$globals` and `$locals` variables.

### Actual behavior

consul-template fails because the `plugin` function takes a variable number of `string` arguments rather than a `map[string]interface{}`.

### Steps to reproduce

1. Attempt to pass a non-string argument to the `plugin` template function

### References

This is a dupe of #654, which was closed because the submitter never responded to requests for details

gregsymons:
The former wording implied that the plugin would get JSON regardless of
what the arguments to `plugin` were, but the actual implementation just
execs the plugin with the string arguments.

Fixes #1177

### DIFF
--- a/README.md
+++ b/README.md
@@ -1444,7 +1444,9 @@ plugin.
 {{ plugin "my-plugin" }}
 ```
 
-This is most commonly combined with a JSON filter for customization:
+The plugin can take an arbitrary number of string arguments, and can be the 
+target of a pipeline that produces strings as well. This is most commonly 
+combined with a JSON filter for customization:
 
 ```liquid
 {{ tree "foo" | explode | toJSON | plugin "my-plugin" }}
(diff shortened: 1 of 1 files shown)


pearkes:
Thanks for this!

PR #1182, opened 2019-02-15
Add Vault agent token file to configuration

kyhavlov:
This PR takes the work done by @izhukov in #1153, makes some tweaks to the readme/tests and fixes a couple minor merge conflicts.


config/vault.go
```diff
@@ -233,18 +246,26 @@ func (c *VaultConfig) Finalize() {
 	}
 	c.SSL.Finalize()
 
+	// Order of precedence
+	// 1. `vault_agent_token_file` configuration value
+	// 2. `VAULT_TOKEN` environment variable
+	// 3. `token` configuration value
```
chrishoffman: I think it is a little odd to go from config file value, to env var, then back to config config file value.

config/vault.go
```diff
@@ -233,18 +246,26 @@ func (c *VaultConfig) Finalize() {
...
+	// 4. `~/.vault-token` file
```
chrishoffman: I don't think we should rely on this file or maybe worth further discussion.

kyhavlov: Agreed - I couldn't remember this file being part of the typical Vault defaults but wasn't sure whether to remove it.

config/vault.go
kyhavlov: Does `token` > `vault_agent_token_file` contents > `VAULT_TOKEN` make more sense? I couldn't find a standard order in the Vault docs for this situation.

config/vault.go
chrishoffman: I just checked the code and I am pretty sure the token helper, which uses the .vault-token file, is only used by the CLI and not the default API config.  I would leave it out since it seems to only be used with interactive users.

config/vault.go
chrishoffman: I think that makes more sense but now am also questioning it.  Maybe explicit tokens should take precedence over the file since that would probably be a more explicit operation.  I don't have strong opinions, so whatever seems to make the most sense.

kyhavlov: I think maybe `vault_agent_token_file` > `token` > `VAULT_TOKEN` is the way to go. If a `vault_agent_token_file` is being provided it's presumable more up to date than the static `token` field, which should take precedence over the env var.

atheiman:
@kyhavlov why was reading vault token from `~/.vault-token` (or any path) and consul-template renewing the token removed?

To expand on this, my use case is a vault login initContainer in a kubernetes pod runs `vault login` and stores the token to a volume to be shared with a consul-template sidecar container. the consul-template sidecar container uses the token from `vault login` initContainer to render secrets for an app container to use.

For this to work, one of these has to happen:
- consul-template needs to be able to read the token from a file and renew the token (the functionality that was removed)
- consul-template needs to be able to login to vault using the kubernetes auth method (i dont suggest implementing that)
- vault-agent also has to be run in a sidecar container to update the token file
  - the vault login init container still has to be there so that the token file is created before consul-template starts
- the initial vault token has to be passed to consul-template as an env var (change container command to `VAULT_TOKEN=$(cat /.vault-token) consul-template ...`) or append a flag to the command (`-token=$(cat /.vault-token)`

Does what im asking make sense?

atheiman:
@jefferai I saw you also had some comments on other issues last year suggesting functionality for reading vault token from a file and renewing it should be removed.. Maybe you have a suggestion for my question above that Im not thinking about?

jefferai:
@atheiman I don't really know anything about this change, but it seems like the difference is just that now you can specify where the file is read from (including ~/.vault-token) rather than a hardcoded path, which seems like a win.

atheiman:
but now if you specify to read the token from a file, consul-template cannot renew the token

jurgenweber:
I am with @atheiman here, it seems the this has broken the functionality of having consul-template renew its Vault token which is sorely needed if you have wish consul-template to dynamically keep the secrets updated.

I am not sure why you have made -vault-agent-token-file mutually exclusive to being able to renew a Vault token 

```
	if c.VaultAgentTokenFile != nil {
		c.Token = stringFromFile([]string{*c.VaultAgentTokenFile}, "")
		c.RenewToken = Bool(false)
	}
```  or why you have removed it reading `~/.vault-token` on launch.

```
		if StringVal(c.Token) == "" {
			if homePath != "" {
				c.Token = stringFromFile([]string{
					homePath + "/.vault-token",
				}, "")
			}
		}
```
? Is there a security concern or can we revert one and rls 0.20.1?

goffinf:
As one of the contributors on a related thread that noted this change in behaviour late last year (and struggled to find any clearly documented stance from HashiCorp), like @atheiman  and @jurgenweber removing the ability for consul-template to auto-renew the vault client auth token is a significant retrograde step. Using an initContainer and consul-templates sidecar is a *really* common pattern that allows the security posture of an application to be maintained whilst not burdening the application developer with how that happens. Time and use limited tokens are central to any approach to minimising the potential blast radius of a vulnerability, so we should be doing everything possible to move that logic into standard implementation patterns where the likelihood of an incorrect or path of least resistance approach is avoided. This isn’t a ‘nice to have’, it isn’t just a productivity aid, it’s an essential requirement for any serious CISO operation, mine included.

sethvargo:
🙋‍♂ as the original author of this tool, I can attest that the decision to no longer read from `~/.vault-token` is a pretty horrible regression...

devlounge:
Please bring this token renewal feature back

jurgenweber:
yeah, so PR above to bring this back. Let me know if I got it right (or wrong). :)

PR #1183, opened 2019-02-15
Wait longer when refreshing secrets without renewable leases

### LINKED ISSUE #1126: Non-renewable leases aren't used to full potential
### Consul Template version

```consul-template v0.19.4 (68b1da2)```

### Configuration

```liquid
{{with secret "aws/sts/somerole" "ttl=12h"}}
AWS_KINESIS_ACCESS_KEY={{.Data.access_key}}
AWS_KINESIS_SECRET_KEY={{.Data.secret_key}}
AWS_KINESIS_SECURITY_TOKEN={{.Data.security_token}}
{{end}}
```

### Command

```shell
consul-template -template ct.hcl -dry -log-level=trace
```

### Debug output

```
2018/07/24 15:54:57.621191 [TRACE] vault.write(aws/sts/somerole -> 54d618e4): secret is not renewable, sleeping for 3h3m17.348840674s
```

### Expected behavior

Non-renewable vault lease would cause re-render of template towards the end of it's TTL.

### Actual behavior

Template is re-rendered roughly 1/3 thru the leases TTL.  This looks to be caused by the logic in [vault_common.go:67](https://github.com/hashicorp/consul-template/blob/master/dependency/vault_common.go#L67).  Our use case is actually the template stanza in Nomad job specs.  When injecting AWS STS values via the template it's causing the template to render every 2-4h on a 12h TTL which is causing unnecessary churn of instances.

A quick test of checking the renewability of the lease in `vaultRenewDuration()` appears to do the right thing:
```go
if vaultSecretRenewable(s) {
    // Renew at 1/3 the remaining lease. This will give us an opportunity to retry
    // at least one more time should the first renewal fail.
    sleep = sleep / 3.0

    // Use a randomness so many clients do not hit Vault simultaneously.
    sleep = sleep * (rand.Float64() + 1) / 2.0
} else {
    // For non-renewable leases setting renew duration to use much of the secret lease as possible.
    // Ranging 85%-99% so many clients do not hit Vault simultaneously.
    sleep = sleep * (.85 + rand.Float64()*0.14)
}
```

If this is a valid issue and the solution is on the right track, I can add some tests around it and do a PR.

### Steps to reproduce

1. Setup a vault mount with non-renewable leases (AWS with STS is a good example)
2. Use template with non-renewable leases in it

kyhavlov:
This PR changes the behavior around refreshing secrets with non-renewable leases to wait for 85-95% of the lease time (rather than 1/6-1/3rd, as with renewable leases). We could also probably increase the range on renewable leases as well to something like 25-35% instead of 16-33%, to eliminate some unnecessarily early refreshing.

Fixes #1126.

### DIFF
--- a/dependency/vault_common.go
+++ b/dependency/vault_common.go
@@ -81,12 +81,19 @@ func vaultRenewDuration(s *Secret) time.Duration {
 	// Convert to float seconds.
 	sleep := float64(time.Duration(base) * time.Second)
 
-	// Renew at 1/3 the remaining lease. This will give us an opportunity to retry
-	// at least one more time should the first renewal fail.
-	sleep = sleep / 3.0
+	if vaultSecretRenewable(s) {
+		// Renew at 1/3 the remaining lease. This will give us an opportunity to retry
+		// at least one more time should the first renewal fail.
+		sleep = sleep / 3.0
 
-	// Use a randomness so many clients do not hit Vault simultaneously.
-	sleep = sleep * (rand.Float64() + 1) / 2.0
+		// Use a randomness so many clients do not hit Vault simultaneously.
+		sleep = sleep * (rand.Float64() + 1) / 2.0
+	} else {
+		// For non-renewable leases set the renew duration to use much of the secret
+		// lease as possible. Use a stagger over 85%-95% of the lease duration so that
+		// many clients do not hit Vault simultaneously.
+		sleep = sleep * (.85 + rand.Float64()*0.1)
+	}
 
 	return time.Duration(sleep)
 }
--- a/dependency/vault_common.go
+++ b/dependency/vault_common.go
@@ -86,7 +86,7 @@ func vaultRenewDuration(s *Secret) time.Duration {
 		// at least one more time should the first renewal fail.
 		sleep = sleep / 3.0
 
-		// Use a randomness so many clients do not hit Vault simultaneously.
+		// Use some randomness so many clients do not hit Vault simultaneously.
 		sleep = sleep * (rand.Float64() + 1) / 2.0
 	} else {
 		// For non-renewable leases set the renew duration to use much of the secret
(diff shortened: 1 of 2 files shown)


dependency/vault_common_test.go
```diff
@@ -1,5 +1,21 @@
...
+	renewable := Secret{LeaseDuration: 100, Renewable: true}
+	renewableDur := vaultRenewDuration(&renewable).Seconds()
+	if renewableDur < 16 || renewableDur >= 34 {
+		t.Fatalf("renewable duration is not within 1/6 to 1/3 of lease duration: %f", renewableDur)
+	}
+
+	nonRenewable := Secret{LeaseDuration: 100}
+	nonRenewableDur := vaultRenewDuration(&nonRenewable).Seconds()
+	if nonRenewableDur < 85 || nonRenewableDur > 95 {
+		t.Fatalf("renewable duration is not within 85%% to 95%% of lease duration: %f", nonRenewableDur)
```
pearkes:
```suggestion
		t.Fatalf("renewable duration is not within 85% to 95% of lease duration: %f", nonRenewableDur)
```

dependency/vault_common.go
```diff
@@ -81,12 +81,19 @@ func vaultRenewDuration(s *Secret) time.Duration {
...
-	sleep = sleep / 3.0
-
-	// Use a randomness so many clients do not hit Vault simultaneously.
-	sleep = sleep * (rand.Float64() + 1) / 2.0
+	if vaultSecretRenewable(s) {
+		// Renew at 1/3 the remaining lease. This will give us an opportunity to retry
+		// at least one more time should the first renewal fail.
+		sleep = sleep / 3.0
+
+		// Use a randomness so many clients do not hit Vault simultaneously.
```
pearkes:
```suggestion
		// Use some randomness so many clients do not hit Vault simultaneously.
```

dependency/vault_common_test.go
kyhavlov: This is intentional, the double % is to escape the character within the format string

pearkes: TIL!

PR #1191, opened 2019-02-25
added "light" docker image

### LINKED ISSUE #1190: Feature: consul-template docker image suitable for Kubernetes PodShareProcessNamespace feature
I'm trying to use consul-template as a sidecar container to manage config for an app container in the same pod. Everything works well except the `scratch` and `alpine` images provided for consul-template make this more difficult than it should be.

Below is the relevant consul-template config, `/consul-template-config/` is a `configMap` mount containing the consul-template config and the app config file template and `/app-config/` is an `emptyDir` mount to share the generated app config file with the app container.

```hcl
template {
  # Render app config template to be shared with app container
  source = "/consul-template-config/app_config.py.ctmpl"
  destination = "/app-config/app_config.py"
  # Signal app process to reload config file when it is generated
  command = "/bin/sh -c \"kill -HUP $(pidof app)\""
}
```

#### Problem with the `alpine` image

Running `kill -HUP $(pidof app)` results in `sh: can't kill pid 53: Operation not permitted`. I think this is because `consul-template` doesnt run as `root` and has no permission to signal the app container process. The way I know to give the consul-template user permission to do this is with `sudo`, which is not installed or configured for `consul-template` (and I don't think belongs in a container).

#### Problem with the `scratch` image

`scratch` doesn't have the needed binaries installed to find and signal the app container process.

#### Solution

I'd propose a new docker image tagged `light` published alongside `scratch` and `alpine`. This image would be very simple, it should just start from busybox or alpine and add the binary at `/consul-template`. This is the image I'm currently using and its working for the use case I've described:

```Dockerfile
FROM busybox

RUN wget --output-document /consul-template.tgz https://releases.hashicorp.com/consul-template/0.20.0/consul-template_0.20.0_linux_amd64.tgz && \
    tar xvzf /consul-template.tgz && \
    rm -v /consul-template.tgz

ENTRYPOINT ["/consul-template"]
```

atheiman:
This new docker image is suitable for use as a sidecar container in a Kubernetes pod with the `PodShareProcessNamespace` feature. There is more information on why the current images are not a good fit in the linked issue.

Closes #1190

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ LD_FLAGS ?= \
 	-X ${PROJECT}/version.GitCommit=${GIT_COMMIT}
 
 # List of Docker targets to build
-DOCKER_TARGETS ?= alpine scratch
+DOCKER_TARGETS ?= alpine light scratch
 
 # List of tests to run
 TEST ?= ./...
--- a/docker/light/Dockerfile
+++ b/docker/light/Dockerfile
@@ -1,0 +1,6 @@
+FROM busybox
+LABEL maintainer "Seth Vargo <seth@sethvargo.com> (@sethvargo)"
+
+ADD "./pkg/linux_amd64/consul-template" "/"
+
+ENTRYPOINT ["/consul-template"]
--- a/docker/light/Dockerfile
+++ b/docker/light/Dockerfile
@@ -1,6 +1,6 @@
-FROM busybox
+FROM alpine:latest
 LABEL maintainer "Seth Vargo <seth@sethvargo.com> (@sethvargo)"
 
-ADD "./pkg/linux_amd64/consul-template" "/"
+ADD "./pkg/linux_amd64/consul-template" "/bin/consul-template"
 
-ENTRYPOINT ["/consul-template"]
+ENTRYPOINT ["/bin/consul-template"]


atheiman:
@pearkes @kyhavlov you two seem to be the most active contributors to this - could you take a few minutes to review this contribution?

pearkes:
@atheiman thanks for the contribution and the idea! I don't see any concern with adding this additional target, given it has no dependencies and doesn't add any management overheard I can think of. We'll be doing a Docker release soon (#1192) and this should be available following that.

atheiman:
If anyone finds this, the `light` tag has been released: https://hub.docker.com/r/hashicorp/consul-template/tags

```
$ docker run --rm -it hashicorp/consul-template:light -version
consul-template v0.20.0 (b709612c)
$ docker run --rm -it --entrypoint=/bin/sh hashicorp/consul-template:light
/ # which consul-template
/bin/consul-template
/ # cat /etc/os-release
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.9.0
PRETTY_NAME="Alpine Linux v3.9"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://bugs.alpinelinux.org/"
```

PR #1193, opened 2019-03-01
docker: try a variety of gpg key servers

pearkes:
This makes the Docker build process more reliable by providing more than one keyserver.

The same approach is used in https://github.com/hashicorp/consul-k8s/blob/c6b378a7471cebfad947006e6cddc3082bf05081/build-support/docker/Release.dockerfile#L29-L39.


bofm:
The problem may be caused by the keyservers being unreachable via ipv6. If that is the case, using `ipv4.pool.sks-keyservers.net` could solve the problem.

pearkes:
Thanks @bofm – this PR does resolve the problem from my experience vs. just the MIT keyserver. Good note on IPV6, if this change fails to work perhaps we should try adding that host as well.

PR #1208, opened 2019-04-22
Migrate to gomod

syndbg:
Goals:

* Migrate `consul-template` to go modules. This fixes troubles for anyone attempting to use this project as a library.
* Update `Makefile` to support above go modules change. 
* Bump Travis and Golang image used for building to latest Golang 1.12.4 . From what I've seen so far in the project, supporting/testing against multiple Golang versions has never been a target, am I right? If I was to support both 1.11.x and 1.12.x it would be a huge trouble, so that's why I skipped it.
* Update documentation to remove outdated/stale information about the `Makefile` and `dep`.
* Migrate "download consul" logic from `.travis.yml` to a `Makefile` target. This eases everyone's project setup. Now to run the tests all I do is `sudo make download-consul` followed by `make test`. No need to download Consul manually by following a link. Also, I make sure that I test using the same Consul used in CI. Docs are also updated to reflect this.

Did I miss something?

Imo this is a huge milestone that would modernize the project a bit, since it's pretty stale by nowadays.


syndbg:
Ready for review. Test suite passes and overall looks good to me.

@pearkes

greut:
It looks good. Although I'd keep the `vendor` directory and avoid the `go.sum`. That way, it's backward compatible with older Go versions.

eikenb:
Hey @syndbg, thanks for this...

.. and while I appreciate this PR @greut is right in that we really want to refactor before doing this so as to not pull in Vault and all of its dependencies. I'll leave this open for now as you might want to just rework this once that work is done.

tustvold:
I presume the plan is to modify the circle CI configuration to run vault as a service? At the same time might it also be worth moving to doing the same with consul given that `consul/testutil` no longer exists and doing so removes the additional download step?  I'd be happy to submit a PR based on this one making those changes if that is of interest - it should just be some minor changes to `dependency_test.go` as it is the only thing that depends on anything outside of `vault/api`?

_As an aside depending directly on vault is made particularly fun at the moment by the fact vault currently depends on its API but at a different version to itself, so you end up with an ambiguous import (see [here](https://github.com/hashicorp/vault/blob/master/go.mod#L85))._

eikenb:
I didn't use docker, as I don't like requiring it for basic tests as it's generally easier to wrap things in docker at a higher level if you don't integrate docker into things. Otherwise see https://github.com/hashicorp/consul-template/pull/1230 for details on how I implemented it.

eikenb:
@syndbg Thanks for doing this and even though I ended up doing this myself having your version around was handy for checking some things.

PR #1213, opened 2019-05-25
deps: bump/fix

greut:
I've taken a look to the `go.mod` branch and it seems it will require quite some work (in the dependency_test section).

This PR aims at fixing some dependencies by pinning them. The next chunk-load of changes would be to rewrite some of the code to use consul/api, consul/sdk and vault/api, vault/sdk rather than the main projects.


jefferai:
FYI there is internal work ongoing to port to go modules, which includes this and the other items you mentioned.

greut:
@jefferai awesome, I'll be patient then. Cheers,

PR #1218, opened 2019-06-19
convert from travis to circleci for CI

eikenb:


eikenb:
I keep hoping I can get circleci going in a branch and it never works as I intend. :(

PR #1219, opened 2019-06-20
various Makefile fixes and updates

eikenb:
* use `go env` to get GOPATH and handle multiple paths
* changes to build, dist for easier testing
* remove bashisms
* add pristine build setup for release
* use ./... in tests (removed old vendor workaround)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,10 +3,9 @@ MKFILE_PATH := $(lastword $(MAKEFILE_LIST))
 CURRENT_DIR := $(patsubst %/,%,$(dir $(realpath $(MKFILE_PATH))))
 
 # Ensure GOPATH
-GOPATH ?= $(HOME)/go
-
-# List all our actual files, excluding vendor
-GOFILES ?= $(shell go list $(TEST) | grep -v /vendor/)
+GOPATH ?= $(shell go env GOPATH)
+# assume last entry in GOPATH is home to project
+GOPATH := $(lastword $(subst :, ,${GOPATH}))
 
 # Tags specific for building
 GOTAGS ?=
@@ -15,7 +14,7 @@ GOTAGS ?=
 GOMAXPROCS ?= 4
 
 # Get the project metadata
-GOVERSION := 1.11.5
+GOVERSION := 1.12.5
 PROJECT := $(CURRENT_DIR:$(GOPATH)/src/%=%)
 OWNER := $(notdir $(patsubst %/,%,$(dir $(PROJECT))))
 NAME := $(notdir $(PROJECT))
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -114,15 +117,19 @@ dev:
...
 	@echo "           be signed. Set the environment variable GPG_KEY to the ID of"
 	@echo "           the GPG key to continue."
 	@exit 127
 else
-	@$(MAKE) -f "${MKFILE_PATH}" _cleanup
-	@$(MAKE) -f "${MKFILE_PATH}" -j4 build
-	@$(MAKE) -f "${MKFILE_PATH}" _compress _checksum _sign
+	@$(MAKE) -f "${MKFILE_PATH}" _sign
 endif
 .PHONY: dist
```
banks: Should this now be `.PHONY release`?

Makefile
```diff
@@ -167,20 +173,23 @@ $(foreach target,$(DOCKER_TARGETS),$(eval $(call make-docker-target,$(target))))
 # test runs the test suite.
 test:
 	@echo "==> Testing ${NAME}"
-	@go test -timeout=30s -parallel=20 -tags="${GOTAGS}" ${GOFILES} ${TESTARGS}
+	@go test -timeout=30s -parallel=20 -tags="${GOTAGS}" ./... ${TESTARGS}
```
banks: This is probably fine for this tool. In Consul we have `$GOTEST_PKGS` which defaults to this but allows running only subpackage tests easily which is important in a large codebase. I don't know if it's really important here though as there are relatively few packages etc.

Makefile
```diff
@@ -167,20 +173,23 @@ $(foreach target,$(DOCKER_TARGETS),$(eval $(call make-docker-target,$(target))))
...
 test-race:
 	@echo "==> Testing ${NAME} (race)"
-	@go test -timeout=60s -race -tags="${GOTAGS}" ${GOFILES} ${TESTARGS}
+	@go test -timeout=60s -race -tags="${GOTAGS}" ./... ${TESTARGS}
 .PHONY: test-race
 
 # _cleanup removes any previous binaries
 _cleanup:
 	@rm -rf "${CURRENT_DIR}/pkg/"
 	@rm -rf "${CURRENT_DIR}/bin/"
```
banks: nit: not your change but no `.PHONY _cleanup` here.

Makefile
eikenb: Thanks for the reviews Paul. :)

My thought was that the `make test` call was primarily for automated testing (circleci, etc) and that if you were working on things and wanted to run specific tests you'd just use `go test` directly. 

Though I can see where that might come in handy if you are working with the same set of things over and over and want to just set them as environment variables. I'll consider that when I start working on the test suite to see if it would be handy.

Makefile
eikenb: Doh! Copied it for the GPG_KEY stuff and forgot to re-name that bit. I'll get that fixed.

Thanks.

Makefile
eikenb: I forget those sometimes as well as they are ignored by shell completion by default so I don't think of them being real targets. But they are legal and AFAIK skipping them is only a convention.

TLDR; I'll add it.

PR #1223, opened 2019-06-27
Added jq to be installed in the docker image

Gaardsholt:
Small change, just think that it would be great to have jq installed

### DIFF
--- a/docker/alpine/Dockerfile
+++ b/docker/alpine/Dockerfile
@@ -14,7 +14,7 @@ RUN addgroup consul-template && \
     adduser -S -G consul-template consul-template
 
 # Set up certificates, our base tools, and Consul Template (CT).
-RUN apk add --no-cache ca-certificates curl gnupg libcap openssl && \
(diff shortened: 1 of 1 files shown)


greut:
you probably ain't gonna need it.

Gaardsholt:
@greut I need it ;)

greut:
then, it's probably easier to derive and host your own image.

Gaardsholt:
@hashicorp-cla I have signed the CLA - same day as I created the pull request

eikenb:
Hey @Gaardsholt, thanks for taking the time to submit this.

I'd prefer to keep the docker images restricted in scope to contain only what is absolutely necessary to run CT. This makes the base image smaller and simpler to work with while also making it less burdensome when someone, like in your case, wants to customize them as they aren't already bloated with a bunch of stuff they don't need.

In fact this reminds me I need to review what is in the existing images as I think they already have a couple unnecessary things installed.

Issue #1209, opened 2019-04-24
0.20.0 is failing to render templates before running exec


eedwards-sk:
### Overview

from 0.19.5 to 0.20.0, exec with daemon no longer works correctly

the sub process is launched before a template is rendered

### Consul Template version

0.20.0

### Expected behavior

What should have happened?

Template rendered before running the command

### Actual behavior

What actually happened?

Command is ran before the template is rendered, template never gets rendered

eedwards-sk:
related:

#1196
#1207

eikenb:
Hey @eedwards-sk, thanks for reporting the issue and taking the time to link in the other related issues.

When you had this issue did you have a `wait` set like #1196 and #1207 did? I can reproduce the issue with `wait` set, but not without.

Thanks.

eikenb:
I figured out this was probably not the once/wait issue as exec/wait have a separate issue. I'm going to leave it closed though as a duplicate of #1229

Issue #1207, opened 2019-04-19
Consul Template does not render templates before exec in Alpine Docker Image


ingshtrom:
### Consul Template version
v0.20.0

### Configuration

[Here is a repo](https://github.com/ingshtrom/consul-template-exec-bug) with all of the files necessary to reproduce. The README shows how to run it locally and in docker. 

### Expected behavior

They should both behave the way the local version of consul-template is working, which is each template is rendered and after all templates are rendered, then exec is run.

### Actual behavior

The docker version of consul-template (1) is not tested from what I can tell and (2) in Docker alpine, exec is run before the templates are actually rendered, which breaks because my exec expects the templates to be there.

### Steps to reproduce

See the linked repo README for instructions on running the reproduction

ingshtrom:
The template I was using was
```
log_level = "info"
reload_signal = "SIGHUP"
kill_signal = "SIGTERM"
max_stale = "30s"

wait {
  min = "1m"
  max = "5m"
}

consul {
  retry {
    enabled     = true
    attempts    = 15
    backoff     = "250ms"
    max_backoff = "10s"
  }
}

template {
	source      = "/local/in.tpl"
	destination = "/local/out.txt"
}

template {
	source      = "/local/in2.tpl"
	destination = "/local/out2.txt"
}

exec {
	command = "/local/test.sh"
}
```

When I remove the global `wait` everything works as expected for both local and in-container versions of consul-template. I believe it might be similar to #1196.

eikenb:
I'm trying to figure out why you'd want both `wait` and `once`. From what I've gathered so far wait/quiescence was designed to prevent templates from re-rendering in quick succession when a watched value is flapping. When using once this doesn't seem to make sense as you are only rendering once and would get the state at that time (ie. flapping wouldn't matter).

Basically I'm curious if you have a use case where both wait and once make sense together. Thanks.

ingshtrom:
Right, so my thought was that `wait` would be essentially ignored for `once` runs. 

The use case I have is related to rendering a HAProxy template. (bear with me)

So I have the application, HAProxy, which consumes multiple configurations which are rendered from Consul Template. When the application starts up, it needs the configurations to all be there, otherwise we get in a restart loop. So what I do is run Consul Template in `once` mode prior to starting HAProxy. After that, Consul Template runs normally to keep getting continual updates as Consul values change.

If you are familiar with Kubernetes, this looks like using the same container image for an `initContainer`, with the additional `-once` flag.

[...]

Issue #1196, opened 2019-03-05
v0.20.0 -once doesn't wait for templates to finish


errnoh:
### Consul Template version

consul-template v0.20.0 (b709612c)

### Configuration

```hcl
reload_signal = "SIGHUP"
kill_signal = "SIGINT"
max_stale = "10m"
log_level = "warn"
pid_file = "/tmp/consul-template.pid"

wait {
  min = "5s"
  max = "10s"
}

vault {
  address = "http://127.0.0.1:8200"
  grace = "5m"
  unwrap_token = false
  renew_token = false

  retry {
    enabled = true
    attempts = 0
    backoff = "250ms"
    max_backoff = "1m"
  }
}

template {
        contents = "{{ with secret \"pki_test/issue/foobar\" \"common_name=foobar.baz.consul\" \"alt_names=localhost\" \"ttl=10h\" \"ip_sans=192.168.1.2,127.0.0.1\" }}{{ .Data.certificate }}\n{{ .Data.issuing_ca }}{{ end }}"
        destination      = "/tmp/foobar.pem"
        perms = 0644
}
```

```liquid
# Template embedded in the config, but same result happens even if using a template file.
```

### Command

```shell
consul-template -once -log-level=trace -config="/tmp/consul-template.hcl"
```

### Debug output

```
2019/03/05 12:15:40.814135 [INFO] (runner) creating watcher
2019/03/05 12:15:40.814995 [INFO] (runner) starting
2019/03/05 12:15:40.815926 [INFO] creating pid file at "/tmp/consul-template.pid"
2019/03/05 12:15:40.816414 [DEBUG] (runner) running initial templates
2019/03/05 12:15:40.816695 [DEBUG] (runner) initiating run
2019/03/05 12:15:40.817154 [DEBUG] (runner) checking template c8d01b3b00a0457ee4318bc007dc95f4
2019/03/05 12:15:40.817833 [DEBUG] (runner) was not watching 1 dependencies
2019/03/05 12:15:40.818064 [DEBUG] (watcher) adding vault.write(pki_test/issue/foobar -> 3386df62)
2019/03/05 12:15:40.818956 [TRACE] (watcher) vault.write(pki_test/issue/foobar -> 3386df62) starting
2019/03/05 12:15:40.819121 [DEBUG] (runner) diffing and updating dependencies
[...]

PR #1227, opened 2019-07-03
once-wait-conflict

eikenb:
fix -once to ignore wait/quiescence timers

addresses issues: #1196, #1207, #1209


eikenb:
Thanks @freddygv! I updated the PR name to be more descriptive as suggested and added the related issues to the description (didn't use the Fixes syntax as that's already in a commit).

PR #1230, opened 2019-07-16
tests now use an external vault process

eikenb:
Replace use of core vault package for testing with an external vault
process. Previously each set of tests that needed vault would create a
vault service in the test process and all tests would use the default
'secret' kv path. Now it uses a separate kv secrets path for each set of
tests, setting the version for each.

The code for starting up the Vault process was based off what was
already used for Consul via its sdk/testutils package.

In the process various test conflicts were addressed and an external
process (consul and vault) process leak was fixed [1].

[1] the leak was caused by use of log.Fatal which calls os.Exit which
exist the process without running any cleanup code.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,13 +3,18 @@ jobs:
   build:
     environment:
       CONSUL_VERSION: 1.5.1
+      VAULT_VERSION: 1.1.3
     docker:
       - image: circleci/golang:latest
     working_directory: /go/src/github.com/hashicorp/consul-template
     steps:
       - checkout
       - run: |
-          sudo curl -sLo consul.zip https://releases.hashicorp.com/consul/${CONSUL_VERSION}/consul_${CONSUL_VERSION}_linux_amd64.zip
+          curl -sLo consul.zip https://releases.hashicorp.com/consul/${CONSUL_VERSION}/consul_${CONSUL_VERSION}_linux_amd64.zip
           unzip consul.zip
           sudo cp consul /usr/local/bin/
+      - run: |
+          curl -sLo vault.zip https://releases.hashicorp.com/vault/${VAULT_VERSION}/vault_${VAULT_VERSION}_linux_amd64.zip
+          unzip vault.zip
+          sudo cp vault /usr/local/bin/
       - run: make test
--- a/dependency/dependency_test.go
+++ b/dependency/dependency_test.go
@@ -1,51 +1,44 @@
 package dependency
 
 import (
-	"context"
 	"fmt"
 	"io/ioutil"
 	"log"
-	"net"
 	"os"
+	"os/exec"
 	"reflect"
+	"runtime"
 	"testing"
-	"time"
 
 	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/consul/testutil"
-	"github.com/hashicorp/vault/builtin/logical/pki"
-	"github.com/hashicorp/vault/builtin/logical/transit"
-	"github.com/hashicorp/vault/helper/namespace"
-	"github.com/hashicorp/vault/http"
-	"github.com/hashicorp/vault/logical"
-	"github.com/hashicorp/vault/physical/inmem"
-	"github.com/hashicorp/vault/vault"
-
-	hclog "github.com/hashicorp/go-hclog"
-	logicalKv "github.com/hashicorp/vault-plugin-secrets-kv"
+	vapi "github.com/hashicorp/vault/api"
 )
 
+const vaultAddr = "http://127.0.0.1:8200"
+const vaultToken = "a_token"
+
 var testConsul *testutil.TestServer
+var testVault *vaultServer
 var testClients *ClientSet
 
 func TestMain(m *testing.M) {
-	consul, err := testutil.NewTestServerConfig(func(c *testutil.TestServerConfig) {
-		c.LogLevel = "warn"
-		c.Stdout = ioutil.Discard
-		c.Stderr = ioutil.Discard
-	})
 	log.SetOutput(ioutil.Discard)
-	if err != nil {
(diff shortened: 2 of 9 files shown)


.circleci/config.yml
```diff
@@ -12,4 +13,8 @@ jobs:
           sudo curl -sLo consul.zip https://releases.hashicorp.com/consul/${CONSUL_VERSION}/consul_${CONSUL_VERSION}_linux_amd64.zip
           unzip consul.zip
           sudo cp consul /usr/local/bin/
+      - run: |
+          sudo curl -sLo vault.zip https://releases.hashicorp.com/vault/${VAULT_VERSION}/vault_${VAULT_VERSION}_linux_amd64.zip
```
greut:
```suggestion
          curl -sLo vault.zip https://releases.hashicorp.com/vault/${VAULT_VERSION}/vault_${VAULT_VERSION}_linux_amd64.zip
```

eikenb: Yeah... sudo isn't needed on either of those curl calls. It seems to be a remnant of a previous, half-baked version of the curl install that spread. I'll eliminate that and force push up a fixed version in a few.

Thanks.

dependency/dependency_test.go
```diff
@@ -58,18 +51,20 @@ func TestMain(m *testing.M) {
 		},
 	}
 
-	if err := testClients.consul.client.Agent().ServiceRegister(serviceMetaService); err != nil {
-		panic(err)
+	consul_agent := testClients.consul.client.Agent()
```
freddygv: Snake case names should be replaced with camel case

dependency/dependency_test.go
```diff
@@ -80,9 +75,79 @@ func TestMain(m *testing.M) {
...
+			c.Stderr = ioutil.Discard
+		})
+	if err != nil {
+		Fatalf("failed to start consul server: %v", err)
+	}
+	testConsul = consul
+}
+
+type vaultServer struct {
+	secrets_path string
```
freddygv: snake case -> camel case

dependency/dependency_test.go
```diff
@@ -80,9 +75,79 @@ func TestMain(m *testing.M) {
...
+	cmd          *exec.Cmd
+}
+
+func runTestVault() {
```
freddygv: Why not pass in `testing.T`?

dependency/vault_read_test.go
```diff
@@ -83,11 +83,20 @@ func TestNewVaultReadQuery(t *testing.T) {
...
 }
 
-func TestVaultReadQuery_Fetch(t *testing.T) {
+func TestVaultReadQuery_Fetch_KVv1(t *testing.T) {
 	t.Parallel()
 
-	clients, vault := testVaultServer(t)
-	defer vault.Stop()
+	clients, vault := testVaultServer("read_fetch_v1", "1")
+	secrets_path := vault.secrets_path
```
freddygv: snake case -> camel case

dependency/vault_list_test.go
```diff
@@ -68,13 +68,16 @@ func TestNewVaultListQuery(t *testing.T) {
 func TestVaultListQuery_Fetch(t *testing.T) {
 	t.Parallel()
 
-	clients, vault := testVaultServer(t)
-	defer vault.Stop()
+	clients, vault := testVaultServer("listfetch", "1")
+	secrets_path := vault.secrets_path
```
freddygv: snake case -> camel case

dependency/vault_read_test.go
```diff
@@ -218,18 +227,8 @@ func TestVaultReadQuery_Fetch(t *testing.T) {
...
-	vc := clients.Vault()
-	if err := vc.Sys().TuneMount("secret", api.MountConfigInput{
-		Options: map[string]string{
-			"version": "2",
-		},
-	}); err != nil {
-		t.Fatalf("Error tuning secrets engine: %s", err)
-	}
+	clients, vault := testVaultServer("read_fetch_v2", "2")
+	secrets_path := vault.secrets_path
```
freddygv: snake case -> camel case

dependency/dependency_test.go
eikenb: There is no `testing.T` instance available to pass in here as this is called from TestMain.

dependency/dependency_test.go
```diff
@@ -80,9 +75,79 @@ func TestMain(m *testing.M) {
...
+}
+
+func (v vaultServer) Stop() error {
+	if v.cmd != nil && v.cmd.Process != nil {
+		return v.cmd.Process.Signal(os.Interrupt)
+	}
+	return nil
+}
+
+func testVaultServer(secrets_path, version string) (*ClientSet, *vaultServer) {
```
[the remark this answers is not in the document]
eikenb: I think I was probably working on TestMain at the same time where I don't have the option for that (testing.M doesn't have Fatal) and just did the same thing there. But better to use the built-ins where possible... I'll make that change.

dependency/vault_read_test.go
eikenb: Changing the struct variable to camelCase... leaving local alone.

dependency/vault_list_test.go
eikenb: Changing the struct variable to camelCase... leaving local alone.

dependency/vault_read_test.go
eikenb: Changing the struct variable to camelCase... leaving local alone.

dependency/dependency_test.go
eikenb: In this case I think you are right... cameCase is better for struct variables, particularly as this is referenced in other tests.

dependency/dependency_test.go
eikenb: I don't agree with using camelCase for everything and snake_case for nothing. I think snake_case for local variables is perfectly fine.

PR #1180, opened 2019-02-09
Versioned KV support

kyhavlov:
This PR adds support for reading secrets from the v2 KV secrets engine in Vault.

Fixes #1121 
Fixes #1146

PR #1231, opened 2019-07-16
Ignore anonymous request failures checking kv status

notnoop:
When making unauthenticated secret lookup requests, ignore errors that
result from checking the secret mount type.  If it's a persistent error,
the error on the actual lookup.

This restores behavior prior to #1180 .

### DIFF
--- a/dependency/vault_common.go
+++ b/dependency/vault_common.go
@@ -225,6 +225,12 @@ func isKVv2(client *api.Client, path string) (string, bool, error) {
 			return "", false, nil
 		}
 
+		// anonymous requests may fail to access /sys/internal/ui path
+		// Vault v1.1.3 returns 500 status code but may return 4XX in future
+		if client.Token() == "" {
+			return "", false, nil
+		}
+
 		return "", false, err
 	}
 
--- a/dependency/vault_read_test.go
+++ b/dependency/vault_read_test.go
@@ -3,6 +3,8 @@ package dependency
 import (
 	"fmt"
 	"net/url"
+	"reflect"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


dependency/vault_common.go
```diff
@@ -225,6 +225,12 @@ func isKVv2(client *api.Client, path string) (string, bool, error) {
 			return "", false, nil
 		}
 
+		// anonymous requests may fail to access /sys/internal/ui path
+		// Vault v1.1.3 returns 500 status code but may return 4XX in future
+		if client.Token() == "" {
+			return "", false, nil
+		}
+
 		return "", false, err
```
notnoop: I've considered always suppressing errors, so we fallback to pre-#1180 behavior, but was worried about transient errors.  It will not be ideal if clients inconsistently wrap lookups as KVv2 and not, when hitting transient errors in this call; and debugging that will be very hard.  Open for better suggestions.

eikenb:
Hey @notnoop, thanks for the PR.

Running this test locally I found it failed every time with...

```
vault_read_test.go:441: expected a missing client token error but found: Error making API request.
        
        URL: GET http://127.0.0.1:34201/v1/auth/token/lookup-self
        Code: 403. Errors:
        
        * permission denied
```

Thinking about it I realized it was because I had VAULT_TOKEN set as an environment variable as part of my setup for some other testing. The vault client sets the client's token automatically if it finds this environment variable even if the token in it is not used or valid.

The documentation is pretty clear that setting the vault token is a... meaningful act and, as such, it conflicting with an alternative system isn't necessarily wrong. I just wanted to note this so you'd be aware and to be sure you had taken it into account.

Thanks.

notnoop:
@eikenb yes - I considered always suppressing errors from isKVv2 (or 403 errors specifically) and treating it like a kv1 lookup.  I decided against it in this PR for fear of transient errors.  Consider the case where consul-template process starts with lack of acl access to a specific secret (hence getting 403) but then granted access; the process would still access it as a KVv1 while newly launched processes will access as KVv2 secret.  I can come with few similar edge casing scenarios.

We can revamp how to do caching isKVv2 or being more finegrained in how to handle errors, but that's beyond the scope of our immediate need so I'd punt on it for another PR if the need arise.

eikenb:
Great! Thanks @notnoop for addressing my comment so fast.

I'll probably go ahead and merge this tomorrow so I can rework the test as I just completely re-worked the vault tests in #1230. But I should be able to rework that test quickly as I just  finished with that one and know what needs to be done.

PR #1232, opened 2019-07-17
Switch to using Go11Modules

prologic:


greut:
You'll probably need #1230 (see #1208). I tend to like having the vendors versioned as well via `go mod vendor`, `go build -mod=vendor`

prologic:
Is there really a point in committing the source code to dependencies today? `go mod` makes all this painless and trivial to manage dependencies without polluting your source tree IHMO.

jefferai:
Duplicate of #1208

greut:
@prologic Agreed, until something gets deleted by someone else. https://status.npmjs.org/incidents/41zfb8qpvrdj

PR #1233, opened 2019-07-17
Fix vault APi change for Skipping TLS Verification const

prologic:
Looks like this was changed in [6ce09bae6a8c443bfc64e0fd5e8461db79433a96](https://github.com/hashicorp/vault/commits/6ce09bae6a8c443bfc64e0fd5e8461db79433a96)


jefferai:
See https://github.com/hashicorp/vault/pull/7135

prologic:
D'oh :)

eikenb:
@prologic ..Thanks for bringing this to my attention. I added an issue so that I wouldn't forget about it before the next release.

And thanks @jefferai for jumping in.

PR #1238, opened 2019-07-24
fix test issues due to meta data added to consul

eikenb:
Several calls had version meta-data added in consul 1.5.2 [1]. This
fixes those tests and bumps circleci tests to use consul 1.5.2.

[1] https://github.com/hashicorp/consul/pull/5455

### DIFF
--- a/dependency/catalog_node_test.go
+++ b/dependency/catalog_node_test.go
@@ -108,6 +108,7 @@ func TestCatalogNodeQuery_Fetch(t *testing.T) {
 						Service: "consul",
 						Port:    testConsul.Config.Ports.Server,
 						Tags:    ServiceTags([]string{}),
+						Meta:    map[string]string{},
 					},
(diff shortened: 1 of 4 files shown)


eikenb:
Missed a test. Scrapping for now.

eikenb:
It wasn't a missing test but a race in another test that seems to show up more often using the newer consul. I did reproduce it using older versions, it just took a bit longer for it to happen.

eikenb:
@pearkes  .. The tests compared expected results with real results and the real results included new content without asking. IMO this is a fairly standard way to test things, so I'm not sure about the difference between "1.5.2 introducing a compatibility issue" and it being  "just the structure of the tests" as they both seem true to me.

One of the reasons I wanted to get this reviewed was to highlight the change in case people didn't consider all the ramifications of adding the version info by default.

PR #1241, opened 2019-07-26
init quiescence after all templates are rendered

### LINKED ISSUE #1229: With "wait" consul-template spawns the child process before rendering the template
### Consul Template version
`consul-template v0.20.0`
(The same setup works fine in `consul-template v0.19.5`)

### Configuration

/etc/consul-template.conf:

```hcl
consul {
  address = "consul.query.consul:28500"
  retry {
    enabled = true
    attempts = 12
    backoff = "250ms"
    max_backoff = "1m"
  }
}
reload_signal = "SIGHUP"
kill_signal = "SIGINT"
log_level = "warn"
pid_file = "/var/run/consul-template/consul-template.pid"
exec {
  command = "/usr/local/bin/pgpool -n -D"
  kill_signal = "SIGINT"
  kill_timeout = "10s"
}
template {
  source = "/usr/local/etc/pgpool.conf.tpl"
  destination = "/usr/local/etc/pgpool.conf"
  wait {
    min = "5s"
    max = "10s"
  }
}
```

### Command

```shell
consul-template -config /etc/consul-template.conf
```

### Expected behavior

After the wait time, it should render the template and then spawn the child process.

### Actual behavior

It spawns the child process before rendering the template (only when you use `wait` in the `template` block). It seems that it executes the command in the `exec` block and then waits for rendering the template.

eikenb:
Hold off on initializing wait/quiescence until all templates have been
rendered once. This fixes an issue with -exec firing before the template
has been rendered and would also make sure templates are rendered for
once (if that wasn't already fixed another way).

Fixes #1229

### DIFF
--- a/manager/runner_test.go
+++ b/manager/runner_test.go
@@ -780,6 +780,74 @@ func TestRunner_Start(t *testing.T) {
 		}
 	})
 
+	// Exec would run before template rendering if Wait was defined.
+	t.Run("exec-wait", func(t *testing.T) {
+		t.Parallel()
+
+		testConsul.SetKVString(t, "exec-wait-foo", "foo")
+
+		firstOut, err := ioutil.TempFile("", "foo")
+		if err != nil {
+			t.Fatal(err)
+		}
+		os.Remove(firstOut.Name())       // remove ioutil created file
+		defer os.Remove(firstOut.Name()) // remove template created file
+
+		c := config.DefaultConfig().Merge(&config.Config{
+			Consul: &config.ConsulConfig{
+				Address: config.String(testConsul.HTTPAddr),
+			},
+			Wait: &config.WaitConfig{
+				Min: config.TimeDuration(5 * time.Millisecond),
+				Max: config.TimeDuration(10 * time.Millisecond),
+			},
+			Exec: &config.ExecConfig{
+				// `cat filename` would fail if template hadn't rendered
+				Command: config.String(`cat ` + firstOut.Name()),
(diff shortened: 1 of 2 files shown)


eikenb:
There is no test case for it currently. I plan on fixing that before merging this PR. Just wanted to get it out in front of people sooner.

eikenb:
@pearkes .. I added the test to reproduce it.

manager/runner_test.go
```diff
@@ -780,6 +780,74 @@ func TestRunner_Start(t *testing.T) {
...
+			Consul: &config.ConsulConfig{
+				Address: config.String(testConsul.HTTPAddr),
+			},
+			Wait: &config.WaitConfig{
+				Min: config.TimeDuration(5 * time.Millisecond),
+				Max: config.TimeDuration(10 * time.Millisecond),
+			},
+			Exec: &config.ExecConfig{
+				// `cat filename` would fail if template hadn't rendered
+				Command: config.String(`cat ` + firstOut.Name()),
```
pearkes: Neat 😄. Took me a second to realize that this is the vital part of the test but totally makes sense. Are there any cross-platform concerns here? Not sure we can really get around it given we're testing `exec`...

eikenb: The `cat` command is POSIX standard which is about as much as I think is possible for cross-platform support.

pearkes: Yeah makes sense.

PR #1243, opened 2019-08-01
Knob to blacklist some template functions

notnoop:
Exposes a knob for consul-template library users to disable some template functions.  Functions like `env` and `plugin` may access sensitive values on host and are best disabled unilaterally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -495,6 +495,10 @@ template {
   left_delimiter  = "{{"
   right_delimiter = "}}"
 
+  # These are functions that are not permitted in the template. If a template
+  # includes one of these functions, it will exit with an error.
+  function_blacklist = []
+
   # This is the `minimum(:maximum)` to wait before rendering a new template to
   # disk and triggering a command, separated by a colon (`:`). If the optional
   # maximum value is omitted, it is assumed to be 4x the required minimum value.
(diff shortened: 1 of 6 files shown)


tgross:
I've pushed a commit on top of this with the following changes:
* moves the configuration down to the template level (which works fine for Nomad's use case)
* ensures we have access to change the blacklist in the configuration structs
* changes the field name per recommendations from @eikenb 

---

Here's what this looks like from the user's perspective:

demo.hcl:

```hcl
template {
  source = "./demo.conf",
  function_blacklist = ["plugin"],
}
```

demo.conf:
```conf
{{ env "SECRET_STUFF" }}
{{ plugin "/bin/cat" "./secret_file.txt" }}
```

demo:
```sh
$ echo 'SUPER_SECRET_SHOULD_NOT_SEE' > ./secret_file.txt
$ make dev
==> Installing consul-template for darwin/amd64

$ SECRET_STUFF=MYSECRET consul-template -log-level=debug -config=./demo.hcl -dry -once
2019/08/02 14:57:43.041320 [ERR] (cli) ./demo.conf: execute: template: :2:3: executing "" at <plugin "/bin/cat" "./secret_file.txt">: error calling plugin: function is disabled
```

eikenb:
@tgross Do you have time to update the config example in the README.md with the function_blacklist? I think between the delimiters and the wait is the best spot.

tgross:
@eikenb I've done that. How's that look?

eikenb:
Looks good. Thanks.

PR #1249, opened 2019-08-07
add path sandboxing to `file` function

tgross:
If a sandbox path is provided in the configuration, the `file` function will prefix the path parameter and prevent relative paths from falling outside the sandbox.

---

@eikenb I've already vendored this into https://github.com/hashicorp/nomad/pull/6075 for discussion there, but for now this is a draft so that we can chat about the implementation and what you think about the config UX.

---

Example of use:

ct-demo.hcl
```hcl
template {
  source = "./ct-demo.conf"
  sandbox_path = "/Users/tim"
}
```

ct-demo.conf
```
{{ file "/etc/passwd" }}
```

Results:
```
$ consul-template -config=./ct-demo.hcl -once -dry
2019/08/07 15:56:39.325342 [ERR] (view) file(/Users/tim/etc/passwd): stat /Users/tim/etc/passwd: no such file or directory (exceeded maximum retries)
2019/08/07 15:56:39.325365 [ERR] (runner) watcher reported error: file(/Users/tim/etc/passwd): stat /Users/tim/etc/passwd: no such file or directory
2019/08/07 15:56:39.325639 [ERR] (cli) file(/Users/tim/etc/passwd): stat /Users/tim/etc/passwd: no such file or directory
```

Removing the `sandbox_path` field has the expected result: dumping the contents of that file to the terminal.


template/template.go
```diff
@@ -138,10 +149,6 @@ type ExecuteInput struct {
 	// Values specified here will take precedence over any values in the
 	// environment when using the `env` function.
 	Env []string
-
-	// BlacklistedFunctions is a set of functions to be disabled
-	// when executing the template
-	BlacklistedFunctions []string
```
eikenb: :+1:

eikenb:
So I think this basically looks fine. Tests, documentation, etc.

Related to the sym-links. It would be pretty trivial to just add a filepath.EvalSymlinks call just before the filepath.Rel call to eliminate sym-linking outside of the box.

tgross:
> Related to the sym-links. It would be pretty trivial to just add a filepath.EvalSymlinks call just before the filepath.Rel call to eliminate sym-linking outside of the box.

Ok, cool. I'll add that as a belt-and-suspenders for Nomad's own isolation (which we wouldn't enjoy for the `exec` driver anyways).

tgross:
@eikenb I've added the symlink walking check, which required a small rework of how the tests were being run to include real files on the file system under `testdata/` but I think I'm happy with the result.

eikenb:
LGTM

PR #1251, opened 2019-08-09
makefile: add support for arm64 platfrom

liusdu:
it works well  on my env
```
# file consul-template
consul-template: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, Go BuildID=yBBsvpv1PEzSZNVawlQb/nVRSLofiQGfncQaZ4euB/7hXFeZH6_F2NAJC0gap1/f3JuZNrjcmTkEhzmJfzq, stripped
[root@localhost ~]# ./consul-template -v
consul-template v0.21.0 (0d999b3)

```

Signed-off-by: Liu Hua <sdu.liu@huawei.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,8 +27,8 @@ GOARCH ?= $(shell go env GOARCH)
 
 # Default os-arch combination to build
 XC_OS ?= darwin freebsd linux netbsd openbsd solaris windows
-XC_ARCH ?= 386 amd64 arm
-XC_EXCLUDE ?= darwin/arm solaris/386 solaris/arm windows/arm
+XC_ARCH ?= 386 amd64 arm arm64
+XC_EXCLUDE ?= darwin/arm solaris/386 solaris/arm windows/arm darwin/arm64 freebsd/arm64 netbsd/arm64 openbsd/arm64 solaris/arm64 windows/arm64
 
 # GPG Signing key (blank by default, means no GPG signing)
 GPG_KEY ?=


eikenb:
Hey @liusdu, thanks for taking the time to submit the PR. I'd be happy to merge this but can't without the CLA. Can you sign it?

liusdu:
Hi  @eikenb  I cannot find what is wrong with it.  Actually I signed the cla. 
I think you can open a new pr and resend this code to fix this problem. I am ok about this~

eikenb:
@liusdu, that is very strange. I'm not sure what would have kept it from registering that you signed the CLA. It is a github thing that they don't provide much insight into.

I guess opening a new PR is the quickest/simplest path forward. Sorry for the inconvenience and thanks.

eikenb:
Hey @liusdu, a team-mate pointed out that while this PR was made by you the commit itself was made by @aarch64 and github won't consider the CLA valid until the person who did the commit has signed it. So if you could get @aarch64 to sign the CLA that should fix it. Thanks.

liusdu:
@evanphx Done! thank you very much!

eikenb:
FYI. I'm going to be changing all the ARM builds to enable CGO due to upstream Go bug: https://github.com/golang/go/issues/32912. It hasn't been enabled before now, but wasn't a problem as I don't think anyone used them (else I suspect someone would have filed an issue). But I assume that you submitted this as you would want to use them. So working binaries are probably needed.

eikenb:
Didn't notice until after I merged it but the arm64 exclude entries (eg. freebsd/arm64) match both the arm and arm64 entries, preventing both.

After some playing around I decided to just keep the more restrictive arm builds (only build arm for Linux) that you have and just removed the '*/arm' entries. I think this is pretty safe as I don't think anyone is using the any of the arm builds as they should all be broken due to that golang bug.

If this ends up being an issue it is easy enough to rework the builds.

PR #1253, opened 2019-08-09
Read paths unmodified if KVv2 status check fails

notnoop:
When vault.read fails to determine whether a path is KVv2, assume it's
KVv1 and read from the path as passed from user.

Here we fallback to behavior prior to https://github.com/hashicorp/consul-template/pull/1180,
so KVv2 check is for enhancement only.  If user truely lacks access, the
secret lookup call would fail and we report error back.

A customer hit a regression after they upgraded nomad with updated
consul-template, where their templates failed to read approle info
because isKVv2 returned an error.  The case is reproduced in
`TestVaultReadQuery_Fetch_NonSecrets` test.

### DIFF
--- a/dependency/vault_read.go
+++ b/dependency/vault_read.go
@@ -147,10 +147,10 @@ func (d *VaultReadQuery) readSecret(clients *ClientSet, opts *QueryOptions) (*ap
 	if d.isKVv2 == nil {
 		mountPath, isKVv2, err := isKVv2(vaultClient, d.rawPath)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


notnoop:
FWIW, I was very conservative in https://github.com/hashicorp/consul-template/pull/1231 change - this fully rollback behavior to prior to #1180 more generally to all errors and logs them.  I kept the special casing for anonymous requests to avoid unnecessary warn logging.

eikenb:
I like this change better than what cv and I talked about yesterday. Simple and clear.
:+1:

PR #1254, opened 2019-08-09
file sandbox: check but don't alter paths passed to 'file'

tgross:
During a discussion with @notnoop about https://github.com/hashicorp/nomad/pull/6075, we discovered that _altering_ the path being passed into the `file` function was going to cut off a lot of existing uses of consul-template in Nomad, particularly around the use of the `NOMAD_TASK_DIR` variable.

This PR provides a check but doesn't prefix the path parameter.

cc @schmichael @eikenb

### DIFF
--- a/template/funcs.go
+++ b/template/funcs.go
@@ -104,7 +104,7 @@ func fileFunc(b *Brain, used, missing *dep.Set, sandboxPath string) func(string)
 		if len(s) == 0 {
 			return "", nil
 		}
-		s, err := sandboxedPath(sandboxPath, s)
+		err := pathInSandbox(sandboxPath, s)
 		if err != nil {
 			return "", err
 		}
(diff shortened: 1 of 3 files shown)


template/funcs.go
```diff
@@ -1166,26 +1169,22 @@ func blacklisted(...string) (string, error) {
...
-// path falls outside the sandbox.
-func sandboxedPath(sandbox, s string) (string, error) {
-	path, err := filepath.Abs(filepath.Join(sandbox, s))
-	if err != nil {
-		return "", err
-	}
+// pathInSandbox determines whether a provided path falls within the sandbox.
+// returns an error only if the file can't be evaluated (missing, invalid
+// symlink, etc.)
+func pathInSandbox(sandbox, path string) (bool, error) {
```
eikenb: Returning both a bool and an error is redundant here as you always return false with an error and true when there is no error. So when you check `!ok` above, you would always use that error and ignore the returned one whenever this returned false.

I'd say skip the bool and just use the standard `if err != nil {return err}` on the result.

tgross: It was feeling semantically weird to return an error on a yes/no question. But I think you're right, let's do that.

tgross: Done!

tgross:
@eikenb https://circleci.com/gh/hashicorp/consul-template/126 is showing what looks like a flaky test but I don't have permissions to re-run on CircleCI.

eikenb:
@tgross Yeah.. that TestRunner_quiescence/snooze test has a race in it somewhere. It hasn't been a  priority as it is infrequent enough and I'm planning on reworking that runner code  anyways.

tgross:
Are we ok to merge then?

template/funcs.go
eikenb: Thanks. And yeah, I get the semantic itch for having it return a bool to a yes/no question-y function. Maybe with a different name or something it wouldn't feel wrong or if it could be reworked such that the error could be handled internally (just changing the bool result). But as is, I think simplifying it is better.

eikenb:
Yep. Going to take care of that in a few.

Issue #1252, opened 2019-08-09
Unable to write to KVv2


nixgadget:
Using `consul-template v0.21.0 (0d999b32)`

I have the following consul-template configuration following the [doc](https://github.com/hashicorp/consul-template#secret).

```
vault {
  ssl {
    ca_cert = "tls/ca.pem"
  }
  retry {
    backoff = "1s"
  }
}
template {
  {{ secret "secret/data/test/admin" "password=test" }}
}
```

However, it keeps throwing the error as if there is no request body.

```
[WARN] (view) vault.write(secret/data/test/admin -> 48e87fe5): vault.write(secret/data/test/admin -> 48e87fe5): Error making API request.

URL: PUT http://<ip>/v1/secret/data/test/admin
Code: 400. Errors:

* no data provided (retry attempt 1 after "250ms")
```

I thought the format is `{{ secret "<PATH>" "<DATA>" }}` but is it not the case ? 
Just as a suggestion it would be good to have an example in the doc too.

eikenb:
Reproduced this with just a template containing...
```
{{ secret "secret/data/test/admin" "password=test" }}
```
And ...
```
$ consul-template -template=template.tmpl
2019/08/09 23:48:39.199330 [WARN] (view) vault.write(secret/data/test/admin -> 48e87fe5): vault.write(secret/data/test/admin -> 48e87fe5): Error making API request.

URL: PUT http://127.0.0.1:8200/v1/secret/data/test/admin
Code: 400. Errors:

* no data provided (retry attempt 1 after "250ms")
2019/08/09 23:48:39.451967 [WARN] (view) vault.write(secret/data/test/admin -> 48e87fe5): vault.write(secret/data/test/admin -> 48e87fe5): Error making API request.

URL: PUT http://127.0.0.1:8200/v1/secret/data/test/admin
Code: 400. Errors:

* no data provided (retry attempt 2 after "500ms")
[...]
```

nixgadget:
@eikenb Thanks for looking into it. I debugged it a little further.

It seems though as the Vault api expects the format,
```
{
    "data": { 
        "password":"abcd"
    }
}
```

However, what consul-template attempts to push is `{"password":"abcd"}` when the following template is used,
`{{ secret "secret/data/test/admin" "password=abcd" }}`

nixgadget:
@eikenb if it helps was able to patch it locally on my end with,

```
dataNew := make(map[string]interface{})
dataNew["data"] = d.data
[...]

PR #1257, opened 2019-08-13
Issue1252: vault write broken

eikenb:
Writing to a Vault KV v1 and v2 were both broken.

### DIFF
--- a/dependency/vault_write.go
+++ b/dependency/vault_write.go
@@ -108,12 +108,14 @@ func (d *VaultWriteQuery) Fetch(clients *ClientSet, opts *QueryOptions) (interfa
 		return nil, nil, errors.Wrap(err, d.String())
 	}
 
-	// Print any warnings
-	printVaultWarnings(d, vaultSecret.Warnings)
-
-	// Create the cloned secret which will be exposed to the template.
-	d.vaultSecret = vaultSecret
-	d.secret = transformSecret(vaultSecret)
+	// vaultSecret == nil when writing to KVv1 engines
+	if vaultSecret != nil {
+		// Print any warnings
+		printVaultWarnings(d, vaultSecret.Warnings)
+		// Create the cloned secret which will be exposed to the template.
+		d.vaultSecret = vaultSecret
+		d.secret = transformSecret(vaultSecret)
+	}
 
 	return respWithMetadata(d.secret)
 }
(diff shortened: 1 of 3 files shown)


dependency/vault_write_test.go
```diff
@@ -71,6 +71,69 @@ func TestNewVaultWriteQuery(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		_, _, err = wq.Fetch(clients, &QueryOptions{})
+		if err != nil {
+			fmt.Println(err)
+		}
+
+		rq, err := NewVaultReadQuery(path)
```
freddygv: should this fail on err?

dependency/vault_write_test.go
```diff
@@ -71,6 +71,69 @@ func TestNewVaultWriteQuery(t *testing.T) {
...
```
freddygv: same as above

dependency/vault_write.go
```diff
@@ -168,12 +170,20 @@ func (d *VaultWriteQuery) writeSecret(clients *ClientSet, opts *QueryOptions) (*
...
+	if isv2 {
+		data = map[string]interface{}{"data": d.data}
+	}
+
+	vaultSecret, err := clients.Vault().Logical().Write(d.path, data)
 	if err != nil {
 		return nil, errors.Wrap(err, d.String())
 	}
+	// always nil when KVv1 engine, when isv2==false
 	if vaultSecret == nil {
```
freddygv: why not combine these with `if vaultSecret == nil && isv2`?

dependency/vault_write_test.go
```diff
@@ -71,6 +71,69 @@ func TestNewVaultWriteQuery(t *testing.T) {
 	}
 }
 
+func TestVaultWriteSecretKV_Fetch(t *testing.T) {
+	t.Parallel()
+
+	// trigger a nil-pointer-deref panic in wq.Fetch() with KVv1 due to
```
freddygv: these comments may seem cryptic to people looking at this test after the fix is in

dependency/vault_write_test.go
eikenb: Sure should. Seems like the linter should pick out unhandled err values like that. I overlooked it as I was focused on using it to test the value read, not so much the reader.

dependency/vault_write_test.go
eikenb: Is there some way I could make the clearer? There are times when I'm looking at a test and wondering why it was added... so I thought I'd start putting in small comments to see if that helped.

freddygv: Yeah that would be useful. You could add a reference to the GH issue for context.

Something like:
```
// previously triggered a nil-pointer-deref panic in wq.Fetch() with KVv1 due to
// writeSecret() returning nil for vaultSecret
// see #1255
```

dependency/vault_write.go
eikenb: Because I missed that refactor. Thanks.

dependency/vault_write_test.go
eikenb: Thanks. I'll use GH-1252 to make it clear it is a github issue. 

I'll make a similar change (past tense + issue reference) to the other test's comment as well.

Thanks!

PR #1260, opened 2019-08-16
Issue1198: error reading deleted secret

eikenb:


dependency/dependency_test.go
```diff
@@ -149,6 +149,16 @@ func (v *vaultServer) CreateSecret(path string, data map[string]interface{},
 	return err
 }
 
+// deleteSecret lets us delete keys as needed for tests
+func (v *vaultServer) deleteSecret(path string) error {
+	path = v.secretsPath + "/" + path
```
adams-sarah: maybe better not to modify the input? maybe instantiate a new variable? or else maybe all on one line like:
```
/* blah */.Delete(v.secretsPath + "/" + path)
```

dependency/vault_read_test.go
```diff
@@ -325,6 +326,34 @@ func TestVaultReadQuery_Fetch_KVv2(t *testing.T) {
...
+		d, err := NewVaultReadQuery(path)
+		if err != nil {
+			t.Fatal(err)
+		}
+		_, _, err = d.Fetch(clients, nil)
+		if err == nil {
+			t.Fatal("Nil received when error expected")
+		}
+		exp_err := fmt.Errorf("no secret exists at %s", path)
+		assert.Equal(t, exp_err, errors.Cause(err))
```
adams-sarah: regardless of whether the team prefers using `assert` or `t.Fatal` for tests, we should at least keep it consistent within a test. I'd like to see here instead, `if exp_err != errors.Cause(err); t.Fatal` or else a change in the rest of the test assertions. wdyt?

eikenb: Yeah... I was going back and forth about using assert here and ended up in between. Assert is used throughout the tests and so I thought staying consistent would be good, but being more idiomatic would also be good.

I'm going to try the more idiomatic approach and see how it works going forward.

dependency/dependency_test.go
eikenb: No variable at all is better IMO.  Thanks!

dependency/vault_read_test.go
eikenb: Ended up with a string comparison for the error compare. Hope the 1.13 errors changes help with this.

PR #1261, opened 2019-08-19
add a table of contents to the README

eikenb:
TOC is generated and updated via the script.
Also added a Makefile target (toc) to run script.


eikenb:
Please note (to anyone who finds the big TOC annoying and finds this PR) that the TOC is a temporary solution to allow for easier doc navigation until we have a chance to restructure the documentation (which is planned).

PR #1265, opened 2019-08-22
change backup to use os.Link

eikenb:
Instead of opening a new file and manually copying over the contents,
just use a hard link. This has different tradeoffs vs manually copying
(eg. it isn't supported on plan9), but should generally be more robust
and has no race.

This also changes it to just log an error vs returning it, in the case
of a os.Link returning an error, as that causes consul-template to exit
(which we don't want for this).


liuxu623:
There is a bug, if xxx.bak is exist, os.Link will return `file exists` error.

PR #1277, opened 2019-09-05
fix sleep on non-renewable vault secrets

### LINKED ISSUE #1272: Renewal check for vault-pki certificates fails, renewing constantly
### Consul Template version

consul-template v0.21.2 (3957cdfc)

### Configuration

```hcl
log_level = "trace"
vault {
  address = "http://192.168.0.254:8200"
   ssl {
     enabled = false
   }
 }
template {
  source = "/etc/consul-template/etcd_ca.tmpl"
  destination = "/tmp/ca.pem"
  # Required to not have matchbox interpolate
  left_delimiter  = "[["
  right_delimiter = "]]"
}
```

```liquid
[[- $hostname := env "HOSTNAME" -]]
[[- $hostip := env "HOSTIP" -]]
[[- $cn_arg := printf "common_name=%s.%s" $hostname "po.k8s.zone" -]]
[[- $ip_arg := printf "ip_sans=127.0.0.1,%s" $hostip -]]
[[- with secret "/ui/poinfra/unicorn-prod-01-bs-etcd/issue/etcd" $cn_arg $ip_arg -]]
[[ .Data.issuing_ca ]][[ end ]]```

### Command

```shell
consul-template -config /etc/consul-template/test.hcl
```

### Debug output

https://gist.github.com/sfudeus/397bfd48758355c3b952ecebfdf3a2c6

### Expected behavior

Our cert ttl is 8h. With consul-template 0.21.0, renewal occurred with around 5h certificate lifetime left. With 0.21.2 certificates are renewed continuously (i.e. approx. all 17s in this test setup. 

Certs should have been fetched and renewed when 1/6 to 1/3 of their TTL is left (see #1267)

### Actual behavior

Certificates are updated continuously.

### Steps to reproduce

1. Start consul-template against vault with pki in given trivial configuration. 
2. Observe behaviour in log output 

### References

Are there any other GitHub issues (open or closed) that should
be linked here? For example:
- #1267

eikenb:
An over-zealous refactoring resulted in <-time.After() channel used for
non-renewable sleep to be ignored. I originally had channel of the time
channels but "simplified" it too much. This reworks it the way the sleep
is passed to the next iteration to just pass the time.Duration of the
sleep down the channel, then time.Sleep-ing when reading off the
channel. This way it can see it has a sleep to do, then do it instead of
skipping it as the time hadn't passed yet.

This also let me write a test for it as I don't need to actually sleep
to see that there is something in the channel.

Fixes #1272

### DIFF
--- a/dependency/vault_read.go
+++ b/dependency/vault_read.go
@@ -19,7 +19,7 @@ var (
 // VaultReadQuery is the dependency to Vault for a secret
 type VaultReadQuery struct {
 	stopCh  chan struct{}
-	sleepCh <-chan time.Time
+	sleepCh chan time.Duration
 
 	rawPath     string
 	queryValues url.Values
@@ -46,7 +46,7 @@ func NewVaultReadQuery(s string) (*VaultReadQuery, error) {
 
 	return &VaultReadQuery{
 		stopCh:      make(chan struct{}, 1),
-		sleepCh:     make(chan time.Time, 1),
+		sleepCh:     make(chan time.Duration, 1),
 		rawPath:     secretURL.Path,
 		queryValues: secretURL.Query(),
 	}, nil
@@ -61,7 +61,8 @@ func (d *VaultReadQuery) Fetch(clients *ClientSet, opts *QueryOptions,
 	default:
 	}
 	select {
-	case <-d.sleepCh:
+	case dur := <-d.sleepCh:
+		time.Sleep(dur)
 	default:
 	}
 
@@ -82,7 +83,7 @@ func (d *VaultReadQuery) Fetch(clients *ClientSet, opts *QueryOptions,
 	if !vaultSecretRenewable(d.secret) {
 		dur := leaseCheckWait(d.secret)
 		log.Printf("[TRACE] %s: non-renewable secret, set sleep for %s", d, dur)
-		d.sleepCh = time.After(dur)
+		d.sleepCh <- dur
 	}
 
 	return respWithMetadata(d.secret)
--- a/dependency/vault_read_test.go
+++ b/dependency/vault_read_test.go
@@ -227,6 +227,39 @@ func TestVaultReadQuery_Fetch_KVv1(t *testing.T) {
 		case <-dataCh:
 		}
 	})
+
+	t.Run("nonrenewable-sleeper", func(t *testing.T) {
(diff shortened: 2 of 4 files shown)


dependency/vault_read.go
```diff
@@ -82,7 +83,7 @@ func (d *VaultReadQuery) Fetch(clients *ClientSet, opts *QueryOptions,
 	if !vaultSecretRenewable(d.secret) {
 		dur := leaseCheckWait(d.secret)
 		log.Printf("[TRACE] %s: non-renewable secret, set sleep for %s", d, dur)
-		d.sleepCh = time.After(dur)
+		d.sleepCh <- time.After(dur)
```
adams-sarah: hm 🤔 this chan(chan) is confusing me. I'm wondering if this can be done in a simpler way?
I'm sure I'm missing something -- but why can't this line just be `time.Sleep(dur)` and get rid of all the sleep channel stuff? `time.After` will block in the same way a `time.Sleep` would.. I guess I don't understand the use of the channel here. Could you explain your thinking to me?

eikenb: The main point is that the sleep needs to be saved until next time the function is called. The function is called repeatedly by the polling code and is responsible for its own retry timing (sleeping). The first time it is run it needs to return immediately, then after that it needs to have a delay. The sleep/delay is checked for and done at the top of that function (~ line 64).

It could be simplified into `chan time.Duration` though... the channel of channels is not necessary. That would simplify it a bit. It would change where it is used like...

From..
```
	select {
	case sleep := <-d.sleepCh:
		<-sleep
	default:
	}
```
To.. (assuming d.sleepCh was `chan time.Duration`)..
``` 
	select {
	case dur := <-d.sleepCh:
		time.Sleep(dur)
	default:
	}
```

adams-sarah: I see! Thanks for explaining. Oh man I like that second way (w/ `time.Sleep(dur)`) so much better, it's a lot clearer to me. Wdyt?

eikenb: I agree, it is cleaner. I think I was just stuck in the time.After rut and didn't really consider alternatives.

eikenb: Just force-pushed that change (along with a re-wording of the commit message).

PR #1279, opened 2019-09-09
add basic rate limiting to consul api calls

eikenb:
Rate limit all API calls to Consul to ~1/100ms to keep consul-template
from flooding consul-agent with requests. This takes the idea from #1066
and simplifies it by making it work this way by default instead of
making it a configurable option.

Closes #1066


eikenb:
If any more complex rate limiting is needed, we should probably switch to https://godoc.org/golang.org/x/time/rate, but this is fine for now.

PR #1290, opened 2019-09-27
Allow overriding the default delimiter

mdgreenfield:
For: #1008

This replaces https://github.com/hashicorp/consul-template/pull/1081 with a signed CLA.

This implements a defaults section that allows you to overwrite the default delimiters used for all templates.


mdgreenfield:
@pearkes anything else needed here?

eikenb:
Hey @mdgreenfield, thanks for the PR.

We'll review it and let you know as soon as we can. It shouldn't be to long.

config/config.go
```diff
@@ -48,6 +48,9 @@ type Config struct {
 	// Dedup is used to configure the dedup settings
 	Dedup *DedupConfig `mapstructure:"deduplicate"`
 
+	// Default is used to configure the defaults for templates
+	Defaults *DefaultsConfig `mapstructure:"defaults"`
```
eikenb: Could you rename `Defaults` to something more descriptive... maybe `DefaultDelims` or similar.
Same for the `DefaultsConfig` struct. Please keep the fact that they match.
Thanks.

eikenb:
Sorry for the silence on this. Other priorities came up but I'm now making a pass reviewing things. Got to this PR first as it came up for discussion internally.

While in general I'm trying to minimize new configuration settings as CT already has a bunch, I'm leaning towards merging this as I think readability is very important. Also making this global makes more sense to me and keeps it simpler than a per-template setting. Anyways, I left a comment with the only change I'd like to see after an initial pass.

Thanks.

mdgreenfield:
Thanks @eikenb. I've rebased with master and updated the names of the vars and the config used in the config file.

PR #1294, opened 2019-10-15
Fix incorrect code example in README.md

rpagliuca:
Replace `toJson` with `toJSON` in one of the README.md code examples, as the function name is case sensitive and throws errors otherwise.


eikenb:
Thanks @rpagliuca. Good catch.

PR #1304, opened 2019-11-07
Issue 1262 consul connect

eikenb:
Initial Consul Connect support. Target is to allow proxies (nginx, haproxy) to route directly to services inside the mesh network. This is the basic support and reference docs. I'm still working on some examples/intros for each.

### DIFF
--- a/dependency/connect_leaf.go
+++ b/dependency/connect_leaf.go
@@ -1,0 +1,77 @@
+package dependency
+
+import (
+	"fmt"
+	"log"
+	"net/url"
+
+	"github.com/pkg/errors"
+)
+
+var (
+	// Ensure implements
+	_ Dependency = (*ConnectLeafQuery)(nil)
+)
+
+type ConnectLeafQuery struct {
+	stopCh chan struct{}
+
+	service string
+}
+
+func NewConnectLeafQuery(service string) *ConnectLeafQuery {
+	return &ConnectLeafQuery{
+		stopCh:  make(chan struct{}, 1),
+		service: service,
+	}
+}
+
+func (d *ConnectLeafQuery) Fetch(clients *ClientSet, opts *QueryOptions) (
+	interface{}, *ResponseMetadata, error,
+) {
+	select {
+	case <-d.stopCh:
+		return nil, nil, ErrStopped
+	default:
+	}
+	opts = opts.Merge(nil)
+	log.Printf("[TRACE] %s: GET %s", d, &url.URL{
+		Path:     "/v1/agent/connect/ca/leaf/" + d.service,
+		RawQuery: opts.String(),
+	})
+
+	cert, md, err := clients.Consul().Agent().ConnectCALeaf(d.service,
+		opts.ToConsulOpts())
+	if err != nil {
+		return nil, nil, errors.Wrap(err, d.String())
+	}
+
+	log.Printf("[TRACE] %s: returned response", d)
+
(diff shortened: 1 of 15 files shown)


dependency/connect_leaf.go
```diff
@@ -0,0 +1,77 @@
...
+func (d *ConnectLeafQuery) Stop() {
+	close(d.stopCh)
+}
+
+func (d *ConnectLeafQuery) CanShare() bool {
+	return false
+}
+
+func (d *ConnectLeafQuery) Type() Type {
+	return TypeConsul
```
rboyer: Do these objects need unique Types or is it fine for them to all share `TypeConsul`?

template/funcs.go
```diff
@@ -472,6 +473,62 @@ func servicesFunc(b *Brain, used, missing *dep.Set) func(...string) ([]*dep.Cata
 	}
 }
 
+// serviceFunc returns or accumulates health service dependencies.
```
rboyer: The comment doesn't match the name of the function.

template/template_test.go
```diff
@@ -1606,8 +1607,87 @@ func TestTemplate_Execute(t *testing.T) {
...
+		},
+		{
+			"func_connect",
+			&NewTemplateInput{
+				Contents: `{{ range connect "webapp" }}{{ .Address }}{{ end }}`,
+			},
+			&ExecuteInput{
+				Brain: func() *Brain {
+					b := NewBrain()
+					d, err := dep.NewHealthServiceQuery("webapp")
```
rboyer:
```suggestion
					d, err := dep.NewHealthConnectQuery("webapp")
```
I suspect your setup will need to be adjusted after you actually use the new code.

template/template_test.go
```diff
@@ -1606,8 +1607,87 @@ func TestTemplate_Execute(t *testing.T) {
...
+					})
+					return b
+				}(),
+			},
+			"1.2.3.45.6.7.8",
+			false,
+		},
 	}
 
+	//	struct {
```
rboyer: stray notes?

eikenb: Sort of... notes that I thought I'd keep. I'm always referring to the struct type at the very top when adding tests at the bottom. Annoying to flip back and forth. Thought I'd just add a commented copy of the struct where I could see it.

template/template_test.go
eikenb: :man_facepalming:

eikenb: And actually no other changes are needed, aside from calling the correct method. Those tests are totally mock based using the data injected into the *Brain (thinking Neo plugging in here).. and that data was correct. So as long as the NewHealthConnectQuery itself doesn't return an error it behaves the same.

template/funcs.go
eikenb: All these methods are so similar that the cut-n-paste urge is high... but that leads to crap like this. :disappointed:

dependency/connect_leaf.go
eikenb: At the moment there are only 2 types. TypeConsul and TypeVault that indicate the backend. So they should both be TypeConsul.

template/funcs.go
eikenb: Fixed.

template/template_test.go
eikenb: Fixed

Issue #1314, opened 2019-11-16
'sockaddr' function returning whitespace after address


errnoh:
### Consul Template version
consul-template v0.23.0 (5cf9d687)

### Configuration

```liquid
{{ sockaddr "GetPrivateIP" }}:80
```

### Command

```shell
consul-template -dry -once -template test.tmpl
```

### Expected behavior

What should have happened?
`192.168.1.1:80`

### Actual behavior

What actually happened?
`192.168.1.1 :80`

### Steps to reproduce

1. Dry run the minimal example template

(Temporary workaround is to pipe it to `trimSpace`)

PR #1315, opened 2019-11-17
Remove extra whitespace from sockaddr func

vtorhonen:
Currently `sockaddr` function explicitly leaves extra whitespace for every template it renders. This PR removes that whitespace. It also improves the related tests so that the (first) loopback address of the machine is string-matched to "127.0.0.1". Previously the test only checked whether "127.0.0.1" is in the result set, without checking the rendered output for an exact match.

Fixes issue #1314.


eikenb:
Thanks for the fix @vtorhonen. I'll be doing a bugfix release soon and will be sure to get this in it.

PR #1318, opened 2019-11-18
Fix documentation for default exec kill signal.

fubhy:


config/exec.go
```diff
@@ -41,7 +41,7 @@ type ExecConfig struct {
 	Env *EnvConfig `mapstructure:"env"`
 
 	// KillSignal is the signal to send to the command to kill it gracefully. The
-	// default value is "SIGTERM".
+	// default value is "SIGINT".
```
sodabrew: Given that this comment was wrong and yet only 30 lines of code below the definition above (line 13), perhaps nuke the comment outright?

fubhy: Agreed.

eikenb:
Thanks @fubhy for the PR and @sodabrew for the help.

PR #1320, opened 2019-11-19
consul namespaces support

eikenb:
Add support for Consul namespaces following the same pattern as for
Vault namespaces.

Depends on code in consul's master branch (unreleased). So won't build unless you update the modules dependencies to the correct master commit. Once the upstream code is released, the module dependencies will be updated and added to this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,6 +264,11 @@ consul {
   # clients can connect.
   address = "127.0.0.1:8500"
 
+  # This is a Consul Enterprise namespace to use for reading/writing. This can
+  # also be set via the CONSUL_NAMESPACE environment variable.
+  # BETA: this is to be considered a beta feature as it has had limited testing
+  namespace = "foo"
+
   # This is the ACL token to use when connecting to Consul. If you did not
   # enable ACLs on your Consul cluster, you do not need to set this option.
   #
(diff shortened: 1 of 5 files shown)


lornasong:
This makes sense to me! I can see the new Consul namespaces' parallel with Vault's. Can approve after module dependencies are updated or happy to approve now - whatever the team norm is.

eikenb:
Thanks @lornasong! I think it could be approved now and if anything but the dependencies change I could ping for another look.

freddygv:
Hi @eikenb, one thing to consider for long-run namespace support is that users may want to query more than one namespace across different template functions.

This ie because:
- Some endpoints allow using the wildcard namespace `'*'` (list methods), while others do not (set, get delete methods).
- Some resources like nodes are only in the default namespace while resources like services/checks/KV can be in user-created namespaces.

If someone wants to use multiple namespaces, I think we would need to enable something like we do for querying nodes:
https://github.com/hashicorp/consul-template#nodes

This way someone can pick a default namespace for their queries in the config, but also overwrite the query options for a particular request. 

For example: If someone wants to query services within the `billing` namespace but also nodes in the default namespace. They could set `billing` as the namespace in the config, and then override the namespace in the `nodes` template function.

eikenb:
Hey @freddygv, thanks for the input.

I thought of that and other possible enhancements but decided there was no good way to know which would be useful at this point. So I thought it better to keep it as simple as possible at this point and see what kind of feature requests come in.

eikenb:
Rebased on newer consul-api so it now passes CI.

PR #1322, opened 2019-11-20
Updating examples for PKI relating to the generate_lease requirement.

arodd:
Updating the PKI example now that the generate_lease setting on the PKI role is no longer a strict requirement.


eikenb:
Hey @arodd, documentation updates are very welcome. Thanks.

PR #1324, opened 2019-12-10
alpine docker image doesn't run entrypoint as root

### LINKED ISSUE #1321: consul-template container runs as root - breaks CISO compliance
Version:  hashicorp/consul-template:0.23.0-alpine

Whilst the consul-template process within the container runs as a non-root user the container itself runs as root.

For us (a large financial services company) running containers with the root user is explicitly dis-allowed by our CISO as a mandatory control and enforced both by image and run-time assurance policies (i.e launch of the container is blocked) Running containers as non-root is also widely recognised as a recommended security best practice.

```
/ # id
uid=0(root) gid=0(root) groups=0(root),1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),20(dialout),26(tape),27(video)
/ # 
/ # whoami
root
/ # 
/ # ps
PID   USER     TIME  COMMAND
    1 root      0:00 {docker-entrypoi} /bin/dumb-init /bin/sh /bin/docker-entrypoint.sh -config=/etc/consul-template/consul-template-config.hcl -log-level=debug
    6 consul-t  0:00 /bin/consul-template -config=/consul-template/config -config=/etc/consul-template/consul-template-config.hcl -log-level=debug
   27 root      0:00 /bin/sh
   37 root      0:00 ps
/ #
```

We run consul-template containers within Kubernetes which also applies a restricted Pod Security Policy which specifically prevents containers running as root (this is also common practice)

```
apiVersion: extensions/v1beta1
kind: PodSecurityPolicy
metadata:
  name: restricted
  ...
spec:
...
  runAsUser:
    rule: MustRunAsNonRoot
```
We have asked Hashicorp support to remediate this issue, however, they have stated that consul-template is a community project and changes are only supported by issues raised in this repo.

eikenb:
Set USER directrive in the Dockerfile to the consul-template user
created to run consul-template.

Made this change as some organizations security policies require it. The
only reason it wasn't done this way before was to run a chown on the
(possibly) bind mounted directories and this is a bad practice anyways
(who knows who the user matches on the external system).

Only change the alpine image this way as the other 2 (scratch and light)
purposely do as little as possible so they don't even create the user.

Fixes #1321

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## v0.24.0 (Dec NN, 2019)
+
+BREAKING CHANGES:
+
+* Alpine Docker image no longer runs as root and so doesn't change ownership of the /consul-template/data and /consul-template/config directories to the consul-template user. See the [Docker Image Use](https://github.com/hashicorp/consul-template#docker-image-use) topic in the documentation for more.
+
+
 ## v0.23.0 (Nov 13, 2019)
 
 IMPROVEMENTS:
--- a/docker/alpine/Dockerfile
+++ b/docker/alpine/Dockerfile
@@ -22,17 +22,22 @@ RUN \
 FROM alpine:latest
 LABEL maintainer "John Eikenberry <jae@zhar.net>"
 
-# This is the release of https://github.com/hashicorp/docker-base to pull in order
-# to provide HashiCorp-built versions of basic utilities like dumb-init and gosu.
+# This is the release of https://github.com/hashicorp/docker-base to pull in
+# order to provide HashiCorp-built versions of basic utilities like dumb-init.
 ARG DOCKER_BASE_VERSION="0.0.4"
 
 # This is the location of the releases.
 ARG HASHICORP_RELEASES="https://releases.hashicorp.com"
 
+# UID and GID of consul-template user and group.
+# These are the defaults, this makes them explicit and overridable.
+ARG UID=100
+ARG GID=1000
+
 # Create a consul-template user and group first so the IDs get set the same way,
 # even as the rest of this may change over time.
-RUN addgroup consul-template && \
-    adduser -S -G consul-template consul-template
+RUN addgroup -g ${GID} consul-template && \
+    adduser -u ${UID} -S -G consul-template consul-template
 
 # Set up certificates, our base tools, and Consul Template (CT).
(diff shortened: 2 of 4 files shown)


eikenb:
I've realized this needs a few tweaks. So don't review just yet.

eikenb:
I made the UID and GID into build time customizable arguments and added documentation to the readme about how you need to make sure about the permissions.

That covers the tweaks I was thinking of and this is now ready for review.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## v0.24.0 (Dec NN, 2019)
+
+BREAKING CHANGES:
+
+* Alpine Docker image no longer runs as root and so doesn't change ownership of the /consul-template/data and /consul-template/config directories to the consul-template user. See the [Docker Image Use](https://github.com/hashicorp/consul-template#docker-image-use) topic in the documentation for more.
```
lornasong: I visited the link for `Docker Image Use` and was brought to the top of the consul-template repo. I didn't see a docker image use topic. Not sure if I did something wrong and just wanted to flag.

docker/alpine/Dockerfile
```diff
@@ -22,17 +22,23 @@ RUN \
 FROM alpine:latest
 LABEL maintainer "John Eikenberry <jae@zhar.net>"
 
-# This is the release of https://github.com/hashicorp/docker-base to pull in order
-# to provide HashiCorp-built versions of basic utilities like dumb-init and gosu.
+# This is the release of https://github.com/hashicorp/docker-base to pull in
+# order to provide HashiCorp-built versions of basic utilities like dumb-init
+# and gosu.
```
lornasong: From the changes below, it seems like `gosu` has been removed. If i understood correctly, then maybe worth removing `and gosu` from comment?

CHANGELOG.md
eikenb: The link goes to a section that was added to the readme in this PR. It should work after the merge based on normal github markdown rules.

docker/alpine/Dockerfile
eikenb: Good catch. I'll remove that. Thanks.

CHANGELOG.md
lornasong: ah 🤦‍♀, thank you

PR #1326, opened 2019-12-12
use gnueabihf for 32-bit arm build

### LINKED ISSUE #1317: arm builds are linked against wrong library
ARM builds for `consul-template` are linked against non-arm library, resulting in them not being executable.

Comparison of dynamically linked `nomad` build and dynamically linked `consul-template` build:

```
> uname -a
Linux scw-sleepy-ishizaka 4.9.93-mainline-rev1 #1 SMP Tue Apr 10 09:42:40 UTC 2018 armv7l armv7l armv7l GNU/Linux
> wget https://releases.hashicorp.com/consul-template/0.23.0/consul-template_0.23.0_linux_arm.zip; unzip consul-template_0.23.0_linux_arm.zip
> ./consul-template
-bash: ./consul-template: No such file or directory
> file ./consul-template
./consul-template: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux.so.3, stripped

> nomad version
Nomad v0.10.1 (0d4e5d949fe073c47a947ea36bfef31a3c49224f)
> file /usr/local/bin/nomad
/usr/local/bin/nomad: ELF 32-bit LSB executable, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-armhf.so.3, for GNU/Linux 3.2.0, BuildID[sha1]=e61e9b348b4607ffe8813aba5593a91ff8224ebf, not stripped

> ls /lib/*.so.3
/lib/ld-linux-armhf.so.3
```

There's a workaround for this: copy/symlink the armhf library to the location consul-template expects it to be at, but that's not solving the real issue.

NOTE: Probably only affects `arm` and not `arm64`: 
arm64: `consul-template: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, stripped`

eikenb:
The 32 bit arm build that was supposed to work on both arm5 and arm6
won't run on systems without moving files around due to linking/naming
clashes. So we're changing the arm6 build to use
arm-linux-gnueabihf-gcc, which links against the expected libraries.

At this point we're just going to drop support for arm5 as I don't think
anyone is using it and it's starting to become obsolete.

Fixes #1317

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,8 +55,8 @@ define make-xc-target
 		@printf "%s%20s %s\n" "-->" "${1}/${2}:" "${PROJECT}"
 		case "$2" in \
 			arm) export CGO_ENABLED="1" ; \
-				  export GOARM=5 \
-				  export CC="arm-linux-gnueabi-gcc" ;; \
+				  export GOARM=6 \
+				  export CC="arm-linux-gnueabihf-gcc" ;; \
 			arm64) export CGO_ENABLED="1" ; \
 				  export CC="aarch64-linux-gnu-gcc" ;; \
 			*) export CGO_ENABLED="0" ;; \


eikenb:
Arm5 was released in 2009 and is only used in very small embedded stuff these days, and not much of that. There are no ARM servers using it and even the small lab setups use newer CPUs. For example the oldest Raspberry-Pi, popular in small/home labs, used Arm-6.

eikenb:
@lornasong Or did you want more explanation in the commit message?

lornasong:
@eikenb - no need for more explanation in the commit message. Thanks for the information about Arm5! I just wanted to understand it a little bit better and had a hard time finding more information when searching. Helpful, thank you!

PR #1327, opened 2019-12-16
Added sha256Hex function

vfoucault:
this will compute the sha256 hash function and output the hex version
of it.

tests and doc updated.

how to use:

```liquide
{{ "bladibla" | sha256Hex }}
```

this will output 

```
54cf4c66bcabb5c20e25331c01dd600b73369e97a947861bd8d3a0e0b8b3d70b
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,7 @@ this functionality might prove useful.
       - [regexMatch](#regexmatch)
       - [regexReplaceAll](#regexreplaceall)
       - [replaceAll](#replaceall)
+      - [sha256Hex](#sha256hex)
       - [split](#split)
       - [timestamp](#timestamp)
       - [toJSON](#tojson)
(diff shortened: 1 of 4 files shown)


eikenb:
Hey @vfoucault, thanks for the PR.

I'll most likely accept this but I'm curious... what is the use case?

vfoucault:
Hey @eikenb, the main goal is to obviously generate  sha256 of passwords. I use it in the analytics data named Clickhouse.

```xml
<?xml version="1.0"?>
<yandex>
    <users>
        <default>
            <password_sha256_hex>54cf4c66bcabb5c20e25331c01dd600b73369e97a947861bd8d3a0e0b8b3d70b</password_sha256_hex>
            <networks incl="networks" replace="replace">
                <ip>::/0</ip>
            </networks>
            <profile>default</profile>
            <quota>default</quota>
        </default>
    </users>
</yandex>
```
The password could be also stored in plaintext, but this is less sexy and secure !

Issue #1096, opened 2018-04-23
Make timeout for blocking query configurable


guidoiaquinti:
To do that we should probably simply expose the `defaultWaitTime` to the CLI/conf package:
https://github.com/hashicorp/consul-template/blob/d0ee7a7a959b1106932665677d8259974907741e/watch/view.go#L13-L16

I might be able to work on this _soon_ but I'm opening the issue anyway to keep track of the progress.

guitmz:
@guidoiaquinti hey do you still plan to work on this? I was thinking about submitting a patch myself if I find the time

guidoiaquinti:
Unfortunately not @guitmz, happy if you can PR

eikenb:
@guidoiaquinti @guitmz 

Would either of you mind giving a bit more context about why you'd like this feature? A use case you have in mind or like? Thanks.

guitmz:
@eikenb sure, a real example that happened to me (which made me fork consul-template and implement this feature) was the following:

I have consul running behind AWS application load balancer which had a keepalive of 120 seconds and everything was just fine.

For reasons, this keepalive had to be changed to 45 seconds and then problems started to appear. See, this means that the load balancer is now closing connections after 45 seconds and consul-template wants them open for 60 seconds. This was causing 5xx error all over the place.

Solution was to fork and make this value less than 60 seconds in consul-template.

PR #1329, opened 2019-12-20
Make timeout for blocking query configurable

guitmz:
This PR aims to introduce a new configuration value, allowing custom block query timeout. I would appreciate feedback on this as I'm not sure if I have covered all the necessary bases.

Closes https://github.com/hashicorp/consul-template/issues/1096

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,6 +346,12 @@ kill_signal = "SIGINT"
 # less cluster load, but are more likely to have outdated data.
 max_stale = "10m"
 
+
+# This is amount of time in seconds to do a blocking query for.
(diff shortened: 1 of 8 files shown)


guitmz:
I'm not really sure why the tests are failing in the CI, as they work fine locally.

eikenb:
Hey @guitmz, thanks for the contribution.

That test has a race condition where it fails sometimes. I've been working to fix those but some are harder to reproduce (and fix) than others. I plan on a major reworking of the tests soon-ish which should hopefully eliminate the rest.

I'm working to get a bugfix release out next, but after that I'll take a look.

PR #1330, opened 2020-01-08
changelog notes for 0.24.0

eikenb:
I'll update the date when doing the actual release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,16 @@
-## v0.24.0 (Dec NN, 2019)
+## v0.24.0 (Jan XX, 2020)
 
 BREAKING CHANGES:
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,9 +1,16 @@
-## v0.24.0 (Dec NN, 2019)
+## v0.23.1 (Jan XX, 2020)
```
lornasong: q about the change `v0.24.0` -> `v.23.1`: i was thinking the 0.24.0 was originally because of the breaking changes - are you thinking that the breaking change no longer warrants incrementing the minor version?

eikenb: Good thought. I had conflicting versions (changelog vs milestone) and went with the micro-version (23.1) as that is more normal for non-feature releases. But the breaking change does seem like it might warrant a larger bump just to help draw attention to it.

PR #1332, opened 2020-01-10
Make user non-numeric to satisfy PSP

funkypenguin:
Hey guys,

I appreciate that the container no longer runs as root, but instead runs as the `consul-template` user.

This PR introduces a small change to the Alpine Dockerfile, in that it uses numeric UID/GID values for the `USER` statement. This is required when using PodSecurityPolicies under Kubernetes, to assure the admission controller that the container will not run as root. (Setting a non-numeric value like `USER consul-template:consul-template` is insufficient, since the Dockerfile could simply have renamed the root account, I guess!)

Cheers!
D


eikenb:
Thanks @funkypenguin, 

I'll try to get this merged and re-build the docker images next week. It doesn't really change anything, almost being just a cosmetic tweak, so I don't think I'll worry about doing a release just to update the docker image.

eikenb:
@funkypenguin I merged this and pushed up a new docker image built with it in place.

Issue #1245, opened 2019-08-02
Update alpine Docker image


eikenb:
Eliminate use of old internal docker-base "package".
Remove any unused alpine packages (possible candidates; curl, libcap, openssl).
Any other cleanup/improvements.

eikenb:
#1321 removed the use of gosu

eikenb:
Consul's Dockerfile for reference.

https://github.com/hashicorp/docker-consul/blob/master/0.X/Dockerfile

PR #1333, opened 2020-01-19
docker: alpine without docker-base

greut:
Alpine comes with tini, which is quite similar to dumb-init. It simplifies a bit the Dockerfile for alpine.

cf. #1245


eikenb:
I was just working on this as well. Alpine also comes with dumb-init and that is what I was going to use. Do you think tini offers any advantages? If not I'd suggest we stick with dumb-init to keep the change to a minimum.

As you beat me to the PR we can use your version once we nail down the tini/dumb-init thing.

greut:
@eikenb sure `dumb-init` it is.

PR #1341, opened 2020-02-21
dependency: fix kv2 data path prefix checking

### LINKED ISSUE #1340: fail when kv2 secret path starts with "data"
### Consul Template version

consul-template v0.22.0 (005b42eb)
vault v1.3.2

### Configuration

```liquid
# Copy-paste your Consul Template template here
{{- with secret "test/dataface/test" }}
name={{ .Data.data.name }}
{{ end }}
```

### Expected behavior

As stated in [docs](https://github.com/hashicorp/consul-template#secret), in vault 0.10.1+ secret path does not have to be prefixed with "data" if it is kv2.

### Actual behavior

When accessing "test/dataface/test" 403 is thrown:

```
2020/02/17 05:33:16.983733 [WARN] (view) vault.read(test/dataface/test): vault.read(test/dataface/test): Error making API request.

URL: GET https://vault-test.tld/v1/test/dataface/test
Code: 403. Errors:

* permission denied (retry attempt 10 after "1m0s")

```
So, secret path have to be prefixed with "data", i.e. only "test/data/dataface/test" works if secret path begins with "data" letters.

When secret path does not begin with "data" letters, it works as expected, i.e both "test/datoface/test" and "test/data/datoface/test" are good.

### Steps to reproduce

1. Create kv2 vault engine named "test"
2. Create secrets "test/dataface/test" and "test/datoface/test" with some "name" in it. Give access to both for consul token.
3. Access them with "with secret". First does not work, second does.

findkim:
The implicit `/data/` path prefix for kv2 secrets were not properly prepended for secret paths matching `/data*`.

Fixes #1340

### DIFF
--- a/dependency/vault_common.go
+++ b/dependency/vault_common.go
@@ -339,8 +339,12 @@ func addPrefixToVKVPath(p, mountPath, apiPrefix string) string {
 		return path.Join(mountPath, apiPrefix)
 	default:
 		p = strings.TrimPrefix(p, mountPath)
-		// Don't add /data to the path if it's been added manually.
-		if strings.HasPrefix(p, apiPrefix) {
+		// Don't add /data/ to the path if it's been added manually.
+		apiPathPrefix := apiPrefix
+		if !strings.HasSuffix(apiPrefix, "/") {
+			apiPathPrefix += "/"
+		}
+		if strings.HasPrefix(p, apiPathPrefix) {
 			return path.Join(mountPath, p)
 		}
 		return path.Join(mountPath, apiPrefix, p)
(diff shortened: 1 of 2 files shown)


dependency/vault_common.go
```diff
@@ -333,14 +333,15 @@ func isKVv2(client *api.Client, path string) (string, bool, error) {
 	return mountPath, false, nil
 }
 
-func addPrefixToVKVPath(p, mountPath, apiPrefix string) string {
+func addPrefixToVKVPath(p, mountPath, apiPrefixRaw string) string {
+	apiPrefix := path.Clean(apiPrefixRaw)
```
eikenb: Why Clean() here? The KV prefixes have a path like structure, they don't follow all the rules. Seems like it might have some unintended side effect.

findkim: Thanks for the context on the prefix structure and good point, updated changes to make it more explicit on the trailing `/` check

Issue #1297, opened 2019-10-29
Renewing a vault token defaults to true even when no vault tokens are in use... should default to false in this case.


ChipV223:
### Consul Template version 

v0.22.0

### Configuration

```
log_level = "debug"

consul {

	address = "127.0.0.1:8501"
	
	retry {
    # This enabled retries. Retries are enabled by default, so this is
    # redundant.
    enabled = true

    # This specifies the number of attempts to make before giving up. Each
    # attempt adds the exponential backoff sleep time. Setting this to
    # zero will implement an unlimited number of retries.
    attempts = 10

    # This is the base amount of time to sleep between retry attempts. Each
    # retry sleeps for an exponent of 2 longer than this base. For 5 retries,
    # the sleep times would be: 250ms, 500ms, 1s, 2s, then 4s.
    backoff = "250ms"

    # This is the maximum amount of time to sleep between retry attempts.
    # When max_backoff is set to zero, there is no upper limit to the
    # exponential sleep between retry attempts.
    # If max_backoff is set to 10s and backoff is set to 1s, sleep times
    # would be: 1s, 2s, 4s, 8s, 10s, 10s, ...
    max_backoff = "1m"
	}
	
	# This block configures the SSL options for connecting to the Consul server.
	ssl {
    # This enables SSL. Specifying any option for SSL will also enable it.
    enabled = true

    # This enables SSL peer verification. The default value is "true", which
    # will check the global CA chain to make sure the given certificates are
    # valid. If you are using a self-signed certificate that you have not added
    # to the CA chain, you may want to disable SSL verification. However, please
    # understand this is a potential security vulnerability.
    verify = true

    # This is the path to the certificate to use to authenticate. If just a
    # certificate is provided, it is assumed to contain both the certificate and
    # the key to convert to an X509 certificate. If both the certificate and
    # key are specified, Consul Template will automatically combine them into an
    # X509 certificate for you.
    cert = "C:\\Consul_Service2\\certs\\server\\dc1-server-consul-0.pem"
    key  = "C:\\Consul_Service2\\certs\\server\\dc1-server-consul-0-key.pem"

    # This is the path to the certificate authority to use as a CA. This is
    # useful for self-signed certificates or for organizations using their own
    # internal certificate authority.
    ca_cert = "C:\\Consul_Service2\\certs\\consul-agent-ca.pem"

  }
}

template {
  source = "C:\\Consul_Service2\\template\\all_services.tpl"
[...]

PR #1352, opened 2020-03-03
don't renew vault token when no token is set

eikenb:
Adds check in config to change the vault token renew check to not fire
if there is no vault token to renew. Avoids warning spams in the logs.

### DIFF
--- a/config/vault.go
+++ b/config/vault.go
@@ -193,16 +193,6 @@ func (c *VaultConfig) Finalize() {
 		c.Namespace = stringFromEnv([]string{"VAULT_NAMESPACE"}, "")
 	}
 
-	if c.RenewToken == nil {
-		default_renew := DefaultVaultRenewToken
-		if c.VaultAgentTokenFile != nil {
-			default_renew = false
-		}
-		c.RenewToken = boolFromEnv([]string{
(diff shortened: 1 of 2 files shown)


config/vault.go
```diff
@@ -258,6 +248,19 @@ func (c *VaultConfig) Finalize() {
 		c.Token = stringFromFile([]string{*c.VaultAgentTokenFile}, "")
 	}
 
+	// must be after c.Token setting, as default depends on that.
+	if c.RenewToken == nil {
+		default_renew := DefaultVaultRenewToken
+		if c.VaultAgentTokenFile != nil {
```
findkim: If the `VaultAgentTokenFile` exists but is empty and doesn't contain a token, should renew token be configured?

eikenb: If the file is set, then it currently sets the default to false/not renew. So in the case that VaultAgentTokenFile was set (!nil) but was empty, it would not try to renew. It would end up setting the token to a blank string which might cause some issues, but that is a separate thing.

findkim: interesting, thanks for clarifying here -- it does sound like a separate issue.

eikenb:
Fixes #1297

PR #1355, opened 2020-03-06
simplified the code in a couple places

eikenb:
While looking into an issue I saw these and went ahead and fixed them.
One is taking advantage of map defaults.
The other eliminates struct elements set to their default values.

Almost skipped the PR/review, but decided I should have these sorts of things reviewed as standard practice in case I overlook something.


eikenb:
@findkim 
> Improves readability!

My thoughts exactly. Thanks!

PR #1358, opened 2020-03-19
Add lease calculation for rotating secrets

jasonodonnell:
There is a subset of non-renewable Vault secrets (static-roles) that have their password rotated on an interval, defined by `rotation_period`.  Their password values change, however, since they're not renewable secrets Consul Template defaults to a 5 minute sleep period.

This updates Consul Template to consider those types of secrets and use their rotation value as the sleep duration for faster updates.

### DIFF
--- a/dependency/vault_common.go
+++ b/dependency/vault_common.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"crypto/x509"
+	"encoding/json"
 	"encoding/pem"
 
 	"github.com/hashicorp/vault/api"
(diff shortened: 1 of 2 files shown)


eikenb:
Thanks @jasonodonnell, [rotation_period](https://www.vaultproject.io/api/secret/databases/index.html#create-static-role) sounds perfect for this.

One thing. Would you add a test case to vault_common_test.go, TestVaultRenewDuration() for this case. Should be very simple.

Thanks.

jasonodonnell:
Done @eikenb!

PR #1359, opened 2020-03-23
Fix - tests support consul 1.7 

fredrikhgrelland:
Since consul 1.7 the response for tagged addresses have changed.: https://github.com/hashicorp/consul/issues/6531

This patch update tests to new 1.7 response syntax.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
   build:
     environment:
       GO111MODULE: "on"
-      CONSUL_VERSION: 1.6.4
+      CONSUL_VERSION: 1.7.2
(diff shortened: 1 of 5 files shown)


eikenb:
I'm debating about this one. Obviously we need to fix the test for 1.7+, but I prefer to keep compatibility with older versions whenever possible as I might need to debug things against older versions for some cases and don't want the test suite fighting me.

eikenb:
@fredrikhgrelland Thanks for taking the time and trouble to fix and submit this!

I've decided to fix it differently though, in a way that tests will still work with older versions. This helped me though by telling me exactly what was broken in the best way possible. So thanks so much for that.

PR #1367, opened 2020-04-16
Allow to set application name in syslog

jpmenil:
Currently "consul-template" appear as progname in syslog,
which is quite annoying when you run multiple programs.

Signed-off-by: Jean-Philippe Menil <jpmenil@gmail.com>


eikenb:
Looks good. Thanks!

PR #1370, opened 2020-04-21
move render events to after template command

### LINKED ISSUE #1369: Event vs Command race when used as library
### Consul Template version
v0.24.1

### Configuration
```hcl
template {
  contents = "hi "
  destination = "hi.out"
  command = "sh -c 'sleep 3; cat hi.out;'"
  command_timeout = "0s"
}

template {
  contents = "bye\n"
  destination = "bye.out"
  command = "sh -c 'sleep 5; cat bye.out;'"
  command_timeout = "0s"
}
```

### Command
```shell
consul-template -config config.hcl  -once
```

### Expected behavior
Templates rendered and commands run before consul-template exits.

### Actual behavior
Templates rendered and commands run after consul-template exists.

### Details
So the above shows a version of the problem. If you have 0s timeout on a command, consul-template w/  `-once` will exit before they get run. In cases where consul-template is run as a batch job in a container, the container could well exit before the commands finish (killing them). This is pretty easy to mitigate with the consul-template command by adding a timeout > 0s.

*But* when using CT as a library, the "standard" way to monitor and know when to exit is to watch for messages on the `TemplateRenderedCh()`. The problem is that the message is sent down that channel before the commands are run, so even with a timeout > 0s the message has been sent and the program could exit.

The simplest solution is to move the code that sends the message from its current location, right before running the commands, to just after running the commands.

From here... https://github.com/hashicorp/consul-template/blob/dc274bc6f619f2c22724349478d483f4aff1a7e1/manager/runner.go#L559-L574

To here... https://github.com/hashicorp/consul-template/blob/dc274bc6f619f2c22724349478d483f4aff1a7e1/manager/runner.go#L603

eikenb:
When using CT as a library, the "standard" way to monitor and know when
to exit is to watch for messages on the TemplateRenderedCh(). The
problem is that the message is sent down that channel before the
commands are run, so the program could exit before the command has run.

The problem is more obvious when CT is run in a transient container,
where even if the commands are run the container might exit before they
finish and they get killed.

Moving the render events to after the command avoids these issues (the
container exit issue also requires a command_timeout > 0s).

Fixes #1369

### DIFF
--- a/README.md
+++ b/README.md
@@ -555,7 +555,9 @@ template {
   command = "restart service foo"
 
   # This is the maximum amount of time to wait for the optional command to
-  # return. Default is 30s.
+  # return. If you set the timeout to 0s the command is run in the background
+  # without monitoring it for errors. If also using Once, consul-template can
+  # exit before the command is finished. Default is 30s.
   command_timeout = "60s"
 
   # Exit with an error when accessing a struct or map field/key that does not
--- a/manager/runner.go
+++ b/manager/runner.go
@@ -556,23 +556,6 @@ func (r *Runner) Run() error {
 		}
 	}
 
-	// Check if we need to deliver any rendered signals
-	if wouldRenderAny || renderedAny {
-		// Send the signal that a template got rendered
-		select {
-		case r.renderedCh <- struct{}{}:
-		default:
-		}
-	}
-
-	// Check if we need to deliver any event signals
-	if newRenderEvent {
-		select {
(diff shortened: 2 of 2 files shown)


eikenb:
I'm going to update the PR to tweak to documentation to mention the 0s/fire-n-forget case.

eikenb:
@findkim I updated the PR with the documentation change.

PR #1378, opened 2020-05-05
Add metrics reporting

### LINKED ISSUE #826: statsd/dogstatsd support for tracking consul template events
It would be really handy to have built in support for sending statsd/dogstatsd metrics on notable consul-template events. The most obvious of which being restarts. For statsd that'd probably be something like `consul-template.<identifier>.restart`, for dostatsd `consul-template.restart#template:identifier`. They may not currently be something obvious to use for identifier currently, perhaps the "flag" that enables stats would include that.

This originally occurred to us while trying to debug/track down https://github.com/hashicorp/consul/issues/2430 where it would have been really useful to be able to track which processes were being restarted when at a glance. 

Originally brought up on the [list](https://groups.google.com/forum/#!msg/consul-tool/XTvu0xk3dJs/ad-fWvhIDQAJ) where @slackpad mentioned "I can think of some other situations where watching restarts could be useful, such as with quiescence timers." General/immediate solution suggested there was to use a wrapper around the restart command that emits the stats.

### LINKED ISSUE #1078: Integrate OpenCensus for tags, stats, and tracing
Would you be interested in a PR to integrate [OpenCensus](https://github.com/census-instrumentation/opencensus-go) in `consul-template`?

My goal is to increase visibility around application performance and behavior monitoring data.

I have a first (draft) implementation to natively expose some simple stats:
```
➜  ~ curl -s localhost:9999/metrics | grep build_info
# HELP consul_template_build_info A metric with a constant '1' value labeled by revision and version from which consul-template was built.
# TYPE consul_template_build_info gauge
consul_template_build_info{gitcommit="",version="0.19.5.dev"} 1
```

### LINKED ISSUE #1338: Monitoring using Prometheus
This is related to monitoring Consul Template. The discussion in #570 has converged to sending health data to Consul (issue has been stale since early 2018). I'm proposing to expose metrics using the Prometheus (soon OpenMetrics format)

Here are my thoughts:
- Works well in Kubernetes settings. Prometheus is most common monitoring solution right now. 
- We are using Consul Template as a sidecar. Hence, there is no port conflict (which is one of the concerns by @sethvargo  [in #570](https://github.com/hashicorp/consul-template/issues/570#issuecomment-202093286)). The http server could be disabled by default, to prevent breaking changes.
- Consul Template can be used just for Vault, in which case there is no Consul to send data to.
- Other HashiCorp products adapt Prometheus as well (e.g. see https://github.com/hashicorp/vault/pull/5308)

findkim:
This is the first step towards adding support for Consul Template monitoring. The changes integrate with [OpenTelemetry](https://github.com/open-telemetry/opentelemetry-go) to add stdout, prometheus, and dogstatsd metric sinks. Metric configuration and [metric sample outputs](https://github.com/findkim/consul-template/blob/aed64af2aa353744640d5c827df2969e4f2c1bef/README.md#metric-samples) are added to the README documentation.

Things to consider when configuring telemetry:
* Namespace and global labels are currently not supported in this first version of metrics. Collecting metrics on multiple Consul Template instances may result in inaccurate gauge values.
* Statsd is not yet supported by opentelemetry-go https://github.com/open-telemetry/opentelemetry-go/pull/591; the support for statsd will be updated when the library does. The dogstatsd export option may or may not work for your metrics collector depending on the tags delimiter.

Enhancement request https://github.com/hashicorp/consul-template/issues/1078 to integrate OpenCensus received a number of community votes, and since it has been merged to become OpenTelemetry, we decided to choose the OpenTelemetry library. OpenTelemetry was chosen over [armon/go-metrics](https://github.com/armon/go-metrics), which has been commonly used in various HashiCorp projects, for Consul Template as an experiment towards a new library that has dedicated support to expand on new sinks and has tracing functionality. For a period of time, there may be a mismatch in supported sinks between Consul and Consul-Template, but we do hope to update this regularly as OpenTelemetry expands.

Resolves #1338, resolves #826, resolves #1078


config/convert.go
```diff
@@ -53,6 +62,15 @@ func FileModeVal(o *os.FileMode) os.FileMode {
 	return *o
 }
 
+// FileModeCopy returns a copy of the os.FireMode
+func FileModeCopy(o *os.FileMode) *os.FileMode {
```
eikenb, line 66: This doesn't seem to be used.

config/convert.go
```diff
@@ -85,6 +103,15 @@ func IntVal(i *int) int {
 	return *i
 }
 
+// IntCopy returns a copy of the int pointer
+func IntCopy(i *int) *int {
```
eikenb, line 107: Also not used.

config/convert.go
```diff
@@ -116,6 +183,15 @@ func SignalVal(s *os.Signal) os.Signal {
 	return *s
 }
 
+// SignalCopy returns a copy of the os.Signal
+func SignalCopy(s *os.Signal) *os.Signal {
```
eikenb, line 187: Not used.

config/convert.go
findkim: Thanks for reviewing the PR. I have removed the unused helper functions.

Issue #1328, opened 2019-12-18
Support log-level as an environment variable


eikenb:
When working on consul-template I find myself switching between log levels quite a bit and having to keep editing scripts or commands. If it supported an environment variable for setting the log-level it would make this much easier.

The cli argument should override the environment variables.

My initial proposed variable name is `CONSULT_TEMPLATE_LOG_LEVEL`, though that is a bit long.

sterchelen:
@eikenb There is already an environment variable for setting the log level.
Look at https://github.com/hashicorp/consul-template/blob/master/config/config.go#L526-L530
You can use `CONSUL_TEMPLATE_LOG` or `CT_LOG`...

sterchelen:
I propose to remove `CT_LOG` and add `_LEVEL` as suffix for `CONSUL_TEMPLATE_LOG`

eikenb:
@sterchelen.. Thanks for pointing that out, obviously missed it.

PR #1383, opened 2020-05-16
Simplify log level environment variable usage

sterchelen:
- Add _LEVEL as suffix to CONSUL_TEMPLATE_LOG
- Remove CT_LOG
- Add reference to the environment variable on the readme file

ref #1328

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -517,6 +517,7 @@ func (c *Config) Finalize() {
 		c.LogLevel = stringFromEnv([]string{
 			"CT_LOG",
 			"CONSUL_TEMPLATE_LOG",
+			"CONSUL_TEMPLATE_LOG_LEVEL",
 		}, DefaultLogLevel)
 	}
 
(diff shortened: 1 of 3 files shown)


sterchelen:
Hey, any news about this PR?

config/config.go
```diff
@@ -525,8 +525,7 @@ func (c *Config) Finalize() {
 
 	if c.LogLevel == nil {
 		c.LogLevel = stringFromEnv([]string{
-			"CT_LOG",
-			"CONSUL_TEMPLATE_LOG",
+			"CONSUL_TEMPLATE_LOG_LEVEL",
```
eikenb, line 528: How about we just make it all 3. I really don't like breaking backwards compatibility unless absolutely necessary, but I think the new one is better. Thanks.

sterchelen: Yes agree with you. I will do it this week.

sterchelen: @eikenb As expected, just adding the new env variable

PR #1388, opened 2020-06-23
Change whitelist/blacklist to allowlist/denylist

### LINKED ISSUE #1387: Replace whitelist/blacklist terminology with allowlist/denylist
Update consul-template to use inclusive language, notably replacing the usage of whitelist/blacklist.

### Consul Template version
```
consul-template -v
 v0.25.0 ()
```

### Expected behavior
Configuration options for exec
```
exec {
  env {
    allowlist = []
    denylist = []
  }
}
```

### Actual behavior
```
exec {
  env {
    whitelist = []
    blacklist = []
  }
}
```

### References
- This mirrors Consul's efforts in https://github.com/hashicorp/consul/pull/7971
- Config documentation in [README.md](https://github.com/hashicorp/consul-template/blob/v0.25.0/README.md#configuration-file-format)

findkim:
Updates Consul Template to use inclusive language while keeping backward compatibility for the changed configuration option names. Configuration documentation, inline documentation, and variable names are updated to have minimal occurrences of the renamed values.

* `template.function_blacklist` is renamed to `template.function_denylist`
* `exec.env.blacklist` is renamed to `exec.env.denylist`
* `exec.env.whitelist` is renamed to `exec.env.allowlist`

We encourage practitioners to update their configuration to match these changes, but existing configurations are still backward compatibile.

Resolves #1387


config/config_test.go
```diff
@@ -384,16 +384,16 @@ func TestParse(t *testing.T) {
 			false,
 		},
 		{
-			"exec_env_blacklist",
+			"exec_env_denylist",
 			`exec {
 				env {
-					blacklist = ["a", "b"]
+					Denylist = ["a", "b"]
```
eikenb, line 390: lowercase?

findkim: good catch! The test still passes -- it's interesting that the keys are case insensitive 🤔

PR #1392, opened 2020-07-10
README: Add a note on parseFloat and printf

### LINKED ISSUE #1391: Float value truncated in rendered output
### Consul Template version

```sh
./consul-template -v
 v0.25.0 ()
```
Tested with commit c831cdcce03bfdcaf509590326eb2b465ebff1bf (master at the time of filing this issue)

### Steps to reproduce

```sh
$ cat in.tpl
{{ "1.0" | parseFloat }}
{{ "1.1" | parseFloat }}

$ ./consul-template -template "in.tpl:out.txt" -once

$ cat out.txt
1
1.1
```

### Expected output

```
1.0
1.1
```

### Actual output

```
1
1.1
```

### fmt

I understand that `fmt.*` family of functions exhibit similar behaviour. 

```go
func main() {
	f := 1.1
	fmt.Println(f)  // prints 1.1
	f = 1.0
	fmt.Println(f). // prints 1
}
```

### Usecase

I'm using consul-template to render a config file (in TOML format).

```toml
[SomeConfig]
transaction_sampling = {{ key "transaction_sampling" || parseFloat }}
```

is rendered into

```toml
transaction_sampling = 1
```

when value of key in consul is `1.0`

A go program then loads the rendered config file and unmarshals the config into following struct:

```go
type Config struct {
    TransactionSampling float64 `toml:"transaction_sampling"`
}
```

I'm using https://github.com/pelletier/go-toml for unmarshaling toml into a struct. And since the value is rendered as `1` and not `1.0`, the unmarshaling into `float64` [fails](https://github.com/pelletier/go-toml/pull/390). It is not unreasonable for `go-toml` library to fail this implicit conversion.

ppai-plivo:
Closes #1391

### DIFF
--- a/README.md
+++ b/README.md
@@ -1792,6 +1792,19 @@ Takes the given string and parses it as a base-10 float64:
 {{ "1.2" | parseFloat }}
 ```
 
+Note: The float64 value is rendered as per [Go's text/template][text-template]
+package and has the same precision and format in output as when
+`fmt.Fprint(w, value)` is called. For example, when the input string is `1.0`,
+the rendered output is `1`. However, if you pipe it to another function, it is
+passed as a floating point, so with the built-in printf you can format it as
+you like, for example:
+
+```liquid
+{{ "1.0" | parseFloat | printf "%.1f" }}
+```
+
+This is rendered as `1.0`.
+
 ##### `parseInt`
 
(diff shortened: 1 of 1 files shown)


eikenb:
Thanks!

Issue #1393, opened 2020-07-10
consul-template does not render an empty file


lneva-fastly:
### Consul Template version

`consul-template v0.25.0 (99efa642)`

### Problem Description

If a template produces no content (totally empty file), consul-template will not render an empty file to disk.  consul-template will act as if it has rendered the template but there will be no empty file on disk.

To reproduce, create a template that produces no content, a totally empty file (no whitespace).  Ensure that the destination file is empty as well.  Run consul-template in debug mode.

You will see a "rendering ..." line but no "rendered ..." line.

### Expected behavior

Consul-template should do what I told it to do: render an empty file.

### Actual behavior

If the destination file does not exist, no empty file is rendered.  If the destination file exists and is _not_ empty, it is replaced with an empty file.

### Discussion

Consul-template checks to see if the destination file exists with the desired content before rendering.  A nonexistent file is treated as if it were an existing, empty file, and therefore CT decides that it does not need to render the file since the content has not changed.

Why would I care about this?  Well, I can think of scenarios in which I would want an empty file to be rendered as a placeholder until content is available in consul KV or somesuch.

In my case though, I accidentally had an empty entry in Vault, which I was unaware of.  I was incredibly confused as to why consul-template silently failed to render this one file and spent a lot of time debugging consul-template itself, tried upgrading, etc.  If CT had rendered the file as empty (as I ostensibly told it to) that would have helped me narrow in on my problem much sooner.

### Proposed Solution

Always render the template to disk if the target file does not exist.  Skip the byte equality check in the `Render` function.

eikenb:
Hey @lneva-fastly, thanks for reporting.

I'm debating if this is an enhancement or a bug. The reason it does this is that the code that renders the file to disk only does that if the contents are different and it also doesn't error on file not existing... so it compares the returned value from a non-existing file, an empty byte slice ([]byte), with the returned value form the empty eval'd template. They are they same, both empty, and it skips it. 
(Note I saw that you mentioned the equality check, I put this in here as context for others.)

The behavior is not completely explicit, but pretty obvious. Though it isn't mentioned in the commit message that added the check as it only mentioned the side effect of not running the commands.

Basically we need to decide on the importance of the current behavior. If the current behavior seems like something that is relied upon then we'd need to fix this in a backwards compatible way (enhancement)  but if we don't think it is relied upon much (bug) we could just fix it with a notice in the changelog.

Thoughts?

lneva-fastly:
[...]

PR #1397, opened 2020-07-23
render initial file even when template is empty

eikenb:
When first rendering a template to a file if the file doesn't exist and
the evaluated template is empty, it wouldn't write the file to disk. It
should always render the file to disk on successful template evaluation.

Fixes: #1393


eikenb:
Ported this from [hashicat](https://github.com/hashicorp/hcat) where I had already fixed it.

PR #1403, opened 2020-08-12
add options to set pgid while creating a child process

sudhabindu1:
This PR adds a `setPgid` bool field to `Child` struct, so if the users want the child process (and any other processes they might spawn) to have their own pgid, they can set the field value to be true.


eikenb:
Hey @sudhabindu1, thanks for the PR.

Would you might giving me a bit more details on your use case for this? To use this it looks like you must be using consul-template as a library and I'm collecting data on every case of this to help with my library refactor (https://github.com/hashicorp/hcat).

Thanks!

sudhabindu1:
@eikenb You're right. I am using `child` package of consul-template in an init system / process supervisor and the PR implements the requirements that we need in terms of flexibility in spawning child process.

sudhabindu1:
@eikenb Am I missing something to get this under review process?

eikenb:
Hey @sudhabindu1,

You're not missing any steps. I'm just going through and reviewing things in passes, swinging back as people reply and I make it back to them. 

You are missing one thing for this, a test. Could you add a test to child_test.go that exercises this new feature.  A simple one like TestPid() should suffice. Thanks.

eikenb:
Note to myself that I reviewed the use of syscall.SysProcAttr and it doesn't use any of the platform specific fields.

child/child.go
```diff
@@ -157,6 +171,29 @@ func (c *Child) Pid() int {
...
+		log.Fatalln(err)
+	}
+	return pgid
+}
+
+// Sid returns the session ID of the child process
+func (c *Child) Sid() int {
+	c.RLock()
+	defer c.RUnlock()
+	sid, err := unix.Getsid(c.cmd.Process.Pid)
```
eikenb, line 190: Use of x/sys/unix isn't cross platform and while we don't officially support windows at this point, I don't like to purposely break it. Would you please consider refactoring this into an OS specific file with a stub/dummy version for windows.

Sorry I missed this earlier when I was checking the SysProcAttr use... not sure how I missed it.

Thanks.

eikenb:
Hey @sudhabindu1,
 
As part of working on a current issue with shell parsing, I'm reworking calls to all use a subshell and need to add setpgid support as that requires it to send signals properly. I'm just doing it separately instead of trying to merge this patch as I needed to handle things a bit differently as I wanted all signals to go to the process group instead of the individual  processes.

I just pushed up the PR,  #1494. Assuming it is merged, I'll be closing this PR. If you have any tweaks for your use case at that point I'd be happy to look into that.

So sorry for not merging this, but hopefully getting this change in is good news.

Thanks.

eikenb:
I've merged the setpgid support and if it doesn't meet your needs please open an issue or submit a new PR with the changes you'd like. Thanks.

sudhabindu1:
Thanks, I will check it out! @eikenb

PR #1406, opened 2020-08-17
[feature] Add go-spew for debugging templates

angrycub:
This PR adds go-spew to provide a `spew_dump` and `spew_printf` template function, which can be used in the process of template development or debugging. 

I am totally flexible on what we call the functions and just used `spew_` as a first proposal.  Future work might include the ability to provide a spew configuration.


eikenb:
Hey @angrycub, thanks for another PR!

Would you might adding a small entry to the documentation on these methods in docs/templateing-language.md. They can be combined in just one entry.. mainly just don't want functions added without any mention in the docs.

Thanks!

eikenb:
Thanks for the documentation updates!

The `liquid` -> `golang` switch in the code syntax notation confused me for a bit as it changed very little in the highlighting, but it did seem like it was more technically correct and was nicely split into it's own commit. So I decided I'm good with it.

Thanks again.

PR #1421, opened 2020-10-30
switch to hashicorp docker mirror

alvin-huang:
I only modified the CI config. I didn't modify anything in `docker/*/Dockerfile` since those are usually run locally and you need credentials to push anyway so you will be authenticated to pull from Dockerhub.


eikenb:
Thanks to you kind sir!

PR #1312, opened 2019-11-12
POC: Add sprig library

angrycub:
This PR adds the [Sprig template functions](https://github.com/Masterminds/sprig) library.  For clarity and ease of documentation, the sprig functions are renamed `sprig_«function»`.

PR #1425, opened 2020-11-06
Allow external FuncMaps

angrycub:
Since Nomad uses CT as its underlying template engine, it would be nice to be able to pass functions from Nomad itself into the template configurations that it generates.  I thought exposing a route to inject functions for applications that use CT internally might be a good middle ground to prevent even more template application sprawl.

This would allow Nomad to inject purpose-built templating functions into the rendering pipeline without having to make additional changes to consul-template.

Custom Functions could also be used to simplify the FunctionBlockList by allowing someone to implement their own no-op function to opaque an implementation provided in CT.

This could let us resolve #1312 and #1406 by allowing them to be passed into CT from Nomad using this scheme.

### DIFF
--- a/config/template.go
+++ b/config/template.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"regexp"
 	"strings"
+	"text/template"
 	"time"
 )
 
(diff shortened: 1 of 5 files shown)


eikenb:
Hey @angrycub,

We're already working on this feature as part of [hashicat](https://github.com/hashicorp/hcat), the in-progress library which replaces consul-template (as a library). It's progress was paused for a bit due to [CTS](https://github.com/hashicorp/consul-terraform-sync/) deadlines around hashiconf, but I'm working on it again now.

eikenb:
We finally have CTS done enough that I've been fully freed up to get back to work on other projects and am reviewing these PRs again. I'd like to just reject/close this one, looking to moving Nomad hashicat for the feature. 

@angrycub, is that OK?

angrycub:
Sure. No reason to keep this one.

PR #1431, opened 2020-12-11
fix: clarify kvv2 .data value is not a key reference, but a kvv2 syntax change

druwadi:
Putting this readme change in, the current note wasn't explicit enough for me to realize that the extra `.data` value was syntax and not referencing something key related while reading through the docs.


eikenb:
Thanks for the PR @druwadi, that's a nice clarification of the kv2 syntax. Thanks.

Issue #1047, opened 2017-11-15
configuration parameter "error_on_missing_key" not available via cli


romonopoly:
Would it be possible to allow "error_on_missing_key" to be configured for a template via command line call?

sethvargo:
Hi @romonopoly 

Thank you for opening an issue. That option is template specific. The only template options we permit on the command line are the source file, destination file, and command. All other arguments must be in the `template` stanza in a configuration file.

romonopoly:
Hi @sethvargo,

Thank you for a quick response.

I'm using the once mode as follows:
`consul-template -once -template "$configFile.ctmpl:$configFile"`
where the variable $configFile changes.

Is there a way to use "error_on_missing_key" with my use case, without generating a new config file for every invocation of consul-template?

romonopoly:
Hi @sethvargo,

Is it possible to reopen this issue based on my comment above?

sethvargo:
HI @romonopoly 

No, this is not possible.

romonopoly:
Hi @sethvargo, could you provide a more professional response?

Something along the lines of: "using -once is not a fully supported feature and is intended only for testing"
Or: "I'm very busy and don't have time to examine this at the moment"
Or: "Please create a PR if you'd like this functionality"
Etc.

sethvargo:
Hi @romonopoly 

I apologize if you think my response was unprofessional. I have answered your question multiple times, and I'm not sure how I can make it more clear. 

The `error_on_missing_key` key is a template-specific option and therefore must be in the template-specific configuration stanza in a configuration file. The CLI is a one-dimension (key=value). Since it's possible to specify multiple templates to one Consul Template process, there's no way to represent these template-specific options without a configuration file.

For example, let's say I have two templates - how would I specify the first should be error_on_missing_key, but they second should not? There are also many other template-specific options like the delimiter to use, file permissions for the destination template, whether to create destination directories, etc. None of those values can be specified on the command line because they are template-specific. 

It's not mathematically possible to map a multi-dimensional array into a single string. This is what I mean when I say "this is not possible". It's not possible. There is no pull request to open. There is no limiting time factor. The limiting factor is mathematics.

romonopoly:
Hi @sethvargo,

I apologize for my misunderstanding of your previous comments, thank you for the clarification.

aclowkey:
+1
[...]

PR #1420, opened 2020-10-29
add template-error-fatal

phemmer:
This adds a few `template-error-fatal` options to allow consul-template to continue running in the event of an error while rendering a template.  
The option is exposed individually within each template, globally within the config, and as a global command line flag. The individual template setting overrides the global config setting, and the command line flag overrides all.

No tests yet, but I can create them upon review.

Closes #1419 #1289

PR #1434, opened 2020-12-16
Add global -error-on-missing-key CLI argument

imalsogreg:
This addresses #1047 by adding a `-error-on-missing-key` CLI argument.

The existing `error_on_missing_key` configuration field is template specific. So the meaning of the CLI argument is a default, which will be applied to each template configuration that doesn't specify its own specific `error_on_missing_key`.

This allows a user to more conveniently turn on missing-key error checking in situations where it is not convenient to write per-template configs, e.g. when we have many calls to `consul-template` that already use CLI arguments rather than config files to specify template source and target locations.


eikenb:
Hey @imalsogreg, thanks for the PR.

I'm not sure about this as the idea was already turned down for pretty good reasons. I see that this attempts to work around those concerns by having it change the global default instead of being a global override, which makes sense. But I don't think it convinces me.

If I were to consider it, I'd want it reworked as I don't like adding arguments to Finalize. That call only deals with making sure things are internally good based on the current config... passing an arg changes the semantics and I already think the config semantics are overly complex.

Thinking about it some more I have a few lines of thought...
1. I think consul-template has to many arguments now, and those are kind of all over the place.
2. @jsok's solution in the comments of #1047 has a certain appeal as it's more general (allows you to set any template specific options) and builds on a existing (already complex) argument instead of adding a new one.
3. really seems to me that error_on_missing_key should default to "true" when using `-once`, but I'm loath to change that  as it would certainly break something.

imalsogreg:
@eikenb Thanks for having a look!

What would your preference be, between
1. Keep the current behavior
2. Try to rework my global-default logic so that it doesn't add arguments to `Finalize()`
3. Implement the `in_tmpl:out_tmpl[:opts]:cmd` suggestion from issue #1047 
4. Change the behavior of `-once` to be strict (breaking old users)
5. Add `-once-strict`

Totally understand if you don't want to add complexity to the config story here :)

eikenb:
A colleague pointed out #1420 which looks pretty similar to this in ways and has me leaning toward your 2nd option (rework w/o Finalize change). You could probably even find some ideas on how to work around the Finalize thing there.

I do have to say your 5th option does have a certain appeal. Maybe even changing `-once` to take an optional argument so `-once` is like now but `-once=strict` adds this. Not 100% on this but it seems to have some promise. I guess it'd depend on how easy that to thread through the logic to where it'd need to go.

Really threading through the logic is the hardest part given the current code organization. Whichever does that best would be my first choice... maybe compare what #1420 does vs. converting config.Once from a boolean to a enum-ish type and see which looks cleaner.

Thanks.

toransahu:
@imalsogreg Thanks for the pull request. I also want this to be merged as a consumer of `consul-template`. BTW, what has stopped this from being merged?

PR #1449, opened 2021-01-25
remove telemetry docs

eikenb:
Telemetry was added and then removed and the docs were added and
refactored in the window they were present. I missed that they were
present and this removes them.


findkim (approved):
Thank you for the docs merge + cleanup! Linking the original PR for posterity https://github.com/hashicorp/consul-template/pull/1378

PR #1459, opened 2021-03-13
Docs: Change whitelist/blacklist to allowlist/denylist

fhemberger:
PR #1388 already changed whitelist/blacklist to allowlist/denylist, but the
example in the configuration documentation was left out.


eikenb:
Thanks for catching this @fhemberger, I think we lost these updates in a merge conflict and missed it.

PR #1461, opened 2021-03-30
Fix/master: add envOrDefault helper

jtrinklein:
Addresses issue #829 

Adds a helper `envOrDefault` as a compliment to the `env` helper. This behaves similar to how `keyOrDefault` behaves.

The helper will attempt to retrieve an environment variable. If the variable is found and if the value has a non-zero length, the value is used. Otherwise, the default value is used.

Change includes tests and documentation.


jtrinklein:
Hi @eikenb here's the PR for the envOrDefault helper

template/funcs.go
```diff
@@ -80,6 +80,30 @@ func envFunc(env []string) func(string) (string, error) {
...
+			split := strings.SplitN(e, "=", 2)
+			k, v := split[0], split[1]
+			if k == s {
+				if len(v) == 0 {
+					return def, nil
+				}
+				return v, nil
+			}
+		}
+		val := os.Getenv(s)
```
eikenb, line 99: I was going to suggest using os.LookupEnv() here to only use the default if the Environment variable was unset, to allow for setting it to empty string.. but the description specifically calls out the behavior of using the default when the env is zero length. 

Was there a use case where this behavior was specifically desired or is it more just by chance? If the latter then I would suggest switching to LookupEnv to allow for the broader use case with empty strings.

Thanks.

jtrinklein: I don't have a specific use case for when a variable is present and empty vs not present. I'll switch it to use LookupEnv and update the documentation

eikenb: Thanks!

eikenb (approved):

PR #1464, opened 2021-04-23
Adds an md5sum function

geofffranks:
Useful for computing md5sums of strings to be used in places
where sha256 is not supported.


eikenb:
A bit surprised this isn't already present. Thanks for the PR @geofffranks, looks good.

PR #1470, opened 2021-05-16
add: make targets for GOARM ARM5/6/7

Dombo:
### Why
I'm setting up a network of paspberrypi's (mixed versions, acquired over the past few years) and need consul-template to support my architecture (ARMv7).

### What
I have simply extended the dynamic make target in line with the golang documentation on [supported architectures](https://github.com/golang/go/wiki/GoArm#supported-architectures).

### Reviewer
I built and tested this on my local - arm5/6/7. The compiled binaries worked on the various arm 32bit pi's I have here.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ GOARCH ?= $(shell go env GOARCH)
 
 # Default os-arch combination to build
 XC_OS ?= darwin freebsd linux netbsd openbsd solaris windows
-XC_ARCH ?= 386 amd64 arm arm64
+XC_ARCH ?= 386 amd64 arm5 arm6 arm7 arm64
 # XC_EXCLUDE "arm64" entries excludes both arm and arm64
 XC_EXCLUDE ?= darwin/arm64 freebsd/arm64 netbsd/arm64 openbsd/arm64 solaris/arm64 windows/arm64 solaris/386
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -54,16 +54,26 @@ define make-xc-target
   else
 		@printf "%s%20s %s\n" "-->" "${1}/${2}:" "${PROJECT}"
 		case "$2" in \
-			arm) export CGO_ENABLED="1" ; \
+			arm5) export CGO_ENABLED="1" ; \
```
eikenb, line 57: These all fail to compile for me with...

`arm-linux-gnueabi-gcc: error: unrecognized command line option '-m64'`

Disabling CGO (`export CGO_ENABLED="0"`) fixes it.

We could disable CGO as the bug that required that has been fixed (and theirs even a PR to remove it, #1471), but I'm curious if it did work for you and if so if you knew why? I've been digging around but haven't found anything useful yet.

eikenb:
Given the fix for this was removing CGO anyways, keeping the Arm switch makes no sense anymore and everything should just work now with the pure Go builds for Arm.

eikenb:
@Dombo, if for some reason these Arm/GCC cases are still needed please open a new bug.

CGO was enabled previous to work around a bug in the Go compiler. It was fixed several releases back, far enough to not worry about it anymore and I went ahead and just disabled CGO completely again. Thanks again for this PR even though we didn't use it.

Dombo:
@eikenb you're welcome, glad to try and help. Sorry about the long reply, don't have much energy during the week to put into stuff like this.

I'll test the 32 bit arm artefact when it's up on the release page.

eikenb:
@Dombo .. Great, thanks!

PR #1474, opened 2021-05-27
make connect and health service ids different

### LINKED ISSUE #1458: Wrong deduplication for `service` and `connect` dependencies
`dependency.HealthService.String()` returns the same value for `service` and `connect` invocation, and that breaks templates when both functions are called with the same parameter.
In my case I have an external service registered in the catalog, and a terminating-gateway configured for it. I want to get IP:Port from `connect`,  and some metadata from `service`, and that doesn't work
### Consul Template version

consul-template v0.25.1

### Configuration

```hcl
consul-template -dry -once -template ./sample.tpl
```

This template is not very useful but it's perfect to pinpoint the bug
```liquid
{{ range service "myservice" }} {{ .Name}} {{end}}
{{ range connect "myservice" }} {{ .Name}} {{end}}
{{ printf "%#v" (service "myservice") }}
{{ printf "%#v" (connect "myservice") }}
```

"myservice" is registered directly into consul catalog, and a terminating-gateway is assigned to it via config entry

### Command

```shell
consul-template -dry -once -template ./sample.tpl
```

### Expected behavior

I would expect the output to be:
```liquid
[]*dependency.HealthService{(*dependency.HealthService)(0xc000001234)}
[]*dependency.HealthService{(*dependency.HealthService)(0xc000005678)}
 myservice
 terminating-gateway
```

### Actual behavior

What actually happened?

```liquid
[]*dependency.HealthService{(*dependency.HealthService)(0xc000001234)}
[]*dependency.HealthService{(*dependency.HealthService)(0xc000001234)}
 myservice
 myservice
```

### Steps to reproduce

Any template that uses both functions will trigger it.

eikenb:
The IDs need to be different as they are used as the cache keys and
using the same ID means `connect` and `service` would overwrite each
other if both used. (this is bad)

The `connect` function re-uses the health `service` code as it is
identical except for an additional option. When adding connect the ID
was unintentially left the same, it should be different.

Fixes #1458

### DIFF
--- a/dependency/health_service.go
+++ b/dependency/health_service.go
@@ -237,6 +237,9 @@ func (d *HealthServiceQuery) String() string {
 	if len(d.filters) > 0 {
 		name = name + "|" + strings.Join(d.filters, ",")
 	}
+	if d.connect {
+		return fmt.Sprintf("health.connect(%s)", name)
+	}
 	return fmt.Sprintf("health.service(%s)", name)
 }
 
--- a/dependency/health_service_test.go
+++ b/dependency/health_service_test.go
@@ -463,3 +463,37 @@ func TestHealthServiceQuery_String(t *testing.T) {
 		})
 	}
 }
+
+func TestHealthServiceQueryConnect_String(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		name string
+		fact func(string) (*HealthServiceQuery, error)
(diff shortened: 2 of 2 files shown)


findkim (approved):
Simple fix that makes sense, appreciate the community bug ticket identifying this problem!

PR #1480, opened 2021-06-10
fix data race issue setting vault's default lease

eikenb:
It is using a module/global variable at the moment and race tests are
normally run, so it was missed during review. This is the common
sync.Once pattern for this sort of thing that fixes the race.


lornasong (approved):
Makes sense to me! I hadn't come across the sync.once pattern before - glad to learn about this!

PR #1483, opened 2021-06-15
remove stuff related to old build system

eikenb:
clear out Makefile targets for release and Docker builds
remove all old Docker configuration and support files

Issue #1481, opened 2021-06-11
New simpler Docker image in 0.26.0 is insufficient (need removed features)


eikenb:
I created this issue as I expect there to be a at least a few people form whom the new Docker images will break their current workflows. If this is you please comment here with...

1. the image you currently use
2. which features you rely on

I apologize for any inconvenience this may have caused and if a simple base image can't be made to work for everyone, I'll be sure to make getting a 0.26.1 with an updated docker image out a top priority. I didn't want to just propagate the old Docker setup as it needs some serious work, but I really wanted some feedback on what's currently used before doing that.

Thanks.

jklawrence:
1. We were using `0.25.2-alpine`.
2. We relied on the `ENTRYPOINT` instruction to specify the executable `/bin/consul-template` for us.

We pass arguments for the consul-template executable directly in our Kubernetes deployment file's args field, like:
```
        args:
          [
            "-config=/path/to/mounted/hcl/file.hcl",
            "-log-level=debug",
            "-once",
          ]
```
Docker equivalent:
```
docker run hashicorp/consul-template -config=/etc/consul-template/test.hcl -log-level=debug -once
```

Now that there is no entrypoint we need to specify the executable in addition to its arguments:
```
        args:
          [
            "/bin/consul-template",
            "-config=/path/to/mounted/hcl/file.hcl",
            "-log-level=debug",
            "-once",
          ]
```
Docker equivalent:
```
docker run hashicorp/consul-template /bin/consul-template -config=/etc/consul-template/test.hcl -log-level=debug -once
```

eikenb:
Hey @jklawrence, thanks for replying. I really appreciate the feedback.

Mind if I ask how do you get the config file in the container? Do you include it as part of a derived image (using our image as a base image and extending it), or via a bind mount  or some other way?

Thanks.

jklawrence:
Hi @eikenb. We use the `consul-template` image directly and add the config file via a bind mount.

For background, we make use of both the `vault` and `consul-template` containers as initContainers to load secrets from our Vault instance and output formatted configuration file(s) for our application container.

eikenb:
Thanks. I'll be pushing up a PR with a new Dockerfile + supporting scripts soon. Condensing things down to 1 Alpine Linux based image where I've reviewed, modernized and simplified things as I can.

eikenb:
PR is up... #1484.

jklawrence:
[...]

PR #1484, opened 2021-06-15
new docker image build setup

eikenb:
Update to use new standard of downloading application from official release sources, including cryptographic verification.

Cleans up entrypoint and makes separate init process optional. Standardizing on a minimal Alpine Linux based image.

Note dumb-init was removed as it is not required to for consul-template, as it handles all the signaling properly, and docker has it built in so if a user wants it they can add it by running with `--init`. Shouldn't normally be necessary but might be useful if you have ill-behaved template commands.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,43 @@
+# This Dockerfile creates a production release image for the project. This
+# downloads the release from releases.hashicorp.com and therefore requires that
+# the release is published before building the Docker image.
+#
+# We don't rebuild the software because we want the exact checksums and
+# binary signatures to match the software and our builds aren't fully
+# reproducible currently.
+FROM alpine:3
+
+# NAME and VERSION are the name of the software in releases.hashicorp.com
+# and the version to download.
+ARG NAME=consul-template
+ARG VERSION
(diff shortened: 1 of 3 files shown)


eikenb:
See #1481 for some context.

docker/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+# This is the location of the releases.
+ENV HASHICORP_RELEASES=https://releases.hashicorp.com
+
+# Create a user and group first so the IDs get set the same way,
+# even as the rest of this may change over time.
+RUN addgroup -g ${GID} ${NAME} && adduser -u ${UID} -S -G ${NAME} ${NAME}
+
+# Set up certificates, base tools, and software.
+COPY fetch-n-verify.sh docker-entrypoint.sh /
+RUN /fetch-n-verify.sh # removes self when done
```
findkim, line 37: I like this new approach to download the official released binary instead of building from source. Separates the dockerfile from the release process and the bulkiness of building from source.

findkim (approved):
This looks like a solid and slim replacement of what's being removed in #1483

Issue #1489, opened 2021-07-01
Secret function doesn't honour at-sign (@) to load file contents into an attribute


tomwerneruk:
### Consul Template version

compiled from main @ commit 51401f314b34a4a44f8e82a2320e609d6da5553a

### Configuration

```hcl
vault {
  address = "http://127.0.0.1:8200"
}

template {
  source      = "./in.tmpl"
  destination = "./out.cer"
}

```

```liquid
{{ with secret "pki_int/sign/example-dot-com" "ttl=60s" "csr=@./MYCSR.csr" }}
{{ .Data.certificate }}
{{ end }}
```
### Command

```shell
consul-template -template "in.tmpl:out.cer"
or
consul-template -config config.hcl 
```

### Expected behavior

The Vault convention of prefixing a file with an @ then causes the contents to be read and used as the value

### Actual behavior

The string is used literally, passing the value '@./MYCSR.csr' to Vault, causing an HTTP 400

### Steps to reproduce

1. Create a template and CSR as above
2. Configure Vault using https://learn.hashicorp.com/tutorials/vault/pki-engine?in=vault/secrets-management
2. Execute consul-template against Vault

PR #1490, opened 2021-07-01
Add vault file handing to secret function

tomwerneruk:
Proposed fix for #1489 to load file contents when an attribute is prefixed by an @ sign.


eikenb:
Hey @tomwerneruk,

Sorry for the long sleep on this issue. I was finally getting around to reviewing and including the community PRs and this one looks good except for a missing test. If you are still interested would you consider adding a test? Thanks.

angrycub:
I was looking at this and thought it looked interesting, but could you accomplish the same behavior with a template variable, `file`, and `secret`?

I'd be worried about keeping the @-reference safe, since some implementations that depend on CT explicitly deny `file`, but you wouldn't be able to flag out this behavior.

PR #1494, opened 2021-07-27
Add setpgid for all called commands

eikenb:
Need to make sure signals are sent to all processes as sub-shells interfere with signal propagation. Using Setpgid to create a process group for the command and all it's subprocesses fixes this by allowing child to send the signals to the pgid, which thus gets sent to all processes in that group.

This is required for upcoming change to use sub-shells for all executed commands to eliminate need for shell parsing.

With this change it is always set to true and the pgid is always used to send signals to the process(es).

In the future it could be made configurable if needed.

Might be easier to review using the commits. The second commit is test cleanups and fixes where the first one contains the new code and tests.

### DIFF
--- a/child/child.go
+++ b/child/child.go
@@ -65,6 +65,9 @@ type Child struct {
 	stopLock sync.RWMutex
 	stopCh   chan struct{}
 	stopped  bool
+
+	// whether to set process group id or not (default on)
+	setpgid bool
 }
 
 // NewInput is input to the NewChild function.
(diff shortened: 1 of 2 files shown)


child/child.go
```diff
@@ -353,7 +358,18 @@ func (c *Child) signal(s os.Signal) error {
 	if !c.running() {
 		return nil
 	}
-	return c.cmd.Process.Signal(s)
+
+	sig, ok := s.(syscall.Signal)
+	if !ok {
+		return fmt.Errorf("[ERR] (child) bad signal: %s", s)
```
findkim, line 364: This is formatted more as of a log line than an error message. Can you update it?

findkim (approved):
👍👍

child/child.go
eikenb: Missed that in a refactor. Thanks.

PR #1495, opened 2021-07-28
Add writeToFile function

### LINKED ISSUE #1077: Generating multiple similar files from the same template
I happened to need this several times until now and I'm pretty sure I am not the only one. A few ways I see it to be implemented:

1. Add a writeToFile function in the template. This could be used in conjunction with nested templates and get the thing going really fast, probably also very fast to implement (would be more difficult to write tests I guess than actually implement it). Not sure what would be the biggest downsides of this.

2. Add a template_set functionality in the configuration file. The biggest problem I see is that usually you want to generate this kind of templates by looping through a list of keys. This would introduce a "first" in the consul-template code, as the engine would have to implement logic outside the template to do this kind of looping according to some pattern presented by the user.

If we agree on a design together, I can work on it and submit a PR myself.

Thanks,

anhdat:
Fixes #1077
Based on the implementation of [savetofile plugin](https://github.com/ekristen/consul-template-plugin-savetofile).
`writeToFile` writes the content to a file with username, group name, permissions. There are optional flags to select appending mode or add a newline.

For example:

```golang
{{ key "my/key/path" | writeToFile "/my/file/path.txt" "my-user" "my-group" "0644" }}
{{ key "my/key/path" | writeToFile "/my/file/path.txt" "my-user" "my-group" "0644" "append" }}
{{ key "my/key/path" | writeToFile "/my/file/path.txt" "my-user" "my-group" "0644" "append,newline" }}
```

This is my first PR in this repository, please guide me on how to finish this PR. Thank you in advance!

### DIFF
--- a/child/child.go
+++ b/child/child.go
@@ -65,6 +65,9 @@ type Child struct {
 	stopLock sync.RWMutex
 	stopCh   chan struct{}
 	stopped  bool
+
+	// whether to set process group id or not (default on)
+	setpgid bool
 }
 
 // NewInput is input to the NewChild function.
@@ -135,6 +138,7 @@ func New(i *NewInput) (*Child, error) {
 		killTimeout:  i.KillTimeout,
 		splay:        i.Splay,
 		stopCh:       make(chan struct{}, 1),
+		setpgid:      true,
 	}
 
 	return child, nil
@@ -264,6 +268,7 @@ func (c *Child) start() error {
 	cmd.Stdout = c.stdout
 	cmd.Stderr = c.stderr
 	cmd.Env = c.env
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: c.setpgid}
 	if err := cmd.Start(); err != nil {
 		return err
 	}
@@ -353,7 +358,18 @@ func (c *Child) signal(s os.Signal) error {
 	if !c.running() {
 		return nil
 	}
-	return c.cmd.Process.Signal(s)
+
+	sig, ok := s.(syscall.Signal)
+	if !ok {
+		return fmt.Errorf("bad signal: %s", s)
+	}
+	pid := c.cmd.Process.Pid
+	if c.setpgid {
+		// kill takes negative pid to indicate that you want to use gpid
+		pid = -(pid)
+	}
+	// kill is syscall's only signal API
+	return syscall.Kill(pid, sig)
 }
 
 func (c *Child) reload() error {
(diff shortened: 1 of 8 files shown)


eikenb:
Thankjs for the PR @anhdat,

I know this is frequently requested and will definitely look at this for the next feature release. I'm currently debating about doing a bugfix 0.26.1 release or just jumping to 0.27.0... if the latter I'll definitely include this.

eikenb:
One thing... I'm not going to consider it a blocker, as I'd like to get this feature added, but if you get a chance I'd love to see a test added for this. If you decide to add test(s) there are 2 possible places, in template_tests.go as an entry in the test table in TestTemplate_Execute or you can  test it directly in funcs_test.go. Either one would be great and, if you do decide to write the test, I'll leave it to you to pick which you method you prefer.

Thanks.

anhdat:
Hi @eikenb, thanks for the guidance. I've added the tests to the `template_tests.go` file. The `writeToFile` function requires a different testing setup, so I created a function `Test_writeToFile` alongside `TestTemplate_Execute`. Please let me know if this works. Thanks.

anhdat:
[The build](https://app.circleci.com/pipelines/github/hashicorp/consul-template/580/workflows/700c51bf-3a35-4ea3-a26d-37165961da7e/jobs/588) is failing in `consul-template/manager`:

```
foo--- FAIL: TestDedup_IsLeader (2.00s)
    dedup_test.go:46: timeout
--- FAIL: TestDedup_UpdateDeps (2.00s)
    dedup_test.go:76: timeout
--- FAIL: TestDedup_FollowerUpdate (2.00s)
    dedup_test.go:135: timeout
FAIL
FAIL	github.com/hashicorp/consul-template/manager	5.094s
```

Would a retry fix this?

eikenb:
Thanks for the tests!

A re-run did fix that test failure. I've been working to eliminate them, but there are still a number of race conditions lurking in the tests. That dedup one has been particularly annoying.

The github-actions CI is still failing due to it finding a data race in a test that passes on every other platform (???)... I'm going to look into that, but I'm only testing github-actions CI at this point so I'm not going to consider it a blocker. I think it really just comes down to the use of t.Parallel() and I plan on removing all of those very soon anyways (they make debugging tests a PITA) and then, maybe, add some back if they actually help.

eikenb (approved):

wavedocs:
Guys, I propose to supplement the code that will function to create nested directories if they are missing.

```
diff --git a/template/funcs.go b/template/funcs.go
index 66990b1c..376e6afe 100644
--- a/template/funcs.go
+++ b/template/funcs.go
@@ -1628,6 +1628,15 @@ func writeToFile(path, username, groupName, permissions string, args ...string)
                        return "", err
                }
        } else {
+               dirPath := filepath.Dir(path)
+
+               if _, err := os.Stat(dirPath); err != nil {
+                       err := os.MkdirAll(dirPath, os.ModePerm)
+                       if err != nil {
+                               return "", err
+                       }
+               }
+
                f, err = os.Create(path)
                if err != nil {
                        return "", err

eikenb:
Hey @wavedocs,

Thanks for raising your concern. I suggest submitting a PR with this change and linking it back here. Having a PR will make it easy to merge and for others to take a more active part of the discussion. If you do submit a PR with this please add a test for the added functionality in addition to the feature.

Thanks.

mbillow:
@wavedocs @eikenb I included the directory creation and subsequent tests in #1549, FYI.

PR #1496, opened 2021-07-29
replace shell parsing w/ direct shell call

### LINKED ISSUE #1482: Consul Template 0.26.0 issues with command substitution
After upgrading from 0.25.1 -> 0.26.0, commands are no longer parsed the same way. Commands that use command substitution like `sh -c 'echo $(which consul)'` fail. I noticed escaping the `$` with `\\$` works, but as far as I can tell this was an undocumented breaking change.

### Consul Template version

 v0.26.0 (3b7f233a)

### Configuration

```json
{
  "template": {
    "source": "/tmp/test.tmpl",
    "destination": "/tmp/test.out",
    "command": "sh -c 'echo $(which consul)'"
  }
}
```

(note the template is arbitrary here, it's the command hook that's the issue)
```liquid
{{range services}}# {{.Name}}{{range service .Name}}
{{.Address}}{{end}}

{{end}}
```

### Command

```bash
consul-template -once -config test.json --log-level trace
```

### Debug output

https://gist.github.com/evandam/9f9aeaa444cc9761086b6ee09b073583

### Expected behavior

Command should process `sh -c 'echo $(which consul)'` as expected, returning something like `/usr/local/bin/consul`

### Actual behavior

```     
2021-06-14T23:35:29.670Z [INFO] (runner) executing command "sh -c 'echo $(which consul)'" from "/root/foo.tmpl" => "/root/foo.out"
2021-06-14T23:35:29.670Z [INFO] (child) spawning: sh -c echo (which consul)which consul)
sh: 1: Syntax error: word unexpected (expecting ")")
2021-06-14T23:35:29.672Z [ERR] (cli) 1 error occurred:
	* failed to execute command "sh -c 'echo $(which consul)'" from "/root/foo.tmpl" => "/root/foo.out": child: command exited with a non-zero exit status:

    sh -c echo (which consul)which consul)

This is assumed to be a failure. Please ensure the command
exits with a zero exit status.
```

### Steps to reproduce

1. Create template file `/tmp/foo.tmpl`
2. Create config file `test.json`
3. `consul-template -once -config test.json`

eikenb:
Instead of trying to parse the shell command line into arguments for input to exec.Command this replaces that all with passing it to an 'sh -c' call. This will "fix" the wack-o-mole game of spot fixing issues that keep coming up with the shell parsing package.

Windows is handled separately (no 'sh') and is setup to only accept/run a single command.

The code changes are pretty small, bulked out a bit by the need to handle windows separately. Most of the code here are tests I added to try to exercise different cases mentioned in previously submitted issues.

Fixes #1482

### DIFF
--- a/manager/command-prep.go
+++ b/manager/command-prep.go
@@ -1,0 +1,12 @@
+package manager
+
+const whitespace = " \t\n\v\f\r"
+
+func prepCommand(command string) ([]string, error) {
+	if len(command) == 0 {
+		return []string{}, nil
+	}
+
+	cmd := []string{"sh", "-c", command}
+	return cmd, nil
+}
--- a/manager/command-prep_windows.go
+++ b/manager/command-prep_windows.go
@@ -1,0 +1,13 @@
+// +build windows
+
+package manager
+
+func prepCommand(command string) ([]string, error) {
+	switch {
+	case len(command) == 0:
+		return []string{}, nil
+	case len(strings.Fields(command)) > 1:
+		return cmd, fmt.Errorf("only single commands supported on windows")
+	}
+	return []string{command}, nil
+}
--- a/manager/runner.go
+++ b/manager/runner.go
@@ -10,15 +10,16 @@ import (
 	"sync"
 	"time"
 
+	"github.com/pkg/errors"
+
 	"github.com/hashicorp/consul-template/child"
 	"github.com/hashicorp/consul-template/config"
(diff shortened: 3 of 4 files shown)


findkim (approved):
Nice test cases 🙂 

That makes sense about avoiding wack-a-mole from recreating shell parsing, and instead defer to the shell to handle the raw string args.

PR #1502, opened 2021-08-09
set sid for child process group

sudhabindu1:
Similar to [`setpgid` being set to true to allow the child processes to have their own pgid](https://github.com/hashicorp/consul-template/commit/5717544fdb617291f90230cbcfd553e71849b6c7). This PR adds the `setsid` flag (which is defaulted to true as of now) so as the child processes have their own session ID. 
The `setpgid` flag however, is removed for the following reasons:
1. Using `Setsid` on `SysProcAttr`, sets a new process group ID as well while creating a new session, so the `setpgid` flag becomes rudimentary since both the flags are set defaulted to `true`.
2. Setting both `Setsid` and `Setpgid` on `SysProcAttr` causes Go to fail with the error [fork/exec: operation not permitted]. Ref: https://forum.golangbridge.org/t/setpgid-pty-operation-not-permitted/1225/27

This can be added in future if users are allowed to configure these properties, with the condition that both of them can't be `true`.

@eikenb


eikenb:
Done some general research, looked at Go's implementation and did some testing... it really seems that Go's exec.Command already does all that is needed to avoid the EPERM error. So I don't see a big issue with using setsid instead of setpgid.

I also read that the only real difference is that a session can contain multiple process groups vs. a process group obviously only having one. I'm having a hard time coming up with a case where you'd need multiple process groups being controlled by consul-template. I'm going to wait until I hear back before changing the existing code as I'd like to better understand why this change is useful.

eikenb:
Found a nice article on process groups/sessions their history and how they came to work like they do. Thought I'd add it here as it was useful context. 

https://lwn.net/Articles/603762/

The article still left me wondering why one would want a session instead of a process group. I had wondered if sessions propagated signals to process groups created by children processes, but they don't. Signals are only propagated to processes via process group and only to process in that specific group.

sudhabindu1:
We are using `child` package in an init system for container (that handles some other platform integrations as well). The current behavior with pgid (= pid of immediate child ) and sid = 1 solves our problem as far as signal forwarding and process supervision is concerned. While at the moment I am not sure of a use case where a different SID for child processes would be required, we are trying to bring our init system at parity with [dumb-init behavior wise which creates a session rooted at the child](https://github.com/Yelp/dumb-init#:~:text=dumb-init%20runs%20as%20PID%201%2C%20acting%20like%20a%20simple%20init%20system.%20It%20launches%20a%20single%20process%20and%20then%20proxies%20all%20received%20signals%20to%20a%20session%20rooted%20at%20that%20child%20process), which is the motive behind this PR. Let me know if it creates any issues (I don't see any tbh)

eikenb:
My one lasting concern is this bit from the syscall docs...

> In order to be sure  that  setsid()  will  succeed,  call fork(2) and have the parent _exit(2), while the child (which by definition can't be a process group leader) calls setsid().

Go does the fork as part of the command call, but the parent (consul-template) obviously doesn't exit. I'm afraid this might cause some issues for some people as I don't want to assume to much about the underlying system calling consul-template. Setting the process group is clearer on what it does and has stricter semantics across platforms.

So while I think it would *probably* be OK, I'm hesitant  because this adds 1 more thing in an already big change.

As you are using the child/ sub-package as a library in this case, what do you think about changing this PR closer to what it was with a flag to trigger setsid usage, but otherwise default to setpgid? This way the setsid option would be available but consul-template  we'd keep using setpgid. This would satisfy your case while keeping the consul-template's behavior slightly clearer (ie. address my concern).

sudhabindu1:
I have made the following changes:
1. `Setsid` can be set from `NewInput` which defaults to `false`.
2. `setpgid` of child is set to `!i.Setsid`
3. `setSetpgid` method is refactored to `setSysProcAttr` so both `setpgid` and `setsid` can be configured.

These changes keep the default behaviors around PGID same while providing an option to set SID, let me know if that fulfills the changes you mentioned above.

eikenb:
Hey @sudhabindu1, sorry for the long silence. Got pulled different ways... but had a chance to review this and it looks good. 

Thanks for all your hard work on this!

eikenb (approved):
:+1:

sudhabindu1:
Thank you @eikenb

PR #1505, opened 2021-08-11
writeToFile func patch

wavedocs:
This code that will function to create nested directories if they are missing.


mbillow:
I included this functionality, including the necessary test coverage, in #1549.

eikenb:
Thanks for contributing this, but PR #1549 included this functionality along with other features and it's been merged. Sorry we didn't use your work, but the functionality is now there.

PR #1511, opened 2021-09-06
fix(doc): removed broken links

erkanzileli:
Greetings.

I just deleted broken links. You can close this PR if you fix these links.


eikenb:
Thanks!

Generating these was automated before we re-orged the docs. Need to look into doing that again.

eikenb:
Hmm... double checked those and the "Running and Process Lifecycle" was correctly removed... but the "Debugging" section seems to have been removed by mistake during the reorg.

So I'm going to merge this to get things correct for what's there now but will also look into adding back the debugging sections. Thanks for bringing that to my attention!

eikenb (approved):

PR #1514, opened 2021-09-07
Disable reload signal optionally

maxixcom:
Sometimes we need to have managed process by consul-template, but do not want to reload it or to send any signals if consul data changes. But at the same time in our case we need to keep actual configs according to template.  Look at the example:

```
template {
  source = "/var/www/html/app/config/config.yml.tpl"
  destination = "/var/www/html/app/config/config.yml"
}

template {
  source = "/var/www/html/app/config/dav.yml.tpl"
  destination = "/var/www/html/app/config/dav.yml"
}

exec {
  command = "php-fpm"
  splay = "5s"
}
```

As you can see we have `php-fpm` and some php application with configs managed by consul kv. Take a notice that we do not have template  for `php-fpm` config, so we do not need to restart or reload it. 

So we have added  new option,  that disables reload spawned process feature:

```
exec {
  command = "php-fpm"
  splay = "5s"
  disable_reload_signal = true // available with this PR
}
```


eikenb:
Hey @maxixcom, thanks for submitting this.

I've been putting this off as I've been internally debating if it should be a new flag/setting or whether it should just get added to the current `reload_signal` setting. Finally decided I'm leaning more toward the latter as it already overloaded the empty case `""` to mean restarting the process so it doesn't seem to big a stretch. I was thinking of maybe something like like "NOOP" or "NOSIG". What do you think of this idea?

kirooshu:
Is this now implemented with the new `reload_signal = ""` which means it will not try to send signals on reload?

PR #1515, opened 2021-09-16
Support for accepting a custom logger for a child process

darora:
Maintains back-compat with existing Child/Input structs by accepting an optional logger object. If one is not provided, we fall back to the current logging scheme, thereby preserving compat with the current output. Tests have been added to ensure this.

I'm using consul-template via [remco](https://github.com/HeavyHorst/remco), which utilizes customized logger objects to differentiate log messages per child process, a feature I quite like. Remco's been using a fork of this project that adds this functionality; I'd love to see it move back to using upstream, if possible.


eikenb:
Hey @darora, thanks for submitting this PR!

While the idea here is sound, I'm not a fan of adding logrus as a dependency. We're going to be converting this project to use https://github.com/hashicorp/go-hclog which would mean 2 logging facilities.

You could use go-hclog instead for this OR you could change this PR to accept an interface for the custom logger that could accept either logrus or go-hclog. One or both might require some wrapping to make it work with the general interface, but IMO that is fine for the ability.

What do you think?

darora:
That sounds reasonable. I'll take a look at the go-hclog project sometime in the next couple of days and change this PR as necessary? Let me know if you have a strong preference for one approach over the other.

darora:
@eikenb just getting back to this now; I've updated the interface to accept stdlib's log.Logger interface. This allows us to easily accept a go-hclog instance (via the hclog.Logger.StandardLogger call), and users of other libraries can easily create a shim if needed.

eikenb:
Hey @darora, sorry for the delay on feedback.

I think I'd more like to see the passed in logger as a replacement for the default instead of in-addition-to. As the passed in logger is just using the built-in Logger interface I'd suggest replacing the `log.Printf()` calls with `c.log.Printf()`... just assign the passed in logger to child as log then use that. If a custom logger isn't passed in (=nil), then use assign the c.log to `log.Default()` which will preserve backwards compatibility.

Make sense?

eikenb:
Changes look good! Thanks @darora!

eikenb (approved):

PR #1516, opened 2021-10-06
Add support for providing Consul ACL Token via a file

### LINKED ISSUE #1429: Add ability to load consul token from a file
There is currently a feature to load a Vault token from a file with the vault_agent_token_file option. This is super handy as it allows consul-template to read the Vault token outputted from a Vault Agent. This file is periodically checked for updates as well which is useful.

There is a similar use-case for loading the Consul token from a file. When running a Vault agent, it is possible to read out a Consul token from Vault and output it into a file using the Vault Agent template option. Unfortunately Consul currently cannot simply load a Consul token so you would have to either template a file that somehow sets the CONSUL_HTTP_TOKEN environment variable and trigger a restart of consul-template, or template out the entire consul-template config file and trigger a reload of consul-template whenever this is changed.

Simply having a "token_file" or similar directive under the consul config block that behaves the same way as the existing vault_agent_token_file vault configuration directive would solve this problem nicely.

lawliet89:
Fixes #1429

Tested it out manually and it works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ this functionality might prove useful.
 - [Plugins](docs/plugins.md)
 - [Caveats](#caveats)
 - [Docker Image Use](#docker-image-use)
-- [Dots in Service Names](#dots-in-service-names)  
+- [Dots in Service Names](#dots-in-service-names)
 - [Termination on Error](#termination-on-error)
 - [Commands](#commands)
   - [Environment](#environment)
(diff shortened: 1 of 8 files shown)


lawliet89:
Hey @eikenb any chance this could be looked at please? I think it'll be quite useful for a few users.

eikenb (approved):
Everything looks good. I'll slot this one in for the next release which should be early next year.

Issue #1497, opened 2021-08-03
Allow setting user/group ownership of template output


drawks:
As asked for in #639 it would great if we could define the user/group ownership of output files. Previously it has been stated that a solution that is completely platform independent would be welcome, however since that comment was made #1185 was merged which preserves user/group only on "unix OSs." It seems like perhaps there is some flexibility on the need for all features to be fully platform independent and if that is the case I'd argue that being able to set owner/group/mode (perhaps even extended attributes and se contexts) on supported platforms would make this tool much more powerful.

deblasis:
Hi there, I came here after some triaging on https://github.com/hashicorp/nomad/issues/5020 which seems to be related.

I agree that being able to set the ownership of the rendered templates can be powerful and sometimes a must-have.
I was wondering, wouldn't a `template.Exec` with a `chown` command be enough in order to achieve this?

deblasis:
I wanted to be my own devil's advocate so I have also PRd #1531, looking forward to hearing back from you guys!

Cheers

eikenb:
>  I was wondering, wouldn't a `template.Exec` with a `chown` command be enough in order to achieve this?

If the consul-template process is run with permissions to do this, then yes that could work. All commands are run with the permissions of the process (and their user/group owner). 

I Read through https://github.com/hashicorp/nomad/issues/5020 and as Nomad doesn't support the Exec (and I don't think they'd want to) it is of no help there. On the other hand, your PR would address their case nicely and I'll take some time to review it.

deblasis:
Awesome! Thank you @eikenb!

eikenb:
Fixed by #1531

PR #1185, opened 2019-02-16
Preserve destination file ownership on unix OSes

kyhavlov:
This PR changes the renderer to preserve the destination file permissions when possible (on operating systems that support Go's FileInfo.Sys() interface).

Fixes #1061.

Issue #639, opened 2016-05-18
Support to define the owner and the group of the destination file


luisbosque:
Currently consul-template supports setting the permissions of the destination file.
It would be great to be able to set the owner and group of it too. This would allow great flexibility without the need to run consul-template as the user that will be owner of the destination file.

Is this not implemented because any special reason or just because nobody needed in the past? I can try to send a PR with owner/group support if it's an acceptable feature.

luisbosque:
Ok, I've found this issue where this feature is discarded:
https://github.com/hashicorp/consul-template/issues/461

sethvargo:
@luisbosque 

Thank you for opening an issue. The problem is that there's really no cross-platform way to do this, and creating the file in advance will preserve the existing permissions so there's a good workaround. Keep in mind we support Windows, bsd, and more, so we need a fully-bc thing. If you can come up with a good PR for that, I'd be happy to review! Thanks! 😄

njhartwell:
@sethvargo It looks like the current implementation does not preserve or copy the user/group attributes from an existing destination file. :+1: for built-in support for this (at least on *nix, if nothing else).

ryancurrah:
Yes I was hoping it would preserve user/group when copying a file but it does not. Also running a command chown then restart doesnt work either it runs chown but fails to restart.

PR #1531, opened 2021-11-03
Allow setting user/group ownership of template output

deblasis:
Adds support for the `uid` and `gid` entries in the `template` stanza that would ultimately be passed to a `os.Chown` command.

It obviously works only on POSIX and Windows is handled with graceful noop with a warning if the user set `uid` or `gid`

Related issues #639, #1497, #1185 and also nomad/5020

Signed-off-by: Alessandro De Blasis <alex@deblasis.net>

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -1026,6 +1026,48 @@ func TestParse(t *testing.T) {
 			false,
 		},
 		{
+			"template_uid",
+			`template {
+				uid = 1000
+			}`,
+			&Config{
+				Templates: &TemplateConfigs{
+					&TemplateConfig{
+						Uid: Int(1000),
+					},
+				},
+			},
+			false,
+		},
+		{
+			"template_gid",
+			`template {
+				gid = 1000
+			}`,
+			&Config{
+				Templates: &TemplateConfigs{
+					&TemplateConfig{
+						Gid: Int(1000),
+					},
+				},
+			},
+			false,
+		},
+		{
+			"template_uid_gid_default",
+			`template {
+			}`,
+			&Config{
+				Templates: &TemplateConfigs{
+					&TemplateConfig{
+						Uid: nil,
+						Gid: nil,
+					},
+				},
+			},
+			false,
+		},
+		{
 			"template_source",
 			`template {
 				source = "source"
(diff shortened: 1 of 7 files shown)


eikenb:
Looks like a problem with the Windows build and the CI system doesn't like the Chown call, might be a mount time restriction on the /tmp directory. Maybe try using a local directory for it instead?

deblasis:
Hi @eikenb, I covered more cases and changed things a bit. It should work now. 
I didn't switch to a local directory because I see that /tmp is used extensively without issues and I think I found the bug elsewhere. 
Let's see if Miss CI is happy about it 😄

lukas-w:
I also did some work on this issue recently but didn't find the time to open a PR yet unfortunately (no tests yet). My implementation supports setting a file owner/group and permission bits (converts octal permissions to ACL entries) on Windows as well. See https://github.com/hashicorp/consul-template/compare/master...lukas-w:f-file-owner. Maybe we can join efforts on this?

3nprob:
@eikenb Did you get to take another look at this after the CI issues were addressed? For being able to use separate uids for different tasks, this is crucial.

While it would be sweet to get Windows support as per @lukas-w , isn't *nix/mac independently worthwhile enough to push through?

eikenb:
Thanks for the prompt @3nprob, I'll take another look at this when I start work on 0.28.0 (should be soon, next in my project queue).

eikenb (approved):
Everything looks good. Thanks!

lukas-w:
> While it would be sweet to get Windows support as per @lukas-w , isn't *nix/mac independently worthwhile enough to push through?

Sure, It's great to see this merged! Another difference in the two implementations though is that mine supports specifying `user` and `group` entries as string so you can pass uid/gid as well as user/group names which I find more useful personally because it allows for more readable configurations and facilitates Windows support (even if it's not implemented now) as Windows doesn't use integer identifiers. I can prepare a PR for this soon if that's something that's desired as to avoid a future interface change in Nomad & consul-template.

eikenb:
Hey @lukas-w,

If you'd like to open a PR I'll be happy to review it.

PR #1534, opened 2021-11-12
Logging to file

### LINKED ISSUE #1416: Log-to-file / logging on Windows with SC.exe
### Consul Template version

v0.25.1

### Expected behavior

Consul-template supports logging to file. Consul-template can be started as a service on Windows, and the resultant messages can be inspected. Consul offers `-log-file` via stdarg and Nomad offers `log_file` in the config.

### Actual behavior

Consul-template only supports Syslog and stdout/stderr logging mechanisms. Windows discards stdout/stderr messages from services. Seemingly no way to get persistent logging without wrapping consul-template.exe in a script that captures stdout and stderr to a file. 

```
Calling: ArraySeq(sc.exe, create, consul-template, binpath=, C:\opt\consul-template\consul-template.exe)
Calling: ArraySeq(sc.exe, config, consul-template, DisplayName=, Consul-Templat, binPath=, C:\opt\consul-template\consul-template.exe -config=C:\opt\consul-template\config-windows.hcl -vault-token=s.REDACTED -consul-token=REDACTED)
Calling: WrappedArray(sc.exe, start, consul-template)
```

### Steps to reproduce

1. Create and start an instance of consul-template as a service with sc.exe on Windows
2. Attempt to inspect behavior. 
3. Unable to inspect behavior - no ability to log messages to a file.

### References

Are there any other GitHub issues (open or closed) that should
be linked here? For example:
- #1382

deblasis:
Hi, I spotted this issue and it sounded like an easy one.
Most of the credits go to the Consul folks because while copying the semantics I ended up taking most of the logic too. 
I just added a few tweaks to match consul-template semantics and structure.

This PR adds:
- Consul-like CLI flags used to control the behaviour of the feature
- `log_file` configuration stanza
- updated docs for CLI and config

Fixes #1416

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -194,6 +194,31 @@ syslog {
   # This is the name of the syslog facility to log to.
   facility = "LOCAL5"
 }
+
+# This block defines the configuration for logging to file
+log_file {
+  # If a path is specified, the feature is enabled
+  # Please refer to the documentation for the -log-file
+  # CLI flag for more information about its behaviour
+  path = "/var/log/something.log"
+
+  # This allow you to control the number of bytes that
+  # should be written to a log before it needs to be
+  # rotated. Unless specified, there is no limit to the
+  # number of bytes that can be written to a log file
+  log_rotate_bytes = 1024000
+
+  # This lets you control time based rotation, by default
+  # logs are rotated every 24h
+  log_rotate_duration = "3h"
+
+  # This lets you control the maximum number of older log
+  # file archives to keep. Defaults to 0 (no files are ever
+  # deleted).
+  # Set to -1 to discard old log files when a new one is
+  # created
+  log_rotate_max_files = 10
+}
 ```
 
 ## Consul
@@ -347,7 +372,7 @@ vault {
   # documentation for more information.
   unwrap_token = true
 
-  # The default lease duration Consul Template will use on a Vault secret that 
+  # The default lease duration Consul Template will use on a Vault secret that
   # does not have a lease duration. This is used to calculate the sleep duration
   # for rechecking a Vault secret value. This field is optional and will default to
   # 5 minutes.
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ this functionality might prove useful.
 - [Math Functions](docs/templating-language.md#math-functions)
 - [Observability](docs/observability.md)
 - [Logging](docs/observability.md#logging)
+  - [Logging to file](docs/observability.md#logging-to-file)
 - [Modes](docs/modes.md)
 - [Once Mode](docs/modes.md#once-mode)
 - [De-Duplication Mode](docs/modes.md#de-duplication-mode)
@@ -44,7 +45,7 @@ this functionality might prove useful.
 - [Plugins](docs/plugins.md)
 - [Caveats](#caveats)
 - [Docker Image Use](#docker-image-use)
-- [Dots in Service Names](#dots-in-service-names)  
+- [Dots in Service Names](#dots-in-service-names)
 - [Termination on Error](#termination-on-error)
 - [Commands](#commands)
   - [Environment](#environment)
@@ -235,7 +236,7 @@ users the ability to further customize their command script.
 #### Multiple Commands
 
 The command configured for running on template rendering must take one of two
-forms. 
+forms.
 
 The first is as a single command without spaces in its name and no arguments.
 This form of command will be called directly by consul-template and is good for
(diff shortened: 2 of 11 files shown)


eikenb:
Hey @deblasis, thanks for the writing this!

After a quick review, I'd say this looks very good. I enabled tests and one is failing, I'll look into why that one test failed on the CI tomorrow.

deblasis:
Hi @eikenb! Thank you!

I had a look at the test run log and then I noticed that I was using `github.com/hashicorp/consul/sdk/testutil.TempDir` instead of `io/ioutil.TempDir`. 
This last commit should the permissions error and also now tmp folder creation is handled like in the other tests that require it.

eikenb:
Thanks for the fix!

Looked it over in a bit more detail and everything looks good. Going to tag this to be included in the next feature release. :rocket:

fredwangwang:
Thanks @deblasis for working on the issue and @eikenb for reviewing! We would love to start using this feature, is there a timeline for when the next release will drop?

eikenb:
Hey @fredwangwang, good to know of the interest.

So I've been playing around with adding [project roadmaps](https://github.com/hashicorp/consul-template/projects/4) to my projects, though I'm still playing with the format. It has 2 other features, expanded exec/command config and multiple namespaces support. The latter requires I finish some other work in the [hashicat library](https://github.com/hashicorp/hcat) I'm working on to replace consul-template's use as a library (and which consul-template itself will of course be using). I'm currently working on it (it has a [roadmap](https://github.com/hashicorp/hcat/projects/1) too) and plan to work on consul-template again once v0.2.0 is done.

So I don't have a date, but that should help give you an idea. Realistically given the work needed and the time of year, it probably won't be until after the new year. I could possibly merge it before that time if people want to try it and are comfortable building it themselves.

fredwangwang:
Thanks for providing the context! 
> merge it before that time if people want to try it and are comfortable building it themselves.

That would be great if possible :)

docs/configuration.md
```diff
@@ -194,6 +194,31 @@ syslog {
...
+
+  # This lets you control time based rotation, by default
+  # logs are rotated every 24h
+  log_rotate_duration = "3h"
+
+  # This lets you control the maximum number of older log
+  # file archives to keep. Defaults to 0 (no files are ever
+  # deleted).
+  # Set to -1 to discard old log files when a new one is
+  # created
```
eikenb, line 219: I'm inclined to have 0 mean discard old logs (keep 0 logs) and -1 mean keep all logs. Would you mind flipping them or, alternatively, expanding a bit on why you went the way you did?

deblasis: I went this way following the "Tao" for consistency since it looks like it's the way Consul handles this:

image

https://www.consul.io/docs/agent/options#_log_rotate_max_files

deblasis: Apart from consistency, I think I see the reason behind your inclination but the default `0 => never delete` seems saner to me because ultimately logs can be valuable, especially if the user turned on this feature which is off by default.

`-log-file` doesn't exactly scream "log today's stuff to file and forget about the past"

For the same reason `-1 => discard old logs` from a UX perspective sounds good to me because it has to be a user's deliberate choice, probably relevant after they realised that they don't need to keep more than 1 file. 
Useful when logs come in handy for troubleshooting a process that dies suddenly. In these cases you would normally look at the most recent log, yesterday's log is probably not relevant.

For everything else, like for example when file logs are ingested by other tools (that would probably clean after themselves and use the filesystem like a queue), I feel like it's best to have a default that is not destructive. 

I am sure I am missing something but these are my 2cents :)

Happy to hear your thoughts and to change as you see fit.

eikenb: You had me at consistency.

As for the default being safer, I don't think that cuts both ways. During my time as an SRE we always problems with logs filling up the disk on the servers. I can't remember how many log rotation setups I had to tweak. But no matter, I think it is a good idea keeping it consistent with Consul so  no changes needed. Thanks for the engagement!

eikenb: > As for the default being safer, I don't think that cuts both ways.

Not "don't"... should have read "[..] I think that cuts both ways."

deblasis: > always problems with logs filling up the disk on the servers

I hear you, I think that the persistent memory industry is very grateful for all these undeleted logs! :)

> I think it is a good idea keeping it consistent with Consul so no changes needed. Thanks for the engagement!

Awesome, you are very welcome. Thank you too!

PR #1541, opened 2022-02-05
Allow setting ownership by name

lukas-w:
As proposed in https://github.com/hashicorp/consul-template/pull/1531#issuecomment-1023156588, this changes the new file ownership specification from integer `uid`/`gid` to string `user`/`group` that can either be a decimal id or a user/group name. This is both for convenience and for future compatibility with Windows. This change allows us to implement Windows support in the future without changing the public interface in consul-template and Nomad. One catch in the context of Nomad is that user/group names would be looked up on the host, not e.g. inside the Docker container as users may expect.

I also fixed a potential nil dereference via af770eccf6ec1fd2b6722621cea4af50042adb1e and made `Render(...)` fail when trying to set file ownership on Windows analogous to how it fails if `Chmod` fails via [`5be1f70`](https://github.com/hashicorp/consul-template/pull/1541/commits/5be1f708cd67dc7136e8534b924cb67cd244cd6f).

Tagging @eikenb because they kindly offered to review.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -1096,12 +1096,12 @@ func TestParse(t *testing.T) {
 		{
 			"template_uid",
 			`template {
-				uid = 1000
+				user = "1000"
 			}`,
 			&Config{
 				Templates: &TemplateConfigs{
 					&TemplateConfig{
-						Uid: Int(1000),
+						User: String("1000"),
 					},
 				},
 			},
@@ -1110,12 +1110,12 @@ func TestParse(t *testing.T) {
 		{
 			"template_gid",
 			`template {
-				gid = 1000
+				group = "1000"
 			}`,
 			&Config{
 				Templates: &TemplateConfigs{
 					&TemplateConfig{
-						Gid: Int(1000),
+						Group: String("1000"),
 					},
 				},
 			},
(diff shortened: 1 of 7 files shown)


eikenb:
Hey @lukas-w, nice PR.

I've looked it over and everything looks good. I want to pull it down and check it over locally to be sure, but I think I'll be able to merge it as is and it should make it into 0.28.0. Thanks!

igor-nikiforov:
@eikenb I'm very sorry for ping you but there is any ETA when this PR will be merged?

eikenb:
Sorry, 0.28.0 went out a bit early due to a customer need which bumped this to 0.29 and distracted me for a bit. I was able to pull it down and look it over though and it still looks good. In other words the answer is... 

Thanks for reminding me, I'll merge it in a few minutes.

eikenb:
Just realized an issue. This changes the behavior in a backwards incompatible way. It wasn't an issue before as it was going in the same release as the original change and so there was no release. Sigh.

I think I'm just going to keep it as a backwards incompatible change (mentioning it in the changelog) as that was the intention before the early release.

@deblasis, if you happen to see this and wouldn't mind, would you chime in with your thoughts on this. I'm going to merge it but could add backwards compatibility to it after the fact if there is a need. Thanks.

deblasis:
Hi @eikenb !

Yeah, I think we still have to support `uid` and `gid` settings while keeping `user` and `group` of course.

I found a couple of minutes and I put together #1551 

Open to suggestions/improvements if needed as always

eikenb:
Removed the breaking-change label as that was mitigated by #1551

Issue #1530, opened 2021-10-28
Document Exec{} block as part of template configuration


eikenb:
Configuration file template block needs updating. It still lists the `command` options and doesn't have the `Exec` block in the documentation at all. It does have a correct `exec{}` block in the example though.

- [ ] mark `command` as deprecated
- [ ] add `exec` block to template section

marco-m-pix4d:
If I understand correctly, also the README needs updating

eikenb:
Hey @marco-m-pix4d,

We split the README documentation up into smaller chunks a while back and, AFAIK, all the needed changes are in the docs that were split out (in docs/configuration.md). The documentation still in the top level README.md, in the `Commands` section, applies equally to the Exec.Command setting. So I think it is OK for now.

marco-m-pix4d:
You are indeed right, thanks for the quick answer @eikenb :-)

Issue #1518, opened 2021-10-07
Support command/exec lists


eikenb:
Currently configured commands/execs can take 2 forms. A single word/command (no spaces) that will get run directly on any supported OS and a full shell command line (with spaces) that is supported on *nix systems with an standard shell (/bin/sh) available. This limits the ability of people who *do not* have the shell available to only running single word commands which might be tough in some setups (might require a small binary wrapper).

To better support multiple command execution environments add support for direct execution lists. That is the command in a list form, with the list elements separated on the spaces of the command. These will be parsed into a Go string slice which can be directly passed to the underlying OS execution system call. This would then allow all commands, even those with arguments, to be directly run without the need for the wrapping shell call.

Some examples of what this will probably look like...

In HCL:
```hcl
command = ["/usr/local/bin/awesome-app", "server"]
```

In JSON:
```json
"command": ["/usr/local/bin/awesome-app", "server"]
```

PR #1543, opened 2022-02-09
support command lists

eikenb:
This updates the exec{} block's `command` setting to support command lists.
That is a command formatted like `["echo", "hello", "world"]`.
It is backwards compatible with the old format, turning single commands
w/o spaces into lists.
Single commands with spaces still use the `sh -c` setup.

This will allow for multi-word commands on Windows and in setups (like
bare docker containers) without `sh` installed.

Fixes: #1518 
Fixes: #1530 (doc fix I did at same time)


mkam (approved):
Looks good to me!

PR #1545, opened 2022-02-25
add -parse-only option

jrwren:
Allows checking of templates for well formedness without rendering.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -422,6 +422,11 @@ func (cli *CLI) ParseFlags(args []string) (
 		return nil
 	}), "once", "")
 
+	flags.Var((funcBoolVar)(func(b bool) error {
+		c.ParseOnly = *(config.Bool(b))
+		return nil
+	}), "parse-only", "")
+
 	flags.Var((funcVar)(func(s string) error {
 		c.PidFile = config.String(s)
 		return nil
@@ -775,6 +780,9 @@ Options:
 
   -once
       Do not run the process as a daemon. This disables wait/quiescence timers.
+
+  -parse-only
+      Do not process templates. Parse them for structure.
 
   -pid-file=<path>
       Path on disk to write the PID of the process
--- a/cli_test.go
+++ b/cli_test.go
@@ -758,6 +758,14 @@ func TestCLI_ParseFlags(t *testing.T) {
 			},
 			false,
 		},
+		{
+			"parse-only",
+			[]string{"-parse-only"},
+			&config.Config{
+				ParseOnly: true,
+			},
+			false,
+		},
 	}
 
 	for i, tc := range cases {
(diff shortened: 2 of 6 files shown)


eikenb:
Hey @jrwren, thanks for the PR!

It looks like you based this around the Once flag which was perfect as this would be anther command line only option. It is missing a few things though, mainly tests. There should be tests for the CLI and config which would follow the same model as Once still. The part from runner would be a bit more involved, but not to bad I think. It would need to test running Start() and the appropriate functionality. There is a once test in there as well and it might not make to bad of a base case. Please feel free to ask if you have any questions. Thanks again for submitting this.

jrwren:
I took a stab at writing some tests, but the tests on master don't run at all for me on my mac.
```sh
$ go test -vv .
exit status 1
FAIL	github.com/hashicorp/consul-template	0.265s
```

No error messages or text output, just `exit status 1`.

So I added what I consider to be a pointless test to cli_test and hope CI passes.

eikenb:
Hey @jrwren, 

I enabled CI tests and things look good there.

To test try `go test -v ./...` and that should run all the tests.

jrwren:
When I run `go test -v ./...` I get many errors, seemingly unrelated to the code I wrote, but now that I look a little deeper, I see it is only `TestVaultConfig_Finalize`.

Can you suggest more areas to test or is this ready?

eikenb (commented):
These are pretty pointless, but they are there and, until possibly removed, I prefer to keep them complete. Thanks.

eikenb:
The tests all pass for me locally and in CI so I'm not sure what issue you're seeing. If you want to include the output of the failing test maybe I can help.

eikenb:
Hey @jrwren, thanks for the continued work on this PR!

Things are looking pretty good with only the final test of the additional code in manager/runner.go. I'll be happy to help out if needed, just ask any questions here.

jrwren:
I added a test to cover the changes in runner.go.

Thank goodness for CircleCI, because I still can't run these on my mac. I don't get the output of failed tests, even with `go test -v`, but this must be something my fault on this end.

eikenb:
Thanks for the test @jrwren, that seems to cover the tests well enough.

I have one "last" thought that occurred to me while looking it over to refresh my memory. That is that it is possible that the commands (template and exec) could be run on that first pass of Run if there are no dependencies. That is the first call to Run usually just registers all the dependencies and starts the async process of looking them up and would work fine for this but if you have no dependencies then it would just be finished and run the commands.

Though IMO this is a small issue (why use consul-templates if you have no dependencies to look up) and not worthy of changing anything. I mainly wanted to mention it here for historical reasons in case it comes up later.

TLDR; LGTM

Thanks for your work on this!

eikenb (approved):

Issue #1526, opened 2021-10-20
[feature request] Add an ability to specify the renewal time for the secrets 


Oloremo:
## Description

As per [documentation](https://github.com/hashicorp/consul-template/blob/master/docs/templating-language.md#secret):
> The Renew API tries to use most of the time the secret is good, renewing at around 90% of the lease time (as set by Vault).

In some cases, we really want to control the time for renewal.

For example here is how it looks like for the TLS certificates that consul-template populate from the Vault:

![image](https://user-images.githubusercontent.com/4280480/138162584-00ff2974-059c-444d-ab24-eba92253a1b3.png)

It's very hard to monitor. We'd really like to set the threshold to like half of the TTL and alert if half of the TTL + 1 hour has passed but certificates are not renewed.

chrisvanmeer:
I would also like this feature. I'm currently having static secrets in a K/V 2 engine and now it takes around 5 minutes for Consul to pickup this change and signal it to Nomad. K/V 1 allowed for a `ttl` k/v pair to specify the ttl for the specific secret. It would be great if either that `ttl` k/v option would be available in K/V 2.

Oloremo:
Was implemented.

eikenb:
Thanks @Oloremo

PR #1546, opened 2022-03-01
configurable lease check wait for non-renewable secrets

rgeoghegan:
This should be a fix for https://github.com/hashicorp/consul-template/issues/1526

Now the vault config has a `lease_renewal_threshold` parameter which controls the
fraction of how long of the original lease duration consul-template should wait to
ask for a new secret on non-renewable secrets (like for PKIs). By default, consul-template
will wait (90 +/- 5)% of the lease time.


config/vault.go
```diff
@@ -310,8 +328,9 @@ func (c *VaultConfig) GoString() string {
 		"Token:%t, "+
 		"VaultAgentTokenFile:%t, "+
 		"Transport:%#v, "+
-		"UnwrapToken:%s"+
+		"UnwrapToken:%s, "+
```
eikenb, line 331: Drive-by formatting fix. Thanks!

eikenb (approved):
Seems like you've covered all the bases.
LGTM.

config/vault.go
rgeoghegan: I was trying to figure out why the unit tests would not pass, and I noticed this typo when my print statements were all weird.

Issue #1259, opened 2019-08-15
All PKI certs are regenerated on reload signal


cagriy:
consul-template re-requests all vault PKI backend based certificates upon receiving reload signal.

**How to recreate:**
1 - Bring up a vault server (v1.2.1 used) instance using -dev option
2 - Create a test PKI backend as described in vault documentation (https://www.vaultproject.io/docs/secrets/pki/index.html) add generate_lease=true while creating PKI role
3 - Create following template file
```
{{ with secret "pki/issue/example-dot-com" "common_name=foo.example.com" }}
{{ .Data.certificate }}{{ end }}
```
4 - Run consul-template (v1.20.0 used for testing) using
```
consul-template -template "./temp.ctmpl:/tmp/cert.crt"
```
5 - Note md5sum of /tmp/cert.crt and consul-template PID number
6 - Execute kill -HUP consul_template_PID
7 - Re-check md5sum of /tmp/cert.crt

**Background:**
This behaviour becomes an issue while using short lived consul tokens created by vault consul backend. If consul token is renewed frequently, this results in consul_template reloading itself frequently. In turn certs keep getting renewed and applications using these certs (such as nginx) will also require reloading.

eikenb:
Hey @cagriy, thanks for the report.

I've been digging into PKI a bit due to a regression and remembered this issue (which I put off as I hadn't learned about vault's PKI backend yet).

In your description of how to reproduce it you send a HUP to consul template. Sending a HUP reloads all configuration and templates by design. It is in the documentation under [Running and Process Lifecycle](https://github.com/hashicorp/consul-template#running-and-process-lifecycle). So this seems to be the correct behavior in this case.

However in your background section you describe it differently, saying... "If consul token is renewed frequently, this results in consul_template reloading itself frequently." Is there another way to reproduce this that would more closely mimic this?

cagriy:
Hi @eikenb , thanks for your answer.

In our setup, there is a requirement to have short living consul and vault tokens. Hence a process renews them every hour and reloads consul-template to start using these new tokens. 

I understand a reload naturally results in reloading all configuration.

Our challenge is that, upon config reload, consul-template asks for vault PKI backend for new certs, although existing certs have sufficient lease on them and valid certificates are already stored on the disk at the specified destinations.

[...]

PR #1549, opened 2022-03-07
writeToFile Function Improvements

mbillow:
## User and Group Changes
The `writeToFile` function now allows empty strings to be passed into the `username` and `groupName` fields which will default to the current user. If neither is set, it will skip the `chown` call completely. This allows the consul-template to run as an init container without additional permissions. 

The `username` and `groupName` fields can now also support raw UIDs and GIDs respectively. This allows vault-agent to write files for users that aren't in `/etc/passwd`.

## Directory Creation
Now consul-template will automatically attempt to create the directories necessary before writing the file.


eikenb (approved):
Thanks for including documentation updates!

mbillow:
@eikenb Is it possible to get this released before the current 0.29.0 target? 

I’m trying, with the help of our Enterprise support team, to get this (and sprig templates in 0.28.0) pulled into the next release of Vault because we need all of this in Vault Agent.

eikenb:
Hey @mbillow,

The only outstanding feature/issue for 0.29.0 is also for Vault Agent, the PKI cert renewal issue (https://github.com/hashicorp/consul-template/issues/1259).

mbillow:
@eikenb Is that something actively being worked on? Is there an ETA? Just trying to properly set expectations on our side since this blocks some work internally.

eikenb:
Yep. I'm working on that next. I've already figured out the solution, but need to write it up w/ tests, etc. There were some other bugs I was going to fix for 0.29 as well, but there are some internal discussions here that might end up with a quicker release.

mbillow:
Apologies if it is coming from the enterprise support team. I’ve been discussing with them the importance this has for us and letting them know I made the changes I did. 

There is a lot of old, custom stuff we can turn off when these changes get merged into Vault that I am _very_ excited to turn off. 😅 

Let me know if there is anything I can do to help! Happy to benefit the open source community while also on the clock. 😉

eikenb:
No apologies necessary @mbillow. The other teams work is important (to us) to help set priorities and is part of the process. Thanks for the thoughts though and am very happy to hear that this will make your life easier.

PR #1551, opened 2022-03-16
Backward compatibility for chown, broken in 0.28.0 -> 0.29.0

deblasis:
This PR simply takes into account `uid` and `gid` settings in the template, introduced in 0.28.0 but replaced by `user` and `group` in 0.29.0, breaking compatibility.

We address the issue by giving precedence to the newer settings but setting the relevant properties upon type conversion if the legacy settings are provided

Uid (0.28.0) -> User (0.29.0)
Gid (0.28.0) -> Group (0.29.0)

Related #1541

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"os"
 	"reflect"
+	"strconv"
 	"syscall"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


eikenb:
Hey @deblasis, thanks for writing this and you are right in that it would be better to keep compatibility for now. Though I do need come up with a system for deprecating and removing things so I can add this (along with a few other things) to it.

eikenb (approved):
Simple change with test coverage. :+1:

deblasis:
> Hey @deblasis, thanks for writing this and you are right in that it would be better to keep compatibility for now. Though I do need come up with a system for deprecating and removing things so I can add this (along with a few other things) to it.

No worries! I am assuming that 1.0 is gonna be the version with these deprecations. I noticed that there is at least another instance of legacy stuff that is still supported but that could be deprecated. Something around "connection timeout".

In the past I have used a simple markdown file named like the next major version where I listed all the legacy things that had to be deprecated. I still remember the folder name: `sunsetting` 🤭

It was descriptive text for documentation but it contained an items list with checkboxes that was parsed by a simple github action that would fail if you tried to merge a branch with a new major tag when the markdown file matching that version contained items left unchecked. 

Not really a fancy system, quite manual and simple but it helped.
I am sure there are better ways, I am curious to find out what you come up with 😊

PR #1556, opened 2022-03-25
eliminates confusing use of a slice

eikenb:
The runner keeps a map of template-ids to template-configs for easy lookup in various places. The template-configs was a slice, though it could only ever have 1 entry. I'm not sure why the slice is there, maybe the original author had some idea that never came to fruition or maybe they just liked iterating... not sure but it I found it very confusing as IMO using a list/slice heavily implies that there could be more than one thing in it.

This eliminates that needless slice so that code quits confusing me.

### DIFF
--- a/manager/runner.go
+++ b/manager/runner.go
@@ -52,7 +52,7 @@ type Runner struct {
 
 	// ctemplatesMap is a map of each template ID to the TemplateConfigs
 	// that made it.
-	ctemplatesMap map[string]config.TemplateConfigs
+	ctemplatesMap map[string]*config.TemplateConfig
 
 	// templates is the list of calculated templates.
 	templates []*template.Template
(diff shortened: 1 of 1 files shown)


manager/runner.go
```diff
@@ -1021,23 +1016,19 @@ func (r *Runner) diffAndUpdateDeps(depsMap map[string]dep.Dependency) {
 }
 
 // TemplateConfigFor returns the TemplateConfig for the given Template
-func (r *Runner) templateConfigsFor(tmpl *template.Template) []*config.TemplateConfig {
+func (r *Runner) templateConfigsFor(tmpl *template.Template) *config.TemplateConfig {
```
mkam, line 1019: Would it help to change this to `templateConfigFor` now that the map returns only one entry?

mkam (approved):
Looks good to me!

manager/runner.go
eikenb: Good idea... I'll update that now. Thanks!

PR #1558, opened 2022-03-29
Added trim functions

### LINKED ISSUE #1544: trimprefix need to added in consul-template
i am trying to access tags from service information and perform some operation on tags strings.
```
{{ range $tag, $services := service "test-service" | byTag }}{{ if $tag | contains "contextPath" }}{{ $tag }}{{ end }}{{ end }}
```

if tags have contextPath=/service-info I need to just print service-info in the output I just found https://www.terraform.io/language/functions/trimprefix so if we can use that or any suggestions are welome

jsanant:
Added trim functions based on this requirement #1544 : 
- `trim()`
- `trimPrefix()`
- `trimSuffix()`

Fixes #1544


jsanant:
@eikenb - Can you please review this PR? If there is anything else that needs to be added please let me know.

eikenb (approved):
Thanks!

PR #1562, opened 2022-04-01
Remove writeToFile reliance on cgo/$USER for UID/GID detection

mbillow:
@eikenb One _very_ tiny change I hope we can sneak in before 0.28.1. I didn't realize that using `user.Current()` relies solely on the `USER` environment variable if the binary isn't built with CGo. We should absolutely just be grabbing the UID and GID from the `os` package.


eikenb:
Sounds like a reasonable change given the intentions of the original PR. I'll merge it and cherry pick it to the 0.28.1 branch (just pushed it today as I wanted to merge some 0.29 work).

PR #1563, opened 2022-04-04
fix name/use of confusing Block flag

eikenb:
The Block flag returned in the ResponseMetadata from Fetch was
confusing. It sounds like it should be about blocking query, but it is
really about a query that would normally block but instead is returning
no data/nil. The logic in this case is to just loop around to call Fetch
again instead of returning the nil data.

The only case of this is KV Get. Consul returns a 404 for KV lookups
on invalid paths but the GO API ignores this error and returns nil data
(and nil error) instead. By having it just loop again it is basically
letting you to have a blocking-like call on a key that doesn't exist yet
(but could come into existance).

This change fixes the naming of the flag to be more obvious in its
meaning, adding a few reminder comments as well as cleaning up a bit
of use from the confusion.

### DIFF
--- a/dependency/kv_get.go
+++ b/dependency/kv_get.go
@@ -21,9 +21,9 @@ var (
 type KVGetQuery struct {
 	stopCh chan struct{}
 
-	dc    string
-	key   string
-	block bool
+	dc         string
+	key        string
+	blockOnNil bool
 }
 
 // NewKVGetQuery parses a string into a dependency.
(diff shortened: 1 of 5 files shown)


dependency/kv_get.go
```diff
@@ -57,6 +57,8 @@ func (d *KVGetQuery) Fetch(clients *ClientSet, opts *QueryOptions) (interface{},
 		RawQuery: opts.String(),
 	})
 
+	// NOTE that the Consul HTTP KV API returns a 404 on failed gets, but the
+	// Go API ignores those for KV gets, returning nil data and error.
```
lornasong, line 61: Does the "Go API" here refer to the Consul API SDK or is it something else?

lornasong (approved):
The `BlockOnNil` name is easier to understand! Appreciated the context in the comments

I was wondering if this type of code is only in CT? Is there anything similar in hcat?

dependency/kv_get.go
eikenb: Yeah, I mean the `github.com/hashicorp/consul/api` package. They have a separate `github.com/hashicorp/consul/sdk` package that is kind of using the SDK moniker, so I'm not 100% on what's best to call it. That was my shot. Maybe "Go API Package" would be clearer?

lornasong: Thanks for clarifying, I think that works! My only thought is maybe also adding "Consul" e.g. "Consul Go API Package"? The "something else" I was considering was the go standard library's api package (which didn't seem likely).

eikenb: Comment updated. Thanks @lornasong!

PR #1565, opened 2022-04-06
Add Nomad Service Discovery Support

schmichael:
Adds 2 functions for retrieving services from Nomad:

- nomadServices - retrieves all services from Nomad
- nomadService - retrieves all registrations for a service

Tag filtering for `nomadService` is currently implemented in
consul-template as the Nomad API does not support it directly.

Tests use a Nomad binary on path and require Nomad 1.3.0-beta or a
binary built from the PR: hashicorp/nomad#12458

Tests also require Docker.


eikenb (approved):
Everything looks good. Spent more time reviewing the integration points vs. the new functionality but saw no issue with either. Thanks!

And extra thanks for the mapstructure api test updates. :heart:

PR #1573, opened 2022-04-11
fix bug with 2 templates, same ID, different paths

eikenb:
The way template configs were being tracked meant that if 2 templates
had the same content but different config otherwise, they would only use
the last config. Having more than 1 template with the same content is
rare but a user ran into it with 2 templates and the same path.

This eliminates the separate tracking of template configs and bundles
them with the templates themselves. Having a 1-1 template/config and
eliminating the separate data structure makes this area of code simpler
to understand and work with while fixing this issue.

This also fixes the TemplateConfigMapping() call which should have kept
backwards compatibility with the signature.

### DIFF
--- a/manager/runner.go
+++ b/manager/runner.go
@@ -50,10 +50,6 @@ type Runner struct {
 	outStream, errStream io.Writer
 	inStream             io.Reader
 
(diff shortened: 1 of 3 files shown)


mkam (approved):
Looks good to me!

PR #1578, opened 2022-05-02
sort: use strictly less-than for less implementations

shoenig:
This PR updates implementations of the `sort.Less` interface to use
strictly less-than evalutation of arguments.

https://pkg.go.dev/sort#Interface

	// Less reports whether the element with index i
	// must sort before the element with index j.
	//
	// If both Less(i, j) and Less(j, i) are false,
	// then the elements at index i and j are considered equal.
	// Sort may place equal elements in any order in the final result,
	// while Stable preserves the original input or

I suspect `<=` only causes a problem when used in conjunction with sort.Stable
as the results will not actually be stable. However we do in fact use sort.Stable
in consul-template.

```
➜ rg sort\.Stable
dependency/catalog_node.go
126:	sort.Stable(ByService(services))

dependency/catalog_services.go
86:	sort.Stable(ByName(catalogServices))

dependency/nomad_services.go
108:	sort.Stable(nomadSortableSnippet(services))

dependency/nomad_service.go
120:	sort.Stable(NomadServiceByName(services))

dependency/catalog_nodes.go
98:		sort.Stable(ByNode(nodes))

dependency/health_service.go
204:		sort.Stable(ByNodeThenID(list))
```

### DIFF
--- a/config/nomad.go
+++ b/config/nomad.go
@@ -2,7 +2,7 @@ package config
 
 import "fmt"
 
-// NomadConfig is the configuration for connecto to a Nomad agent.
+// NomadConfig is the configuration for connecting to a Nomad agent.
 type NomadConfig struct {
 	// Address is the URI to the Nomad agent.
 	Address *string `mapstructure:"address"`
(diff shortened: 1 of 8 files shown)


swenson (approved):
LGTM

It is *probably* fine to use `<=` with the standard Go sort routines, but definitely would no longer guarantee stability with `sort.Stable`.

eikenb:
Did you consider using SliceStable() instead?

I would normally take the latter approach in new code as it meets my criteria (need to sort in 1 place, not a sort-able type), but I'm not 100% here as only fixing the existing code is a much smaller change.

shoenig:
Yeah @eikenb I kinda prefer leaving refactoring / bug fixing separate, though it would be a decent change to make

PR #1587, opened 2022-06-08
fix issue with debug printing of VaultConfig

eikenb:
New floating point field got added to the GoString() method wrong,
making the assumption that the holding struct would not be nil (which it
is often in tests). Found this while debugging a test and it panicing
when trying to print out the differences.


lornasong (approved):
looks like there's a minor conflict to resolve. regardless, makes sense to me!

PR #1596, opened 2022-07-07
Updated dependencies to remove some security alerts.

tsaarni:
* Upgraded from github.com/Masterminds/sprig v2.22.0+incompatible to v3.2.1. This brings in [security update](https://github.com/advisories/GHSA-xg2h-wx96-xgxr) to indirect dependency github.com/Masterminds/goutils v1.1.1 which was uplifted already by https://github.com/hashicorp/consul-template/commit/b43344b95c8a0f622dfcd6584fa99cc21e317a70by where the version was pinned. It would be nice to update the direct dependency as well.
* Changed the import path to github.com/Masterminds/sprig/v3 to align with the new version.
* Updated github.com/imdario/mergo and github.com/hashicorp/nomad/api to uplift indirect dependency gopkg.in/yaml.v3 to version v3.0.1. This makes sure CVE-2022-28948 is addressed.


tsaarni:
Cc @eikenb

I saw there is already commit to uplift goutils v1.1.1 but I wonder what would you think about approach suggested in this PR: update the 1st level depenency as well?  I also included two more uplifts to get `gopkg.in/yaml.v3` security fix.

Unfortunately I did not manage to run `make test` on my machine. I believe I had problems with a test with syslog dependency on my machine. If you don't mind, it would be great to run the tests via the workflows.

eikenb:
Hey @tsaarni, thanks for the PR!

I'm sorry to say that sprig v3 is not a drop in replacement for sprig-v2 as it made several backwards incompatible changes to its template functions. We try to avoid breaking changes if possible and for sprig my thinking was that I'd eventually add sprig3 as a separate feature, not just replace sprig 2 with it. At least until v2 was out of support.

Double checked and Sprig-2 is at the current version. Maybe just skip it and update the other indirect dependencies that need it?

tsaarni:
@eikenb OK I see, thank you!  I updated the PR in case you'd still like to take the `gopkg.in/yaml.v3` update.

eikenb (approved):
:+1: Thanks!

PR #1600, opened 2022-07-15
only use setpgid for 'sh -c' wrapped subshell calls

eikenb:
Previously setpgid was set for all commands. This had an unintended
side effect of breaking certain use cases, like using an interactive
'bash' as the command to be executed/managed as a way to debug/test your
setup.

This was also added as part of the conversion from the old (broken)
shell parser to using 'sh -c' subshells for those calls. After that
I added support for single commands and command lists that don't get
wrapped. Using setpgid is not needed on the non-wrapped calls as it was
added to ensure backwards compatibility with signal propagation for
the wrapped commands (the wrapping shell discards some signals).

Note this is an internal change with no documentation changes required.

### DIFF
--- a/child/child.go
+++ b/child/child.go
@@ -119,10 +119,13 @@ type NewInput struct {
 	Splay time.Duration
 
 	// Setsid flag, if set to true will create the child processes with their own
-	// session ID and Process group ID. The default value of false will cause the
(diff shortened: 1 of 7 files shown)


eikenb:
Fixes #1604

PR #1613, opened 2022-08-02
filter out vault secrets from template errors

eikenb:
Template execute errors would return the values related to the error in certain cases and these could end up in the logs. This makes sure that if those values are vault secrets they are redacted.

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -4,7 +4,9 @@ import (
 	"bytes"
 	"crypto/md5"
 	"encoding/hex"
+	"fmt"
 	"io/ioutil"
+	"strings"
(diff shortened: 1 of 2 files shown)


lornasong (approved):
makes sense to me! as a side note, wondering if it would be useful to have a test case for non-vault secrets to check that they are not redacted?

eikenb:
Thanks for the review @lornasong,

We already blank out all other known cases of secrets. This was a weird side-case of an internal template error displaying the keys related to mis-use (like ranging over a non-range value) and there are no non-vault secret types that can be checked for in a similar way.

PR #1616, opened 2022-08-09
Implements metrics using armon/go-metrics library

### LINKED ISSUE #1395: Replace OpenTelemetry with armon/go-metrics
## Summary
After a few months of weighing the current state of OpenTel vs the current need for metrics wrt to the Consul ecosystem, it has become more evident to me that switching to [`armon/go-metrics`](https://github.com/armon/go-metrics) would provide the most cohesive and comprehensive feature set at this time.

### Background
PR https://github.com/hashicorp/consul-template/pull/1378 recently introduced metrics to Consul Template using OpenTelemetry. There's a short blurb providing context as to why OpenTel was initially chosen over `armon/go-metrics` and that it was an experiment towards a new library for telemetry.

### Context
I tried to create a uniform UX for configuring and emitting metrics over on [Consul ESM #70](https://github.com/hashicorp/consul-esm/pull/70), another Consul ecosystem project, and quickly ran into some short comings when dealing with more complex metric types than counters and histograms. For example, timer values were reduced to histogram integers, and there currently is not a systematic way to denote measurement (seconds, milliseconds, etc).

The community request for `statsd` is not quite satisfied yet due to the difference in tag syntax between statsd variants and would require omitting the use of tags as a whole to continue support for statsd.

### UX
The metrics reported should not change much from the initial work in https://github.com/hashicorp/consul-template/pull/1378 (not yet released). However, the telemetry configuration will be breaking changes compared to the PR and align closer to how it is supported by Consul ([docs](https://www.consul.io/docs/agent/options#telemetry)).

Switching libraries will add feature support for:
* statsd, circonus
* more metric types, like timers and gauges
* tags and global tags
* metrics prefix
* prefix filtering

Eclion:
As suggested in the issue #1395, this PR implements a metric system using the [armon/go-metrics](https://github.com/armon/go-metrics) library along with few metrics.

To achieve such implementation, the metric system was based on the [implementation done in Consul](https://github.com/hashicorp/consul/blob/main/lib/telemetry.go) and the exposed metrics have been implemented based on the PR #1378.

However, this implementation only partially implements the mentioned metrics:
- the `armon/go-metrics` library [doesn't support yet histograms](https://github.com/armon/go-metrics/pull/98) so these metrics are not implemented in this PR.
- to avoid import cycling between the `dependency`, `config` and `telemetry` packages, metrics related to the vault token were not implemented

Resolves #1395


Eclion:
Hello @eikenb !
Do you perhaps know someone who may be available to review this PR ?

c0deaddict:
Any update on this? Would love to see this merged. We're kind of in the dark now when consul-template fails to render a template. Except when a service using the rendered template eventually breaks, but I would like to know that before hand.

cybervedaa:
+1 for this please. this will go a long way in reliably monitoring consul-template

manager/runner.go
```diff
@@ -604,21 +618,42 @@ func (r *Runner) Run() error {
 			// Record that there is at least one new render event
 			newRenderEvent = true
 
+			var label string
```
roncodingenthusiast, line 621: nitpick: since this function is getting super long, what do you think about extracting the logic for getting this label into a small helper function? 


```
func getTelemetryLabel(event *RenderEvent) string {
	var label string

	// Record that at least one template would have been rendered.
	if event.WouldRender {
		label = "would"
	}

	// Record that at least one template was rendered.
	if event.DidRender {
		label = "rendered"
	}

	if event.ForQuiescence {
		label = "quiescence"
	}

	return label
}
```
so here to end up with:

```
label := getTelemetryLabel(event)
```

manager/runner.go
```diff
@@ -604,21 +618,42 @@ func (r *Runner) Run() error {
...
+				label = "rendered"
+			}
+
+			if event.ForQuiescence {
+				label = "quiescence"
+			}
+
+			// Report the template render event
+			if label != "" {
+				telemetry.CounterTemplatesRendered.Add(
```
roncodingenthusiast, line 641: is there a reason we only report this template render event when the label is set? do you think reporting this regardless of the label help us catch some inconsistent template rending issues?

telemetry/telemetry.go
```diff
@@ -0,0 +1,116 @@
...
+		if s != nil {
+			sinks = append(sinks, s)
+		}
+	}
+
+	addSink(statsiteSink)
+	addSink(statsdSink)
+	addSink(dogstatdSink)
+	addSink(circonusSink)
+	addSink(circonusSink)
```
roncodingenthusiast, line 68: is this supposed to be repeated?
```suggestion
```

telemetry/telemetry.go
```diff
@@ -0,0 +1,116 @@
...
+	metricsConf.FilterDefault = cfg.FilterDefault
+	metricsConf.AllowedPrefixes = cfg.AllowedPrefixes
+	metricsConf.BlockedPrefixes = cfg.BlockedPrefixes
+
+	var sinks metrics.FanoutSink
+	var errors error
+	addSink := func(fn func(*config.TelemetryConfig, string) (metrics.MetricSink, error)) {
+		s, err := fn(cfg, metricsConf.HostName)
+		if err != nil {
+			log.Println("error!!!")
```
roncodingenthusiast, line 54: 🐁  this looks like a development debugging message 
```suggestion
```

telemetry/telemetry.go
```diff
@@ -0,0 +1,116 @@
...
+
+func ConfigureSinks(cfg *config.TelemetryConfig, memSink metrics.MetricSink) (metrics.FanoutSink, error) {
+	metricsConf := metrics.DefaultConfig(cfg.MetricsPrefix)
+	metricsConf.EnableHostname = !cfg.DisableHostname
+	metricsConf.FilterDefault = cfg.FilterDefault
+	metricsConf.AllowedPrefixes = cfg.AllowedPrefixes
+	metricsConf.BlockedPrefixes = cfg.BlockedPrefixes
+
+	var sinks metrics.FanoutSink
+	var errors error
```
roncodingenthusiast, line 50: i suggest using a multierror here so that at the end of the function we don't have to check for errors length and we can just do `errors.ErrorOrNil()`
```suggestion
	var errors *multierror.Error
```

telemetry/telemetry.go
```diff
@@ -0,0 +1,116 @@
...
+			errors = multierror.Append(errors, err)
+		}
+	}
+
+	// if no errors where collected, the method should not return
+	/*if len(errors.Errors) == 0 {
+		errors = nil
+	}*/
+
+	return sinks, errors
```
roncodingenthusiast, line 90: if the multierror suggestion above is taken into account then here all we have to do is:
```suggestion
	return sinks, errors.ErrorOrNil()
```

roncodingenthusiast (commented):
🎉 @Eclion Thank you so much for the thorough work that went into making this PR happen. 
Happy to work with you and a few others to get this merged/released within the next couple of weeks. 

✍🏾  I did a first pass review and i am leaving a few comments for consideration. Let me know if any of them don't make sense to you. 

❓ Do you mind updating this PR to the `main` branch, looks like it is quite being now so it will be nice to get this updated before we work on merging it in

Eclion:
> tada @Eclion Thank you so much for the thorough work that went into making this PR happen. Happy to work with you and a few others to get this merged/released within the next couple of weeks.
> 

Hello @roncodingenthusiast !
Thank you very much for your PR!

> ✍🏾 I did a first pass review and i am leaving a few comments for consideration. Let me know if any of them don't make sense to you.
> 
> question Do you mind updating this PR to the `main` branch, looks like it is quite behind now so it will be nice to get this updated before we work on merging it in
> 

Sure! I will rebase my branch before working on your comments.

> 1234 Happy to help/assist in anyway you want me to

dependency/vault_common.go
```diff
@@ -92,14 +92,19 @@ func renewSecret(clients *ClientSet, d renewer) error {
 		case err := <-renewer.DoneCh():
 			if err != nil {
 				log.Printf("[WARN] %s: failed to renew: %s", d, err)
+				//d.recordCounter("status", "stopped")
```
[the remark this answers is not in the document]
Eclion: In the original PR #1378, there were metrics updated each time a token was renewed.
I aimed to apply the same logic but couldn't due as adding metrics in the dependency package now induces import cycling.
I didn't want to re-organise the whole codebase so I just commented out these parts.

I am unsure of how I should proceed though. Should I:
- remove the comments and it will be the job of someone who wants metrics on token renewal?
- update the commented code (in order to at least make it similar to other counters), and add a bit of context in the comments?

README.md
```diff
@@ -405,6 +408,71 @@ $ consul-template -log-level debug ...
 # ...
 ```
 
+## Telemetry
+
+Consul Template uses the [armon/go-metrics](https://github.com/armon/go-metrics) library to implement the Consul Template metric system. It currently supports metrics exported to circonus API, statsd server, statsite server, dogstatsd server, and prometheus endpoint.
```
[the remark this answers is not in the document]
Eclion: Updated!

telemetry/metrics.go
```diff
@@ -0,0 +1,52 @@
...
+	ConstLabels: []metrics.Label{},
+	Description: "The number of commands executed with labels status=(success|error)",
+}
+
+func NewLabel(name string, value string) metrics.Label {
+	return metrics.Label{Name: name, Value: value}
+}
+
+func InitMetrics() {
+	CounterActions.Add(0)
```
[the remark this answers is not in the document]
Eclion: There were none so I added the the others too.
Thanks for pointing out.

config/convert.go
```diff
@@ -22,6 +22,15 @@ func BoolVal(b *bool) bool {
 	return *b
 }
 
+// BoolCopy returns a copy of the boolean pointer
+func BoolCopy(b *bool) *bool {
```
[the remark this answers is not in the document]
Eclion: I removed this part of the code, as I found out it was not used anywhere :sweat_smile:

config/convert.go
```diff
@@ -115,6 +124,46 @@ func IntPresent(i *int) bool {
...
 }
 
+// Uint returns a pointer to the given uint.
+func Uint(i uint) *uint {
+	return &i
+}
+
+// UintVal returns the value of the uint at the pointer, or 0 if the pointer is
+// nil.
+func UintVal(i *uint) uint {
```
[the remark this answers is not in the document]
Eclion: None indeed! Unused code removed.

config/telemetry.go
```diff
@@ -0,0 +1,402 @@
...
+		BlockedPrefixes:                    c.BlockedPrefixes,
+		MetricsPrefix:                      c.MetricsPrefix,
+		StatsdAddr:                         c.StatsdAddr,
+		StatsiteAddr:                       c.StatsiteAddr,
+		PrometheusPort:                     c.PrometheusPort,
+		PrometheusRetentionTime:            c.PrometheusRetentionTime,
+	}
+}
+
+func (c *TelemetryConfig) Merge(o *TelemetryConfig) *TelemetryConfig {
```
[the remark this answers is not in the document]
Eclion: Comment added!

manager/runner.go
Eclion: Suggestion applied!

manager/runner.go
Eclion: As you suggested, reporting it regardless could help us for future updates, rather than skipping missing labels.
`if` branch removed.

telemetry/telemetry.go
Eclion: it shouldn't, and have been removed.

telemetry/telemetry.go
Eclion: Thanks for spotting it! I removed the development debug logging.

telemetry/telemetry.go
Eclion: I didn't know about it until recently, and it is indeed very helpful!
Code updated to include suggestion.

telemetry/telemetry.go
Eclion: code updated.

dependency/vault_common.go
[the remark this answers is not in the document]
Eclion: I removed the comments for now, as we could always implement metrics later.
The main focus of this PR was to at least implement metrics related to the templating.

c0deaddict:
I would very much like for this PR to be merged - and released. Any update on this?

cybervedaa:
I'm eagerly waiting on this to be merged as well. Can't wait for this to be merged

hkrutzer:
Hoping this year will be the year of consul-template metrics. @nathancoleman @jmurret @jm96441n @DanStough I see you have recently done work on this repo, perhaps one of you can give a status update?

rboyer:
Quoting @jmurret's [reply](https://github.com/hashicorp/serf/pull/693#issuecomment-2211215315) on another nearly identical PR:

> Hi all, we will post a more detailed thread on why this is complicated, but the short gist is that this change has to be consistently updated across the dependency chains of libraries used in an application.  `go-metrics` relies on a singleton sink in global state. If any applications have a mix of `armon` metrics and `hashicorp` due to underlying dependencies not aligning on only one of them, metrics will be split among different sinks in the global state. 
This change has to be coordinated with changing and releasing this dependency across multiple libraries and applications in a thoughtful way.  This PR is taking longer than expected due to the prioritization and coordination of those efforts to ensure a smooth transition.  Thank you for your patience and understanding.

config/telemetry.go
```diff
@@ -0,0 +1,405 @@
...
+}
+
+func (c *TelemetryConfig) GoString() string {
+	if c == nil {
+		return "(*TelemetryConfig)(nil)"
+	}
+	return fmt.Sprintf("&TelemetryConfig{"+
+		"Disable:%v, "+
+		"CirconusAPIApp:%s, "+
+		"CirconusAPIToken:%s, "+
```
jm96441n, line 341: are we worried about leaking the API token in logs here?

config/telemetry.go
```diff
@@ -0,0 +1,405 @@
...
+		StatsiteAddr:                       c.StatsiteAddr,
+		PrometheusPort:                     c.PrometheusPort,
+		PrometheusRetentionTime:            c.PrometheusRetentionTime,
+	}
+}
+
+// Merge combines all values in this configuration with the values in the other
+// configuration, with values in the other configuration taking precedence.
+// Maps and slices are merged, most other values are overwritten.
+func (c *TelemetryConfig) Merge(o *TelemetryConfig) *TelemetryConfig {
```
jm96441n, line 246: can we add tests for this merge function?

config/telemetry.go
jm96441n, line 341: are we worried about leaking the API token in logs here?

dependency/dependency.go
```diff
@@ -49,6 +49,14 @@ const (
 	DefaultContextTimeout = 60 * time.Second
 )
 
+func (t Type) String() string {
+	if t > 2 {
```
jm96441n, line 53: minor: could we have a comment here indicating why `2` or use a named constant?

manager/runner.go
```diff
@@ -1478,3 +1511,30 @@ func newWatcher(c *config.Config, clients *dep.ClientSet) *watch.Watcher {
 		RetryFuncNomad:   watch.RetryFunc(c.Nomad.Retry.RetryFunc()),
 	})
 }
+
+func recordDependencyCounts(deps map[string]dep.Dependency) {
```
jm96441n, line 1515: this function doesn't return anything or modify the input in any way, what is it meant to do?

telemetry/telemetry_test.go
```diff
@@ -0,0 +1,135 @@
...
+		if strings.HasPrefix(actualMetric, "# HELP go_") ||
+			strings.HasPrefix(actualMetric, "# TYPE go_") ||
+			strings.HasPrefix(actualMetric, "go_") ||
+			strings.HasPrefix(actualMetric, "# HELP process_") ||
+			strings.HasPrefix(actualMetric, "# TYPE process_") ||
+			strings.HasPrefix(actualMetric, "process_") ||
+			actualMetric == "" {
+			continue
+		}
+
```
jm96441n, line 110: minor but this might be a bit more readable
```suggestion
  prefixes := []string{"# HELP go_", "#TYPE go", "go_", "# HELP process_", "# TYPE process_", "process_"}
	for _, actualMetric := range actualMetrics {
		if slices.ContainsFunc(prefixes, func(p string) bool { return strings.HasPrefix(actualMetric, p) }) ||
			actualMetric == "" {
			continue
		}

```

jm96441n (changes requested):
overall looks good, just a few additional changes and I'm happy to merge

config/telemetry.go
Eclion: Sure!
I added some tests as requested

config/telemetry.go
Eclion: Thanks for spotting this issue. The output value has been replaced by `<empty>` and `<configured>`.

dependency/dependency.go
Eclion: It was used in the upstream PR to default to `unknown` probably because `Nomad` was not part of the `Type` list at that time.
I replaced this section by a switch case, more explicit

telemetry/telemetry_test.go
Eclion: Thanks for the suggestion! I agree with it being more readable.

Eclion:
Hello @jm96441n,
Thank you very much for your review!!
I updated the code to address your comments. Could you please review them once you have a bit of time?

Eclion:
> Nomad embeds consul-template as a library rather than using it as a standalone binary. The HTTP server and telemetry sink should be configurable by passing in an existing server and sink.

Hello @tgross ,

If I am not wrong, there is no need to call `telemetry.Init()` if `metrics.NewGlobal` is called in the code somewhere.

So the only needed call, as far as I understand, would be `telemetry.CounterTemplatesRendered.Add(1)` when a template is rendered.

However, such call is not implemented in the `renderer.Render` method, used by Nomad, but in the `manager.RunnerRun.` method, where we can also keep track of which template was rendered.

tgross:
> However, such call is not implemented in the renderer.Render method, used by Nomad, but in the manager.RunnerRun. method, where we can also keep track of which template was rendered.

The `Render` call there is only so that Nomad sandbox the rendered. Nomad calls `manager.Runner.Start()` which in turn calls `manager.Runner.Run()`. But to your point, you're right only the cli here is calling `Init` and instantiating the sink. No objections from our end.

Tyrael:
hi, any update on this?

PR #1630, opened 2022-08-15
fix explodeMap example error  in doc

bigfish02:


eikenb (approved):
Thanks for the documentation fix! Much appreciated.

PR #1631, opened 2022-08-15
Support Nomad Secure Variables

angrycub:
This PR extends consul-template to support Nomad Secure Variables. It adds 4 additional functions:
- nomadVars
- nomadVarsSafe
- nomadVar
- nomadVarExists

There is a circular dependency issue with Nomad, so the tests won't pass until Nomad has released a version that contains secure variables. I'm not sure how exactly to manage this, but I need to get the PR up at least.


angrycub:
Closed in favor of #1632

PR #1632, opened 2022-08-15
Support Nomad Variables

angrycub:
This PR extends consul-template to support NomadVariables. It adds 4 additional functions:

- nomadVars
- nomadVarsSafe
- nomadVar
- nomadVarExists

There is a circular dependency issue with Nomad, so the tests won't pass until Nomad has released a version that contains secure variables. I'm not sure how exactly to manage this, but I need to get the PR up at least.

### DIFF
--- a/dependency/dependency_test.go
+++ b/dependency/dependency_test.go
@@ -413,3 +413,31 @@ func Fatalf(format string, args ...interface{}) {
 	fmt.Printf(format, args...)
 	runtime.Goexit()
 }
+
+func (v *nomadServer) CreateVariable(path string, data map[string]string, opts *nomadapi.WriteOptions) error {
+	nVar := nomadapi.NewVariable(path)
+	for k, v := range data {
+		nVar.Items[k] = v
+	}
+	_, _, err := testClients.Nomad().Variables().Update(nVar, opts)
+	if err != nil {
+		fmt.Println(err)
+	}
+	return err
+}
+func (v *nomadServer) CreateNamespace(name string, opts *nomadapi.WriteOptions) error {
+	ns := nomadapi.Namespace{Name: name}
+	_, err := testClients.Nomad().Namespaces().Register(&ns, opts)
+	if err != nil {
+		fmt.Println(err)
+	}
+	return err
+}
+
+func (v *nomadServer) DeleteVariable(path string, opts *nomadapi.WriteOptions) error {
+	_, err := testClients.Nomad().Variables().Delete(path, opts)
+	if err != nil {
+		fmt.Println(err)
+	}
+	return err
+}
--- a/docs/templating-language.md
+++ b/docs/templating-language.md
@@ -99,6 +99,10 @@ provides the following functions:
 - [Nomad Functions](#nomad-functions)
   - [nomadServices](#nomadservices)
   - [nomadService](#nomadservice)
+  - [nomadVarList](#nomadvarlist)
+  - [nomadVarListSafe](#nomadvarlistsafe)
+  - [nomadVar](#nomadvar)
+  - [nomadVarExists](#nomadvarexists)
 - [Debugging Functions](#debugging)
   - [spew_dump](#spew_dump)
   - [spew_sdump](#spew_sdump)
@@ -1704,7 +1708,7 @@ For example:
 ## Sprig Functions
 
 Consul-template provides access to the Sprig library
-in templates. To use a Sprig function in your template, prepend `sprig_` to the function name.  A full list of Sprig functions can be found in the [Sprig Function Documentation](http://masterminds.github.io/sprig/)
+in templates. To use a Sprig function in your template, prepend `sprig_` to the function name. A full list of Sprig functions can be found in the [Sprig Function Documentation](http://masterminds.github.io/sprig/)
 
 ---
 
@@ -1819,13 +1823,14 @@ This can also be used with a pipe function.
 ## Nomad Functions
 
 Nomad service registrations can be queried using the `nomadServices` and `nomadService` functions.
+Nomad variables can be queried using the `nomadVarList` and `nomadVar` functions.
 Typically these will be used from within a Nomad [template](https://www.nomadproject.io/docs/job-specification/template#nomad-services) configuration.
 
 ### `nomadServices`
 
 This can be used to query the names of services registered in Nomad.
 
-```nomadServices
+```golang
 {{ range nomadServices }}
   {{ .Name .Tags }}
 {{ end }}
@@ -1835,7 +1840,7 @@ This can be used to query the names of services registered in Nomad.
 
 This can be used to query for additional information about each instance of a service registered in Nomad.
 
-```nomadService
+```golang
 {{ range nomadService "my-app" }}
   {{ .Address }} {{ .Port }}
 {{ end}}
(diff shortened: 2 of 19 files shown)


eikenb:
> [..] so the tests won't pass until Nomad has released a version that contains secure variables.

Is it upstream yet? We can do a commit-id update for now if necessary.

Do you want these merged before the 0.29.2 release I'm planning for this week?

tgross:
> Is it upstream yet? We can do a commit-id update for now if necessary.

Everything we need to support this in the API is on `main`. We're expecting to release Nomad 1.4.0-beta in the next few weeks.

> Do you want these merged before the 0.29.2 release I'm planning for this week?

Maybe we hold off on that? When I sent @angrycub off on this particular adventure 😀 I wasn't keeping in mind the beta. We can always pin to a commit-id on the Nomad side for the beta release and then if there's anything that comes up in the beta window where we really hate the API and need to change it, we're not causing regressions on a released version of `consul-template`.

So what do you (@eikenb and @angrycub) this about the following:
* For this PR we pin the go.mod to a commit-id on Nomad.
* `consul-template` 0.29.2 ships without this.
* We ship Nomad 1.4.0-beta in the next few weeks, pinning our CT dependency on the head of this PR's branch.
* Sometime during the beta window we merge this PR.
* Nomad 1.4.0 GA pins CT to a commit-id on CT's main at that point, or a hypothetical 0.29.x that lands afterwards.

That'll prevent Nomad from causing extra release schedule pressure for y'all, and give us a little breathing room around the Nomad 1.4.0 beta as well.

eikenb:
Thanks @tgross. Sounds like a good plan.

dependency/nomad_secure_variables_get_test.go
```diff
@@ -0,0 +1,294 @@
...
+			assert.Equal(t, tc.exp, act)
+		})
+	}
+	fmt.Println("done")
+}
+
+func TestSVGetQuery_Fetch(t *testing.T) {
+
+	type svmap map[string]string
+	_ = testNomad.CreateSecureVariable("test-kv-get/path", svmap{"bar": "barp"})
```
tgross, line 128: Rather than `fmt.Printf`'ing in the test helper, we should probably return the error here and `t.Fatal` (or `require.NoError`, as we have testify in this repo); there's no reasonable behavior for the test after this point.

template/template_test.go
```diff
@@ -2025,6 +2025,103 @@ func TestTemplate_Execute(t *testing.T) {
 			"([]string) (len=3 cap=3) {\n (string) (len=1) \"a\",\n (string) (len=1) \"b\",\n (string) (len=1) \"c\"\n}\n",
 			false,
 		},
+		{
+			"func_nomadSecureVariable",
+			&NewTemplateInput{
+				Contents: `{{with nomadVar "path" }}{{.k1}}{{end}}`,
```
tgross, line 2031: Is it at all useful to have some tests exercising the namespace argument or some of the field access around metadata vs items?

docs/templating-language.md
```diff
@@ -1838,13 +1843,153 @@ the number of instances desired, a unique but consistent identifier associated w
...
+	Key, Value string
+}
+```
+
+`NomadSVItem` objects also provide helper methods:
+
+* `Metadata`: returns this item's parent's metadata as a `NomadSVMeta`
+
+* `Parent`: returns the consul-template version (NomadSecureVariable) of the
+    secure variable that contains this item.
```
tgross, line 1970: I suspect helper methods are not-too-unintuitive for go devs but for operators it might be useful to specifically document an example of their use here.

dependency/nomad_secure_variables_structs.go
```diff
@@ -0,0 +1,142 @@
...
+	sort.Strings(out)
+	return out
+}
+
+// Values produces a key-sorted list of the Items map's values
+func (v NomadSVItems) Values() []string {
+	out := make([]string, 0, len(v))
+	for _, k := range v.Keys() {
+		out = append(out, v[k].String())
+	}
```
tgross, line 81: This ends up allocating two slices instead of one because of the call to Keys. Can we do `for k, val := range v` instead? (Same applies to `Tuples()` below.)

dependency/dependency_test.go
```diff
@@ -413,3 +413,24 @@ func Fatalf(format string, args ...interface{}) {
...
+}
+
+// deleteSecureVariable lets us delete keys as needed for tests
+func (v *nomadServer) deleteSecureVariable(path string) error {
+	_, err := testClients.Nomad().SecureVariables().Delete(path, nil)
+	if err != nil {
+		fmt.Println(err)
+	}
+	return err
+}
```
tgross, line 436: This function has no callers, so we're probably missing cleanup functions in the tests.

dependency/nomad_secure_variables_list_test.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+
+	for i, tc := range cases {
+		t.Run(fmt.Sprintf("%d_%s", i, tc.name), func(t *testing.T) {
+			d, err := NewSVListQuery(tc.i)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			act, _, err := d.Fetch(testClients, nil)
```
tgross, line 175: It doesn't look like any of the `Fetch` calls include a namespace, so I suspect we don't have good test coverage of the non-default namespace code paths.

dependency/nomad_secure_variables_structs.go
angrycub: My main goal in using `Keys()` to create the other two outputs is that I wanted key-sorted order into the respective slices without repeating the sort code, so that they would range in the same order that go-template would order the map output.  Go template will do an implicit sort of keys so that they range consistently.

tgross: Oh of course! That makes sense.

dependency/nomad_secure_variables_list_test.go
angrycub: Currently, the secure variables functions are only configured for the current namespace.  As you mentioned synchronously, there is a need for namespace support for service discovery as well.

template/template_test.go
angrycub: Added tests to exercise namespace and region

dependency/nomad_secure_variables_list_test.go
angrycub: Updated to support namespaces and region

docs/templating-language.md
```diff
@@ -1810,13 +1814,14 @@ This can also be used with a pipe function.
 ## Nomad Functions
 
 Nomad service registrations can be queried using the `nomadServices` and `nomadService` functions.
+Nomad variables can be queried using the `nomadVarList` and `nomadVar` functions.
```
tgross, line 1817: It might be nice to link Nomad variables to https://www.nomadproject.io/docs/concepts/variables (and Nomad service registrations to https://www.nomadproject.io/docs/job-specification/service#provider )

go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	github.com/hashicorp/go-syslog v1.0.0
 	github.com/hashicorp/hcl v1.0.0
 	github.com/hashicorp/logutils v1.0.0
-	github.com/hashicorp/nomad/api v0.0.0-20220707195938-75f4c2237b28
+	github.com/hashicorp/nomad/api v0.0.0-20220829153708-e1e5bb1dcefb
```
tgross, line 19: Before we merge, let's update this again to the current beta version (at least).

tgross:
I've opened https://github.com/hashicorp/nomad/issues/14762 with a reminder to ourselves to update Nomad to point to the new release of CT once this has been merged.

.github/workflows/ci.yml
```diff
@@ -26,6 +26,7 @@ jobs:
           curl -fsSL https://apt.releases.hashicorp.com/gpg | sudo apt-key add -
           sudo apt-add-repository "deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_release -cs) main"
           sudo apt-get update && sudo apt-get install consul vault nomad
+          wget https://releases.hashicorp.com/nomad/1.4.0-beta.1/nomad_1.4.0-beta.1_linux_amd64.zip && unzip nomad_1.4.0-beta.1_linux_amd64.zip && sudo mv nomad /usr/bin
```
eikenb, line 29: Is this going to be removed before merging?

angrycub: Yeah, it was the only way to actually have the tests pass in the interim.  Once there are real releases, we will pull this out.

angrycub:
Updated the Nomad API dependency to 1.4.1's /api version.  Reverted the test shenanigans. Things look to be green and ready to rock @eikenb, @tgross

docs/templating-language.md
```diff
@@ -1838,13 +1843,153 @@ the number of instances desired, a unique but consistent identifier associated w
...
+
+```golang
+type NomadVarMeta struct {
+	Namespace, Path          string
+	CreateIndex, ModifyIndex uint64
+	CreateTime, ModifyTime   nanoTime
+}
+```
+
+The `nanoTime` type contains Unix nanoseconds since the epoch. It's String method
```
tgross, line 1901:
```suggestion
The `nanoTime` type contains Unix nanoseconds since the epoch. Its String method
```

tgross (approved):
👍

.go-version
```diff
@@ -0,0 +1 @@
+1.19.1
```
eikenb, line 1: I'm moving all my projects to use go.mod for version info. So this file will be ignored and confusing. Would you mind removing it?

eikenb (changes requested):
Everything looks good except that one file. Thanks.

.go-version
angrycub: Will consul-template not be switching to the common release tooling?  The CRT actions use the .go-version file to determine what go version to install.

eikenb: Consul-template already uses CRT. CRT provides a generic example config for the github action workflows that needs to be tweaked and maintained for each project. It is not a fixed script. I didn't want/need the detailed control of .go-version but I did need it to use the appropriate compiler version for the code (as is specified in go.mod). Github actions supports doing that and it was only a [minor change to the build.yml](https://github.com/hashicorp/consul-template/commit/47c9feceb89932789301ae679ee89de61b7011a5).

eikenb (approved):

eikenb:
@angrycub... could you rebase on main/HEAD to get the go.mod build changes? This should allow the tests to pass (or at least fix what's currently failing). Thanks.

PR #1639, opened 2022-09-16
Attempt to resolve GH 17166 where agent crashes when no certs available

peteski22:


dependency/vault_pki_test.go
```diff
@@ -41,6 +41,17 @@ func Test_VaultPKI_notGoodFor(t *testing.T) {
 	}
 }
 
+func Test_VaultPKI_pemsCertNone(t *testing.T) {
+	_, cert, err := pemsCert(make([]byte, 0))
+	if err == nil {
+		t.Fatal("error should be not be nil as we don't have a cert")
```
eikenb, line 47: Could you change these to t.Error(), been trying to move away from using Fatal in tests (unless they'll actually break other tests).

eikenb: Oh, and would you mind adding a comment for that test mentioning that it is checking for the case where the file that should contain the certs, doesn't. Thanks.

peteski22: Hey @eikenb no worries, I moved the check around in the code and tests so that the cert is renewed (and that hopefully fixes the issue once it's re-written to disk after the restart).

eikenb (approved):
The changes to the fix make sense. This fixes the issue without introducing a new error condition that might have unintended side effects. :+1:

PR #1647, opened 2022-09-28
Add docs for template file ownership

kencx:
This PR adds missing documentation for the setting of template file ownership added in #1541.


eikenb (approved):
Thanks for the documentation PR @kencx! :pray:

PR #1657, opened 2022-10-20
add mustEnv function

C-Pro:
## Background
It is important to fail early and not try to deploy a config with wrong data. But when populating values using `env` function a typo in variable name will lead to an empty value instead of an error, like when trying to access nonexistent vault key with `error_on_missing_key` enabled.

This PR addresses it with adding extra `mustEnv` command that fails if environment variable does not exist or is empty.

```
$ cat env.tmpl
URL={{mustEnv "ENV"}}:12312

$ ENV=test ~/go/bin/consul-template -once -template=env.tmpl -dry
>
URL=test:12312

$ ~/go/bin/consul-template -once -template=env.tmpl -dry
2022-10-20T18:21:52.794+0700 [ERR] (cli) env.tmpl: execute: template: :1:6: executing "" at <mustEnv "ENV">: error calling mustEnv: required environment variable ENV is empty
```

### DIFF
--- a/docs/templating-language.md
+++ b/docs/templating-language.md
@@ -47,6 +47,7 @@ provides the following functions:
   - [containsNone](#containsnone)
   - [containsNotAll](#containsnotall)
   - [env](#env)
+  - [mustEnv](#mustEnv)
   - [envOrDefault](#envOrDefault)
   - [executeTemplate](#executetemplate)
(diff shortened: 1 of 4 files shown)


eikenb (approved):
code, docs, tests... :+1:

PR #1662, opened 2022-11-02
fix: user set in dockerfile

matthugs:
my organization is currently auditing whether any docker images we run are currently using the root user. From dockerhub it looks like this line expands out to `USER :` since the env var names don't match. (see https://hub.docker.com/layers/hashicorp/consul-template/0.29/images/sha256-32f8ffb8db51809d93488527d63f97d267e2a10c0cbe96dfa5cbc408266211ab?context=explore) This would fix this so that your images are run as the user you create earlier in the dockerfile.


eikenb (approved):
Looks like that got missed when migrating to our new release process. Thanks!

PR #1678, opened 2022-12-04
Fix 'toTitle' function for a better support for word boundaries and unicode punctuation

### LINKED ISSUE #1677: toTitle function fails to understand word boundaries and unicode punctuation
### Consul Template version

```consul-template v0.29.6 (e4f5d583)```

### Expected behavior

Consider the following template:

```hcl
{{ "this isn't ok" | toTitle }}
{{ "foo\u2026bar" | toTitle }}
```

Expected output:

```
This Isn't Ok
Foo…Bar
```

### Actual behavior

Template above yields the following:

```
This Isn'T Ok
Foo…bar
```

Title casing currently fails to understand word boundaries properly, leading to various issues with hyphenation and unicode punctuation.

### References

Current implementation of `toTitle` function is based on `strings.Title`, which has been deprecated since Go 1.18. Instead, `golang.org/x/text/cases` should be used:

- https://github.com/golang/go/issues/48367

vtorhonen:
- Use `golang.org/x/text/cases` package in the `toTitle` function. Previous method `strings.Title` has been deprecated since Go 1.18. 

- Add a test to title case a sentence with unicode punctuation and a rocket emoji 🚀 

Fixes #1677

### DIFF
--- a/template/funcs.go
+++ b/template/funcs.go
@@ -28,6 +28,8 @@ import (
 	socktmpl "github.com/hashicorp/go-sockaddr/template"
 	"github.com/imdario/mergo"
 	"github.com/pkg/errors"
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
 	yaml "gopkg.in/yaml.v2"
 )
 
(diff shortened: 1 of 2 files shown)


eikenb (approved):
:+1:

PR #1680, opened 2022-12-04
Fix bitsize in parseFloat function

vtorhonen:
`parseFloat` function calls [strconv.ParseFloat](https://pkg.go.dev/strconv#ParseFloat) with an invalid bitsize parameter. This parameter should either be `32` (for 32-bit precision) or `64` (for 64-bit precision). The result is always of type `float64`.

Not a user-impacting change, since the [internal implementation of `strconv.ParseFloat`](https://cs.opensource.google/go/go/+/refs/tags/go1.19.3:src/strconv/atof.go;l=695) defaults to bitsize of `64` if the given bitsize is anything else than `32`.

### DIFF
--- a/template/funcs.go
+++ b/template/funcs.go
@@ -1052,13 +1052,13 @@ func parseBool(s string) (bool, error) {
 	return result, nil
 }
 
-// parseFloat parses a string into a base 10 float
+// parseFloat parses a string into a floating-point number with 64-bit precision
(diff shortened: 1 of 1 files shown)


eikenb (approved):
Odd error.. and I think you're right that this should have no real impact on users as it would work the same but with the current implementation but is good to fix in case of future changes.

PR #1681, opened 2022-12-04
Remove unnecessary type conversions from various places

vtorhonen:
In detail:

- `config/convert_test.go`: `time.Second` & others are already represented (internally) by an `int64` so type conversion after multiplication is not needed
- `manager/runner_test.go`: Same as above + `act` is defined as a string from the start, no need for conversion
- `template/funcs.go`: Upon reflection `v.Int()` already returns an `int64`


eikenb (approved):
The small cleanups keep coming. Thanks!

PR #1690, opened 2023-01-03
respect reload/kill configured signals

### LINKED ISSUE #1671: Respect `splay` for all ways a child process can get a reload signal
### Consul Template version

`0.29.5`

### Configuration

```hcl
log_level = "debug"

# Setting this value to the empty string will cause consul-template
# to not listen for any reload signals but pass them down to nginx instead.
reload_signal = ""

exec {
  kill_signal = "SIGQUIT"
  reload_signal = "SIGHUP"
  kill_timeout = "5m"

  splay = "15s"

  command = ["nohup", "sleep", "100000"]
}

```

### Expected behavior

In current release version of consul-template does not apply the `splay` time when a reload comes from outside of consul-template in the case when we set `reload_signal = ""` at the top level config. I provide a sample code that I was experimenting with just to give an idea of where the problem could be addressed.

The real end goal of what we are trying to achieve is to "debounce" reload signals but respecting the splay time (any reloads that come while the splay is ticking, should be dropped and use that very first reload to actually reload the `exec` process. We have a problem where our `nginx` process is reloading many times per second, causing stability issues (just to give more context).

### References

```golang
// Signal sends the signal to the child process, returning any errors that occur.
func (c *Child) Signal(s os.Signal) error {
	c.logger.Printf("[INFO] (child) receiving signal %q", s.String())
	c.RLock()
	defer c.RUnlock()

	switch s {
	case c.reloadSignal:
		return c.reload()
	case c.killSignal:
		c.kill(true)
		return nil
	default:
		return c.signal(s)
	}
}
```

eikenb:
In cases it would just send the signal instead of matching against configured signals for special behavior.

Fixes #1671

### DIFF
--- a/child/child.go
+++ b/child/child.go
@@ -211,7 +211,15 @@ func (c *Child) Signal(s os.Signal) error {
 	c.logger.Printf("[INFO] (child) receiving signal %q", s.String())
 	c.RLock()
 	defer c.RUnlock()
-	return c.signal(s)
+	switch s {
+	case c.reloadSignal:
+		return c.reload()
+	case c.killSignal:
+		c.kill(true)
+		return nil
+	default:
+		return c.signal(s)
+	}
 }
 
 // Reload sends the reload signal to the child process and does not wait for a


komapa:
thank you, you are the best!

PR #1699, opened 2023-01-09
[dep] Update Sprig from v2 to v3

angrycub:
Updating the sprig dependency to reduce drift from its current online documentation.


roncodingenthusiast (approved):
LGTM

PR #1705, opened 2023-02-03
Implement reload signal debouncing

komapa:
Hello, this is an implementation of debouncing reload signals so that only a single reload happens in the span of a single `splay` period.

I think the changes are pretty self-explanatory and the only thing that needed to change was starting `runner.Signal(s)` in a go routine so that we can do all operations in parallel and not sequential. 

Please let me know if there is a better way to do this. Thank you!


komapa:
Just a note that we have been running this in production for the last 2-3 weeks and it has been performing well for us.

roncodingenthusiast:
@komapa is this still running in production, any issues you have noticed? 

This is looking good, do you also mind rebasing this before we approve/merge this?

komapa:
Hey @roncodingenthusiast, we ran this for more than a month in production and we did not see any issues. We eventually removed it because we were able to stop the reloads altogether and we did not want to keep having to apply the patch. I think it will be great to get this merged in if you think it looks good.

PR #1708, opened 2023-02-09
[feature] Allow external FuncMaps

angrycub:
**Rebased and reopened**
 
Since Nomad uses CT as its underlying template engine, it would be nice to be able to pass functions from Nomad itself into the template configurations that it generates. I thought exposing a means to inject functions for applications that use CT internally might be a good middle ground to prevent even more template application sprawl.

This would allow Nomad to inject purpose-built templating functions into the rendering pipeline without having to make additional changes to consul-template.

Custom functions could also be used to simplify the FunctionBlockList by allowing someone to implement their own no-op function to opaque an implementation provided in CT.

This could let us resolve #1312 and #1406 by allowing them to be passed into CT from Nomad using this scheme.


template/template_test.go
```diff
@@ -2610,3 +2626,70 @@ arHERAKScrZMTrYPLt2YqYoeyO//aCuT9YW6YdIa9jPQhzjeMKXywXLetE+Ip18G
...
+			},
+			&ExecuteInput{
+				Brain: NewBrain(),
+			},
+			"ABCBA lower",
+			false,
+		},
+	}
+
+	//	struct {
```
roncodingenthusiast, line 2669: is this comment still needed? I am assuming it was just forgotten from your testing and can probably be removed

roncodingenthusiast (approved):

PR #1725, opened 2023-03-21
VAULT-12940 Add ability to configure Vault client user agent

VioletHynes:
This adds a new configuration item for Vault, `ClientUserAgent`. When set, Consul Template will use the set `User-Agent` when making requests to Vault.

This change is being made as part of a broader effort for Vault Agent to send its version as part of a `User-Agent` string in requests to Vault. Agent will then consume the latest version of `consul-template`, then use this new config to set the correct `User-Agent`.

Tests were of course added to validate that this works as expected, but let me know if there's anything else I should be testing as part of this change.

### DIFF
--- a/dependency/client_set.go
+++ b/dependency/client_set.go
@@ -86,17 +86,18 @@ type CreateConsulClientInput struct {
 
 // CreateVaultClientInput is used as input to the CreateVaultClient function.
 type CreateVaultClientInput struct {
-	Address     string
-	Namespace   string
-	Token       string
-	UnwrapToken bool
-	SSLEnabled  bool
-	SSLVerify   bool
-	SSLCert     string
-	SSLKey      string
-	SSLCACert   string
-	SSLCAPath   string
-	ServerName  string
+	Address         string
+	Namespace       string
+	Token           string
+	UnwrapToken     bool
+	SSLEnabled      bool
+	SSLVerify       bool
+	SSLCert         string
+	SSLKey          string
+	SSLCACert       string
+	SSLCAPath       string
+	ServerName      string
+	ClientUserAgent string
 
 	K8SAuthRoleName            string
 	K8SServiceAccountTokenPath string
@@ -335,6 +336,11 @@ func (c *ClientSet) CreateVaultClient(i *CreateVaultClientInput) error {
 	client, err := vaultapi.NewClient(vaultConfig)
 	if err != nil {
 		return fmt.Errorf("client set: vault: %s", err)
+	}
+
+	if i.ClientUserAgent != "" {
+		client.SetCloneHeaders(true)
+		client.AddHeader("User-Agent", i.ClientUserAgent)
 	}
 
 	// Set the namespace if given.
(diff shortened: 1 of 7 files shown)


roncodingenthusiast (commented):
Reviewed this with @eikenb. This looks great so far and should be good to go once some docs are added [here](https://github.com/hashicorp/consul-template/blob/main/docs/configuration.md#vault). Thank you for adding tests

dependency/client_set.go
```diff
@@ -86,17 +86,18 @@ type CreateConsulClientInput struct {
...
+	Token           string
+	UnwrapToken     bool
+	SSLEnabled      bool
+	SSLVerify       bool
+	SSLCert         string
+	SSLKey          string
+	SSLCACert       string
+	SSLCAPath       string
+	ServerName      string
+	ClientUserAgent string
```
peteski22, line 100: ❓Do we call this `ClientUserAgent` within `CreateVaultClientInput` (as opposed to just `UserAgent`) because Vault (server) is the client here? Like the client (agent) of the client (server)? :)

VioletHynes: `CreateVaultClientInput` builds a Vault Client, essentially, so `ClientUserAgent` determines the User-Agent of that client. As far as why I didn't call it `UserAgent`, it's mostly just because I found `ClientUserAgent` clearer, personally anyway.

roncodingenthusiast (approved):

dependency/client_set.go
peteski22: Ah OK, thanks. 👍🏼 

Does that mean for example, that `Address` is a client address? I was wondering if there's some fields that are server specific and some that are client specific (`ServerName` looks suspicious in that regard).

Sorry it now sounds like a nitpick, I was just trying to understand it... why wouldn't we follow the same format of non-client prefixed fields and just use `UserAgent`? It feels a bit redundant.

VioletHynes: The `Address` is the server address. Ultimately, this goes into configuring the client.

The configuration is done via HCL, e.g. this:
```
vault {
  address = "https://vault.service.consul:8200"
  namespace = ""
  client_user_agent = "Consul Template"
#...
```

Especially in this context, I feel like the 'client' part of the config tells the user something, i.e. that this configuration is for Consul Template's Vault Client. I definitely err on the side of making things as unambiguous as possible. At least to me, `user_agent` tells me a lot less than `client_user_agent` in that HCL block.

peteski22: Cool, so there are fields here that are related to the server properties which are configured within the client, that makes sense. 😄 

What I was alluding to was being able to easily distinguish between the types of setting, so like `ServerAddress` and now `ClientUserAgent`, so the client specific settings for client and server settings for client were clear in code. 

I agree with what you're saying about the clarity for the end-user, I think the HCL config example makes sense. 👍🏼 

Maybe worth pointing out that the HCL name isn't the same as the name of the field we use in code though, it's just mapped via: `ClientUserAgent *string ``mapstructure:"client_user_agent""```

No more questions from me, thanks!

peteski22 (approved):

PR #1731, opened 2023-03-28
Replace uses of the word "sane"

angrycub:
Switching to more inclusive language by replacing "sane" with "reasonable".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -482,7 +482,7 @@ BUG FIXES:
 
 BREAKING CHANGES:
 
-  * Retry now has a sane maximum default. Previous versions of Consul Template
+  * Retry now has a reasonable maximum default. Previous versions of Consul Template
       would retry indefinitely, potentially allowing the time between retries to
       reach days, months, or years due to the exponential nature. Users wishing
       to use the old behavior should set `max_backoff = 0` in their
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -482,7 +482,7 @@ BUG FIXES:
 
 BREAKING CHANGES:
 
-  * Retry now has a reasonable maximum default. Previous versions of Consul Template
+  * Retry now has a maximum default. Previous versions of Consul Template
       would retry indefinitely, potentially allowing the time between retries to
       reach days, months, or years due to the exponential nature. Users wishing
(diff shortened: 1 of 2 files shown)


manager/runner.go
```diff
@@ -27,9 +27,9 @@ import (
 )
 
 const (
-	// saneViewLimit is the number of views that we consider "sane" before we
```
eikenb, line 30: IMO both of these variable names are to long. What do you think of simply `viewLimit`?

CHANGELOG.md
```diff
@@ -482,7 +482,7 @@ BUG FIXES:
 
 BREAKING CHANGES:
 
-  * Retry now has a sane maximum default. Previous versions of Consul Template
+  * Retry now has a reasonable maximum default. Previous versions of Consul Template
```
eikenb, line 485: In line with my comment below... this use of 'reasonable' becomes optional but would still make sense.

angrycub, line 485:
```suggestion
  * Retry now has a maximum default. Previous versions of Consul Template
```

manager/runner.go
```diff
@@ -27,9 +27,9 @@ import (
...
+	// reasonableViewLimit is the number of views that we consider reasonable before we
```
angrycub, line 30:
```suggestion
	// viewLimit is the number of views that we consider reasonable before we
```

manager/runner.go
```diff
@@ -27,9 +27,9 @@ import (
...
 	// warn the user that they might be DDoSing their Consul cluster.
-	saneViewLimit = 128
+	reasonableViewLimit = 128
```
angrycub, line 32:
```suggestion
	viewLimit = 128
```

manager/runner.go
```diff
@@ -276,7 +276,7 @@ func (r *Runner) Start() {
 
 	for {
 		// Warn the user if they are watching too many dependencies.
-		if r.watcher.Size() > saneViewLimit {
+		if r.watcher.Size() > reasonableViewLimit {
```
angrycub, line 279:
```suggestion
		if r.watcher.Size() > viewLimit {
```

eikenb (approved):
Looks good. Thanks.

PR #1741, opened 2023-04-14
Add support for Vault agent environment variables | VAULT-16059

dhuckins:
The purpose of this PR is to support injecting vault secrets as environment variables for vault agent.

we needed a way to map the template (and `TemplateConfig`) back to the respective environment variable when a template is rendered


averche:
For context, this work is described in more details in the following RFC:

### [RFC: VLT-253](https://docs.google.com/document/d/1HLTkPtxUTtW-wMkePWoHmZoGt9lLg2POslt1ranaWE8/edit#)

config/template.go
```diff
@@ -118,6 +118,11 @@ type TemplateConfig struct {
 	// and causes an error if a relative path tries to traverse outside that
 	// prefix.
 	SandboxPath *string `mapstructure:"sandbox_path"`
+
+	// MapToEnvironmentVariable is the name of the environment variable this
+	// template should map to. This field is only applicable used when
```
eikenb, line 123: "[..] the environment variable this template should map to."

I'm not sure what you mean here by "map to". Would you mind clarifying this bit some so it is more obvious?

eikenb: Maybe take inspiration from the PR message? It makes more sense to me.

eikenb:
Looks fine so far. Needs tests and documentation updates though.

config/template.go
averche: Perhaps a longer comment is warranted here. When `MapToEnvironmentVariable` is specified, Vault Agent will interpret this template as an environment variable template and output its contents to an environment variable rather than a file. Essentially this field is mutually exclusive with `Destination *string`.

eikenb: Yes please. You explained it perfectly here... if you could distill that into a comment I (and other future readers) would greatly appreciate it. Thanks!

config/template.go
```diff
@@ -118,6 +118,13 @@ type TemplateConfig struct {
 	// and causes an error if a relative path tries to traverse outside that
 	// prefix.
 	SandboxPath *string `mapstructure:"sandbox_path"`
+
+	// MapToEnvironmentVariable is the name of the environment variable this
+	// template ahouls map to. It is currently only used by Vault Agent and
```
dhuckins, line 123:
```suggestion
	// template should map to. It is currently only used by Vault Agent and
```

eikenb (approved):
Looks good.

averche (approved):

PR #1762, opened 2023-06-10
Various code quality fixes/linting

thevilledev:
Minor fixes, including:

- Use a more descriptive `FileMode` in tests and in octal format to match `os.Filemode`
- Fix tag in `ClientUserAgent` field for `VaultConfig`
- Omit comparison to bool constants
- Remove unconditionally terminated loops from various tests
- Remove unused `printWarnings()` method from `VaultWriteQuery`
- Remove `once` parameter from `newWatcher()` as it's reachable through `config` parameter as well

Figured it would make sense to put it all into a single PR rather than multiple (almost) one-line changes :)

### DIFF
--- a/dependency/vault_write.go
+++ b/dependency/vault_write.go
@@ -146,12 +146,6 @@ func sha1Map(m map[string]interface{}) string {
 	return fmt.Sprintf("%.4x", h.Sum(nil))
 }
 
-func (d *VaultWriteQuery) printWarnings(warnings []string) {
-	for _, w := range warnings {
-		log.Printf("[WARN] %s: %s", d, w)
-	}
-}
-
 func (d *VaultWriteQuery) writeSecret(clients *ClientSet, opts *QueryOptions) (*api.Secret, error) {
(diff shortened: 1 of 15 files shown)


manager/runner.go
```diff
@@ -1398,7 +1398,7 @@ func NewClientSet(c *config.Config) (*dep.ClientSet, error) {
 }
 
 // newWatcher creates a new watcher.
-func newWatcher(c *config.Config, clients *dep.ClientSet, once bool) *watch.Watcher {
```
roncodingenthusiast, line 1401: good catch 👍🏾

dependency/vault_write.go
```diff
@@ -146,12 +146,6 @@ func sha1Map(m map[string]interface{}) string {
 	return fmt.Sprintf("%.4x", h.Sum(nil))
 }
 
-func (d *VaultWriteQuery) printWarnings(warnings []string) {
```
roncodingenthusiast, line 149: this isn't used but i wonder why it was added in the first place

PR #1768, opened 2023-06-26
chore: refactor deprecated Vault calls

thevilledev:
Minor refactor for Vault related functionality in the `dependency` package:

- Use Vault `NewLifetimeWatcher()` instead of `NewWatcher()` as it is deprecated ([see API client reference here](https://github.com/hashicorp/vault/blob/api/v1.9.2/api/lifetime_watcher.go#L185C6-L187)). For compatibility purposes, `RenewBehaviour` is the same as with the previous functionality.
- Use Vault `ReadRaw()` instead of `RawRequest()` as it is deprecated ([see API client reference here](https://github.com/hashicorp/vault/blob/api/v1.9.2/api/client.go#L1279-L1288)).
- For `VaultReadQuery` dependency, remove `opts` from being passed further down to other functions as it is and won't be used. Query parameters (like `version`) for Vault read calls are stored in a separate `queryValues` field. Other dependencies which implement the same `Fetch()` interface (like `FileQuery`) won't pass it further either if the parameter is not needed.
- Remove unused `stopChan()` function from `VaultPKIQuery`.

### DIFF
--- a/dependency/vault_read.go
+++ b/dependency/vault_read.go
@@ -82,7 +82,7 @@ func (d *VaultReadQuery) Fetch(clients *ClientSet, opts *QueryOptions,
 		}
 	}
 
-	err := d.fetchSecret(clients, opts)
+	err := d.fetchSecret(clients)
 	if err != nil {
 		return nil, nil, errors.Wrap(err, d.String())
 	}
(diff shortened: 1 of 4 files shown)


dependency/vault_read.go
```diff
@@ -140,7 +138,7 @@ func (d *VaultReadQuery) Type() Type {
 	return TypeVault
 }
 
-func (d *VaultReadQuery) readSecret(clients *ClientSet, opts *QueryOptions) (*api.Secret, error) {
```
roncodingenthusiast, line 143: good catch on opts not being used

roncodingenthusiast (approved):
LGTM! great work @thevilledev

PR #1782, opened 2023-08-02
Support for setting Vault CA from VAULT_CACERT_BYTES env

tomhjp:
Adds support to set the CA we should trust when communicating with Vault over TLS through a VAULT_CACERT_BYTES environment variable.

The Vault API package supports this since https://github.com/hashicorp/vault/pull/14753

However, without support in consul-template, we end up with weird compatibility. Vault Agent automatically supports consuming that environment variable through the API package for its auto-auth feature, but consul-template reads in TLS configuration differently so Vault Agent's templating doesn't currently work with VAULT_CACERT_BYTES.

See https://github.com/hashicorp/vault-k8s/pull/507 for an example of attempting to use VAULT_CACERT_BYTES in the Agent Injector.


VioletHynes (approved):
Looks great! Thanks for doing this :)

roncodingenthusiast (approved):
LGTM

tomhjp:
Thanks both!

PR #1794, opened 2023-08-23
Re-open: Implement systemd notification support after all templates have been rendered 

dyudin0821:
This PR reintroduces the functionality to notify systemd once consul-template has completed its initial run. It's worth noting that a similar feature request was made in the past [(Issue #1011)](https://github.com/hashicorp/consul-template/issues/1011), but was not accepted. However, after further consideration and understanding the evolving needs of the community, I believe it's time to revisit this feature.

Here are my reasons why it is important to add:

- As modern infrastructure continues to evolve, the integration between tools becomes increasingly crucial. The ability for consul-template to notify systemd can streamline service initialization, ensuring that dependent services only start once their configurations are ready.

- This feature will provide a more seamless experience for users who rely on systemd for process management. It will eliminate the need for workarounds, such as running consul-template -once.

- Other tools in the ecosystem have integrated with init systems like systemd. This shows a recognized need in the community for such integrations. By adding this functionality to consul-template, we ensure that it remains competitive and relevant in the ever-changing landscape of DevOps tools.

In conclusion, while I understand and respect the decisions made in the past, I believe that the current landscape and the needs of the community warrant a reconsideration of this feature. I hope we can have a constructive discussion around this and work towards the betterment of consul-template.


dyudin0821:
Hello @skpratt 
Any updates?

dyudin0821:
@skpratt This is a friendly reminder, I would really like you to consider this change

skpratt (approved):

PR #1817, opened 2023-10-05
Update to use CRT prepare workflow

sarahethompson:
This update, moves consul-template to use the prepare workflow. This workflow encapsulates several previous workflows, running jobs in parallel to reduce the artifact processing time. See [here](https://hashicorp.atlassian.net/wiki/spaces/RELENG/pages/2489712686/Dec+7th+2022+-+Introducing+the+new+Prepare+workflow) for more info .

See successful prepare run here - https://github.com/hashicorp/crt-workflows-common/actions/runs/6420465781

### DIFF
--- a/.release/ci.hcl
+++ b/.release/ci.hcl
@@ -33,146 +33,17 @@ event "build" {
   }
 }
 
-event "upload-dev" {
+event "prepare" {
   depends = ["build"]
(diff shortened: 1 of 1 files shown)


curtbushko (approved):
Thanks Sarah!

PR #1831, opened 2023-11-07
Setup for consul ent testing

roncodingenthusiast:
- allow us to run consul with enterprise binary for testing enterprise feature 

Screenshot 2023-11-08 at 1 40 51 PM

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,17 +1,26 @@
 name: ci
 
 on:
-  push:
   pull_request:
+  push:
+    branches:
+      - main
+      - release/**
+
+env:
+  CONSUL_LICENSE: ${{ secrets.CONSUL_LICENSE }}
 
 jobs:
   run-tests:
-    name: Run test cases
+    name: Run test cases (consul-version=${{ matrix.consul-version }})
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         os: [ubuntu-latest]
         go: [^1]
+        consul-version: 
+        - 1.17.0
+        - 1.17.0+ent
 
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -20,12 +26,22 @@ jobs:
...
+        run: |
+          CONSUL_VERSION="${{ matrix.consul-version }}"
+          FILENAME="consul_${CONSUL_VERSION}_linux_amd64.zip"
+          curl -sSLO "https://releases.hashicorp.com/consul/${CONSUL_VERSION}/${FILENAME}" && \
+            unzip "${FILENAME}" -d /usr/local/bin && \
+            rm "${FILENAME}"
+          consul version
 
-      - name: Install Consul, Vault and Nomad for integration testing
+      - name: Install Vault and Nomad for integration testing
```
roncodingenthusiast, line 40: removed consul from this list into a separate step

.github/workflows/ci.yml
```diff
@@ -2,16 +2,21 @@ name: ci
 
 on:
   push:
-  pull_request:
```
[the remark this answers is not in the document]
skpratt: I'd prefer to have this run on:

  pull_request:
  push:
    branches:
      # Push events on the main branch
      - main
      - release/**

.github/workflows/ci.yml
skpratt: we could move this install step below and instead just use "apt-get install consul${{ matrix.ent_tag}}", where ent_tag: "", "-enterprise", so we test the latest linux packages and don't have to keep coming back to bump the version

roncodingenthusiast: actually decided to do it this way because everytime there was a new consul version, the CI will start failing. 

Locking it to something specific allows us to be more intentional about when we upgrade consul in this repo rather than have to manually change the test everytime there is a new release. This also follows the same pattern as consul-ecs and a few other repos

roncodingenthusiast: Do you still think this should be changed? or is this fine like this?

kisunji (approved):

.github/workflows/ci.yml
skpratt: But is the test fragile enough to start failing on most version upgrades? I guess I would have bigger concerns about that. We didn't previously lock in the CE consul version, nor do we currently lock in the vault and nomad versions - so I would be interested to know more about why the new tests require it.

roncodingenthusiast: oh i was talking about existing test: https://github.com/search?q=repo%3Ahashicorp%2Fconsul-template%201.17.0&type=code but our chat on slack made sense so I will update this

skpratt (approved):

PR #1842, opened 2023-11-21
[NET-1484] Support for namespaces, partitions in consul endpoints

roncodingenthusiast:
## TODO 

- [x] KV list endpoint 
- [x] KV keys endpoint 
- [x] KV get endpoint 
- [x] Catalog node endpoint
- [x] Catalog nodes endpoint 
- [x] catalog service endpoint 
- [x] catalog services endpoint 
- [x] Docs 

## How to test 
Note: ensure you have consul running 
### previous behavior is preserved 
- insert into kv in default namespace and partition 
```
consul kv put "hashi/test" "default"
```
- create a file `in.tpl`  insert the following to retrieve that key 
```
{{ key "hashi/test" }}
```
- ensure it can still read the key
```
consul-template -template "in.tpl:out.txt" -once
```
- verify: your `out.tpl` file should have:
```
default
```

### specify namespace/partition 
- create a new consul namespace/partition (sorry these names aren't too creative :D) 
```
consul partition create -name "partition"
consul namespace create -name "partition-ns" -partition "partition"
```
- store kv in the new partition/namespace created above
```
consul kv put -partition "partition" -namespace "partition-ns" "hashi/test" "part
ition"
```
- update your `in.tpl` to include namespace and partition
```
{{ key "hashi/test?partition=partition&ns=partition-ns" }}
```
- verify: your `out.tpl` file should have:
```
partition
```

### DIFF
--- a/dependency/catalog_node.go
+++ b/dependency/catalog_node.go
@@ -19,7 +19,7 @@ var (
 	_ Dependency = (*CatalogNodeQuery)(nil)
 
 	// CatalogNodeQueryRe is the regular expression to use.
-	CatalogNodeQueryRe = regexp.MustCompile(`\A` + nodeNameRe + dcRe + `\z`)
+	CatalogNodeQueryRe = regexp.MustCompile(`\A` + nodeNameRe + queryRe + dcRe + `\z`)
 )
 
 func init() {
@@ -31,8 +31,10 @@ func init() {
 type CatalogNodeQuery struct {
 	stopCh chan struct{}
 
-	dc   string
-	name string
+	dc        string
+	name      string
+	namespace string
+	partition string
 }
 
 // CatalogNode is a wrapper around the node and its services.
@@ -60,10 +62,17 @@ func NewCatalogNodeQuery(s string) (*CatalogNodeQuery, error) {
 	}
 
 	m := regexpMatch(CatalogNodeQueryRe, s)
+	queryParams, err := GetConsulQueryOpts(m, "catalog.node")
+	if err != nil {
+		return nil, err
+	}
+
 	return &CatalogNodeQuery{
-		dc:     m["dc"],
-		name:   m["name"],
-		stopCh: make(chan struct{}, 1),
+		dc:        m["dc"],
+		name:      m["name"],
+		stopCh:    make(chan struct{}, 1),
+		namespace: queryParams.Get(QueryNamespace),
+		partition: queryParams.Get(QueryPartition),
 	}, nil
 }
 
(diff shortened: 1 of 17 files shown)


skpratt (approved):

komapa:
I understand this is "alpha" but this does not work at all with `range services` and different namespaces. You get the same services as the first `range services` call because (I am guessing), you did not update the cache key to consider any of the new QUERY params (namespace, partition, peer)

Also, this functionality is pretty useless unless there is also `range namespaces` implemented. Right now, we need to do something like:

```
{{- range $k, $n := (plugin "consul" "namespace" "list" "--format" "json" | parseJSON) -}}
```

jkirschner-hashicorp:
Hi @komapa,

It sounds like you have a specific scenario in mind that you need to work, but unfortunately doesn't work with the current implementation. To help get folks aligned on this, can you share that scenario and sample Consul Template syntax you'd hope would work for that scenario?

komapa:
We just want to get all services for all namespaces that we care about:

```
{{- $namespaces := sprig_list "default" -}}
{{- range $k, $n := (plugin "consul" "namespace" "list" "--format" "json" | parseJSON) -}}
    {{- /* TODO: Figure out how we can check for $n.Meta.external-source eq kubernetes, instead */ -}}
    {{ if eq $n.Description "Auto-generated by consul-k8s" }}
        {{- $namespaces = sprig_append $namespaces $n.Name -}}
    {{- end -}}
{{- end -}}

{{- range $k, $namespace := $namespaces }}
{{- range services (printf "?ns=%s&partition=default" $namespace) }}
{{ $namespace }}: {{ .Name }}
{{- end }}
{{- end }}
```

komapa:
> And based on this comment from before:
> 
> > Also, this functionality is pretty useless unless there is also range namespaces implemented.
> 
> I assume you actually want something like this?
> 
> ```
> {{- range namespaces "?partition=default"}
> {{- range services (printf "?ns=%s&partition=default" .NamespaceName) }}
> {{ $namespace }}: {{ .Name }}
> {{- end }}
> {{- end }}
> ```
> 
> In other words, is the below an accurate summary?
> 
> * **UX Challenge:** The top loop in your example is just to construct the list of namespaces to iterate over, which is unnecessary if `range namespaces` is implemented. This workaround is ugly, but _functional_.
> * **Bug:** However, there's a bug that no workaround can resolve: `range services` is always returning results from the first namespace in the array (index 0 in `range namespaces` loop).

Yes, this is perfect summary. Thank you!

jkirschner-hashicorp:
Since both Consul and Vault have namespaces, it might make sense to have it be `range consulNamespaces` or something like that with variable name `.ConsulNamespace` to make it unambiguous (in case iterating over a Vault namespace tree is ever added, or to just make it clear to the reader that we're talking about a Consul namespace not a Vault one).

jkirschner-hashicorp:
@komapa : The **Bug** described above should be fixed by this PR (to be included in the next Consul Template release): https://github.com/hashicorp/consul-template/pull/1874.

The **UX Challenge** will need to be addressed separately in the future.

Issue #1840, opened 2023-11-17
Consul template unlimitely spawn TCP connections which DDoS Consul/Vault


weichuliu:
We are using `vault agent` to render templates.

It looks like for each `secret` keyword in a template, consul-template spawns a new query to Vault, and all queries are spawned simultaneously. This makes rendering a simple template easily a DoS to the Vault.

For example, if consul-template renders a file with 10k `secret`s, it can be observed that the single process starts 10K concurrent tcp connections to Vault all at a time.

In our environment, 20 pods can easily bring Vault to 700K GoRoutines. Potentially it crashed the Vault quorum.

### Consul Template version

I am using `vault agent` with Vault version 1.15.0.

### Configuration

`config.hcl`:

```hcl
vault {
  address = "http://localhost:8200"
}
auto_auth {
  method {
    type = "token_file"
    config = {
      token_file_path = "/tmp/token" # root
    }
  }
}
log_level = "info"
exit_after_auth = true
template_config {
  exit_on_retry_failure = true
}
template {
  source      = "/tmp/test.gotmpl"
  destination = "/tmp/properties"
}
```

`/tmp/test.gotmpl` -- This simply list `secret/test-kv` path, and render all secrets to under the path to the output file:

```liquid
{{ range secrets "secret/metadata/test-kv" }}
{{ $i := . }}
{{     spew_printf "secret/test-kv/%s\n" $i }}
{{     scratch.MapSet "properties" . ( printf "secret/test-kv/%s" $i ) }}
{{ end }}
{{ range $key, $fullPath := scratch.Get "properties" -}}
{{   with secret $fullPath -}}
{{     if sprig_hasKey .Data.data "data" -}}
{{       $value := .Data.data.data -}}
{{         $key }}={{ $value | trimSpace | replaceAll "\n" "\\n" }}
{{     end -}}
{{   end -}}
{{ end -}}
```

`/tmp/token`: content is `root`.

### Command

1. Setup a dev vault

```shell
vault server -dev -dev-root-token-id=root
```

2. Write 10000 secrets to `test-kv`.

```
export VAULT_ADDR=http://127.0.0.1:8200 VAULT_TOKEN=root ; for i in {0..10000}; do vault kv put -mount secret test-kv/$i data=`date +%s`; done
```

3. I wrote a Python script to track the tcp connections during rendering:

```python
import subprocess
import time
import os


[...]

PR #1858, opened 2023-12-14
Add Transport config parameter for MaxConnsPerHost

ccapurso:
Rendering templates within Vault Agent that contain many secrets can result in aggressively opening a very large number of TCP connections all at once. Opening this many connections essentially results in a DoS against Vault. This PR adds the ability to specify a value for `MaxConnsPerHost` within a client's [Transport](https://pkg.go.dev/net/http#Transport) configuration. Support for this new field was only added for the Vault client.

Fixes https://github.com/hashicorp/consul-template/issues/1840


config/convert_test.go
```diff
@@ -216,7 +216,7 @@ func TestFileModePresent(t *testing.T) {
 		},
 		{
 			"present",
-			FileMode(0644),
+			FileMode(0o644),
```
lornasong, line 219: I wasn't familiar with this file mode. I looked it up and it seems like it is used for Python 3. Since this repo is in Go, I'm not sure if I'm missing something. Could you share some context around this change?

lornasong (commented):
Hey @ccapurso! These changes make sense to me.

I'm not super familiar with this code so did a review by comparing how a similar field(I chose `MaxIdleConnsPerHost`) is implemented. I noticed a couple additional changes with `MaxIdleConnsPerHost`:
 - [transport.go GoString()](https://github.com/hashicorp/consul-template/blob/main/config/transport.go#L187-L198) - it doesn't look like all transport fields are in the GoString. I'm not sure if it would be worthwhile to add `MaxConnsPerHost` here?
 - [config_test.go TestParse](https://github.com/hashicorp/consul-template/blob/main/config/config_test.go#L1430-L1445) - it looks like adding a test case for `MaxConnsPerHost` might keep parity?
 - [cli.go ParseFlags()](https://github.com/hashicorp/consul-template/blob/main/cli.go#L583-L586) - would it make sense to add flag support for this field?

Totally okay if`MaxConnsPerHost` doesn't make sense to be in the 3 places above, so just want to get your input.

Another question mostly because I'm not familiar with things, would you mind sharing a little bit about how you tested these changes? Appreciate it, thank you

config/convert_test.go
indjb: It's not just Python 3.  Since [Go 1.13](https://go.dev/doc/go1.13#language) both formats for octal literals [are allowed](https://go.dev/ref/spec#Integer_literals) and are equivalent.

ccapurso: This was not a direct change but instead a result of running `gofumpt`.

ccapurso:
@lornasong, thank you for the review! I have addressed the three spots you mentioned:
- [transport.go GoString()](https://github.com/hashicorp/consul-template/blob/main/config/transport.go#L187-L198)
- [config_test.go TestParse](https://github.com/hashicorp/consul-template/blob/main/config/config_test.go#L1430-L1445) - - Only added support for Vault at this time
- [cli.go ParseFlags()](https://github.com/hashicorp/consul-template/blob/main/cli.go#L583-L586) - Only added support for Vault at this time

I tested using the reproduction steps provided in https://github.com/hashicorp/consul-template/issues/1840 and ensured that the connections were limited based on the configured maximum.

lornasong (approved):
thanks @ccapurso! the changes make sense to me

PR #1874, opened 2024-02-06
[NET-7571]SeatGeek

kkavish:
- bug fix for SeatGeek: Services function in CT (consul-template) does not update the namespace for each services API call
- opts should be defaulted to the opts at query creation.
- then they should be merged with the opts present in Fetch.
- currently the merge was the other way around.
- ENT test support is not there, so no test cases added as of now.

### Checklist ###
- [ ] Tests added
- [x] CHANGELOG entry added

### DIFF
--- a/dependency/catalog_services.go
+++ b/dependency/catalog_services.go
@@ -71,12 +71,8 @@ func (d *CatalogServicesQuery) Fetch(clients *ClientSet, opts *QueryOptions) (in
 	default:
 	}
 
-	// this overrides the query params present in the query
-	// i.e. overrides the namespace and partition params with ones used the first time
-	// while creating NewCatalogServicesQuery
-
+	// this overrides the query params present in the query with ones present while creating the query
 	// see bug [https://github.com/hashicorp/consul-template/pull/1842#issuecomment-1915723565]
-	// it should be other way around.
 	// default to the query params present while creating NewCatalogServicesQuery
 	// and then merge with the query params present in the query
 	defaultOpts := &QueryOptions{
--- a/dependency/catalog_services.go
+++ b/dependency/catalog_services.go
@@ -71,11 +71,21 @@ func (d *CatalogServicesQuery) Fetch(clients *ClientSet, opts *QueryOptions) (in
 	default:
 	}
 
-	opts = opts.Merge(&QueryOptions{
+	// this overrides the query params present in the query
+	// i.e. overrides the namespace and partition params with ones used the first time
+	// while creating NewCatalogServicesQuery
+
+	// see bug [https://github.com/hashicorp/consul-template/pull/1842#issuecomment-1915723565]
+	// it should be other way around.
+	// default to the query params present while creating NewCatalogServicesQuery
+	// and then merge with the query params present in the query
+	defaultOpts := &QueryOptions{
 		Datacenter:      d.dc,
 		ConsulPartition: d.partition,
 		ConsulNamespace: d.namespace,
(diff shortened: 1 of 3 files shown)


dependency/catalog_services.go
```diff
@@ -71,11 +71,21 @@ func (d *CatalogServicesQuery) Fetch(clients *ClientSet, opts *QueryOptions) (in
...
 
-	opts = opts.Merge(&QueryOptions{
+	// this overrides the query params present in the query
+	// i.e. overrides the namespace and partition params with ones used the first time
+	// while creating NewCatalogServicesQuery
+
+	// see bug [https://github.com/hashicorp/consul-template/pull/1842#issuecomment-1915723565]
+	// it should be other way around.
+	// default to the query params present while creating NewCatalogServicesQuery
+	// and then merge with the query params present in the query
```
tauhid621, line 81: Should we have this comment here as the updated code will not have reference of what was used earlier?

tauhid621 (approved):
Looks good. I see that other [functions](https://github.com/hashicorp/consul-template/blob/9bd3d8e23c7d564c6a6fde60599aeb20a30e9de6/dependency/catalog_nodes.go#L80) are using the ` opts.Merge(&QueryOptions{` way. Would we need to update some of them as well?

dependency/catalog_services.go
kkavish: yes, some of the other functions have that problem too.
that's why I want to enable ENT and write proper tests for them before touching them.

roncodingenthusiast (approved):

komapa:
Unfortunately this is also not complete implementation. I tested it locally and you need to change:

```
// String returns the human-friendly version of this dependency.
func (d *CatalogServicesQuery) String() string {
	if d.dc != "" {
		return fmt.Sprintf("catalog.services(@%s)", d.dc)
	}
	return "catalog.services"
}
```
to

```
// String returns the human-friendly version of this dependency.
func (d *CatalogServicesQuery) String() string {
	opts := &QueryOptions{
		Datacenter:      d.dc,
		ConsulPartition: d.partition,
		ConsulNamespace: d.namespace,
	}

	if opts.String() != "" {
		return fmt.Sprintf("catalog.services(@%s)", opts.String())
	}

	return "catalog.services"
}
```

This way you are actually watching different `catalog.services` for every `dc`, `partition` and `namespace` combination.

kkavish:
> Unfortunately this is also not complete implementation. I tested it locally and you need to change:
> 
> ```
> // String returns the human-friendly version of this dependency.
> func (d *CatalogServicesQuery) String() string {
> 	if d.dc != "" {
> 		return fmt.Sprintf("catalog.services(@%s)", d.dc)
> 	}
> 	return "catalog.services"
> }
> ```
> 
> to
> 
> ```
> // String returns the human-friendly version of this dependency.
> func (d *CatalogServicesQuery) String() string {
> 	opts := &QueryOptions{
> 		Datacenter:      d.dc,
> 		ConsulPartition: d.partition,
> 		ConsulNamespace: d.namespace,
> 	}
> 
> 	if opts.String() != "" {
> 		return fmt.Sprintf("catalog.services(@%s)", opts.String())
> 	}
> 
> 	return "catalog.services"
> }
> ```
> 
> This way you are actually watching different `catalog.services` for every `dc`, `partition` and `namespace` combination.

@komapa thanks for pointing that out. please see  - https://github.com/hashicorp/consul-template/pull/1880

PR #1876, opened 2024-02-08
render templates with external functions

tgross:
Some consumers of `consul-template` use it like a library, where the application runs the runner in-process. For projects like Nomad which need to run with a high level of privilege, this is problematic in that its challenging to secure the operations that read and write from disk without running the entirety of CT as an external process (which carries a lot of overhead for Nomad workloads).

Add a `RendererFunc` and `ReaderFunc` interface to allow Nomad to inject a sandboxed subprocess when reading from disk and writing to disk.

This implementation is currently being used in Nomad 1.7.4, 1.6.7, and 1.5.14 as a mitigation for https://github.com/hashicorp/nomad/issues/19888. See https://github.com/hashicorp/nomad/commit/df865033494574de14da687c999d66b799899b21 for example usage.

Required for https://github.com/hashicorp/nomad/issues/19919

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -8,7 +8,6 @@ import (
 	"crypto/md5"
 	"encoding/hex"
 	"fmt"
-	"os"
 	"strings"
 	"text/template"
 
@@ -117,6 +116,9 @@ type NewTemplateInput struct {
 
 	// Config keeps local reference to config struct
 	Config *config.TemplateConfig
+
+	// ReaderFunc is called to read in any source file
+	ReaderFunc config.Reader
 }
 
 // NewTemplate creates and parses a new Consul Template template at the given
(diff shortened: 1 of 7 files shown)


template/template.go
```diff
@@ -154,7 +156,7 @@ func NewTemplate(i *NewTemplateInput) (*Template, error) {
 	}
 
 	if i.Source != "" {
-		contents, err := os.ReadFile(i.Source)
```
lkysow, line 157: Should we add validation that `i.ReaderFunc` is set so we don't panic?

lkysow (approved):

tgross:
@lkysow I've addressed your comments and rebased on main. I can squash merge this if you're happy with the results?

tgross:
Oh, well there's something wrong with the infra being set up for the Consul ENT test... like Vault isn't accepting writes for some reason during the setup? I see there was a recent set of commits by @kkavish trying to de-flake tests but not sure if this is new or not?

Issue #1789, opened 2023-08-07
Consul Templates Fail to Render when using KV2 secrets with deletion time


Jasonrchoi:
Using the Vault Agent Injector in EKS Cluster A addressed to an external vault in EKS Cluster B, I have a test pod with the following annotations. The `integration-app-test` role was given a policy that grants access to both secret paths. 

### Command

```shell
        vault.hashicorp.com/agent-inject-secret-credentials.txt: secret/staging/app/secure #Generic backend
        vault.hashicorp.com/agent-inject-secret-secure: secretv2/data/application/staging/app #KV2 backend
        vault.hashicorp.com/agent-inject-template-secure: |
          {{ with secret "secretv2/data/application/staging/app" -}}
            {{ range $k, $v := .Data.metadata }}
              export {{ $k | toUpper }}={{ $v }}
            {{ end }}
          {{- end }}
        vault.hashicorp.com/auth-path: auth/kubernetes_integration
        vault.hashicorp.com/log-level: trace
        vault.hashicorp.com/role: integration-app-test

note `secret/` refers to a generic secrets backend, `secretv2/` refers to a 'kv2' backend
```

### Expected behavior

What should have happened? 

Secrets from both backends were rendered into the pod.

### Actual behavior

We found that the "generic" secrets are rendered correctly, while the "kv2" secrets receive a `no secret exists at {PATH}` error and then eventually, `Permission Denied.` Because we were able to access secrets from our generic secret backend, it appeared we could rule out most auth/network issues.

### Steps to reproduce

We think we discovered the cause which is: https://github.com/hashicorp/consul-template/blob/e1484d3c8e65b7ff30b1f840395f072d6b1b424e/dependency/vault_read.go#L178

In the `DeleteKV2v2` method, any `Deletion_Time` associated with the secret, other than null, will cause the method to return true, causing the Read to throw an error that the secret does not exist, even if it clearly does. Once we got rid of deletion time for our secrets, the kv2 secret was rendered correctly by the Vault Agent Injector.

### References
It appears this issue was alluded to in `vault-k8s` but never resolved. 

Are there any other GitHub issues (open or closed) that should
be linked here? For example:
- https://github.com/hashicorp/vault-k8s/issues/176

Issue #1778, opened 2023-07-10
Consul-template not returning Vault kv secrets with future deletion_time set.


marknl:
### Description
When a secret in Vault has a `deletion_time` set, consul-template will not fetch it anymore. This is good for date/time entries that are in the past, but you can also set future date/time values of which I think should still be accessible.

### Consul Template version
Using the consul-template implementation built into Nomad 1.5.6. I'm not really sure which version of consul-template is baked into nomad 1.5.6

### Configuration
```liquid
template {
  data = <<EOD
  {{ with secret "secret/data/email" }}{{ .Data.data.address }}{{ end }}
  EOD
  destination = "email.txt"
}            
```

### Expected behavior
I'm expecting the value of the Vault secret email (which is an email address) to be saved into the file `email.txt`.

### Actual behavior
Nomad throws an error when trying to start the job, saying it cannot find the secret.

### Steps to reproduce

1. `vault server -dev`
2. `vault kv metadata put -mount=secret -delete-version-after=1h email`
3. `vault kv put -mount=secret email address=foo@example.com`
4. Create a consul-template template that will fetch the secret from Vault and save it to a file.
5. Run the consul-template template.
6. consul-template will not find the secret because deletion_time is not empty.

### References
The code that causes this is: https://github.com/hashicorp/consul-template/blame/main/dependency/vault_read.go#L178

I've opened up a discuss thread on the Hashicorp forums a few days ago, but I get the feeling this is a bug, not a feature ;-)
https://discuss.hashicorp.com/t/consul-template-not-returning-vault-kv-items-with-deletion-time-set/55763

PR #1879, opened 2024-02-12
VAULT-528 Fix issue with consul-template not rendering secrets with delete_version_after set on them

VioletHynes:
Some background:

Initially, Vault secrets would only have `deletion_time` set on them if they'd been deleted. Since then, it has become possible to set deletion time in the future, using `delete_version_after` to indicate that a secret 'will' be deleted. As a result, the old logic to assume a secret is deleted if it has a deletion time has become naive.

Resolves https://github.com/hashicorp/consul-template/issues/1789
Resolves https://github.com/hashicorp/consul-template/issues/1778
Resolves https://discuss.hashicorp.com/t/consul-template-not-returning-vault-kv-items-with-deletion-time-set/55763

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ NEW FEATURES:
 
 BUG FIXES:
 * Fetch services query not overriding opts correctly [NET-7571](https://hashicorp.atlassian.net/browse/NET-7571)
+* Consul-template now correctly renders KVv2 secrets with `delete_version_after` set [NET-3777](https://hashicorp.atlassian.net/browse/NET-3777)
 
 ## v0.36.0 (January 3, 2024)
 
--- a/dependency/vault_read.go
+++ b/dependency/vault_read.go
@@ -176,7 +176,20 @@ func (d *VaultReadQuery) readSecret(clients *ClientSet) (*api.Secret, error) {
 func deletedKVv2(s *api.Secret) bool {
 	switch md := s.Data["metadata"].(type) {
 	case map[string]interface{}:
-		return md["deletion_time"] != ""
+		deletionTime, ok := md["deletion_time"].(string)
+		if !ok {
+			// Not a string, end early
+			return false
+		}
+		t, err := time.Parse(time.RFC3339, deletionTime)
+		if err != nil {
+			// Deletion time is either empty, or not a valid string.
+			return false
+		}
+
+		// If now is 'after' the deletion time, then the secret
+		// should be deleted.
+		return time.Now().After(t)
 	}
 	return false
 }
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ NEW FEATURES:
 
 BUG FIXES:
 * Fetch services query not overriding opts correctly [NET-7571](https://hashicorp.atlassian.net/browse/NET-7571)
+* Consul-template now correctly renders KVv2 secrets with `delete_version_after` set [NET-3777](https://hashicorp.atlassian.net/browse/NET-3777)
```
VioletHynes, line 8: I spotted this as part of researching the bug. I've been doing the work under VAULT-528, but I'm happy to close NET-3777 as a duplicate or something once I merge this. However you folks want to handle it and whatever best fits your process!

dependency/vault_read.go
```diff
@@ -176,7 +176,20 @@ func (d *VaultReadQuery) readSecret(clients *ClientSet) (*api.Secret, error) {
 func deletedKVv2(s *api.Secret) bool {
 	switch md := s.Data["metadata"].(type) {
 	case map[string]interface{}:
-		return md["deletion_time"] != ""
+		deletionTime, ok := md["deletion_time"].(string)
+		if !ok {
+			// Not a string, end early
```
kkavish, line 181:
```suggestion
			// key not present, end early
```

VioletHynes: Since we're casting to a string, we'll only go down the `!ok` codepath will only be gone down if the key isn't present or the value isn't a string. I'll improve the comment to show that it could be both :)

VioletHynes: Updated! Please let me know if there's anything else you'd need before approving :)

CHANGELOG.md
kkavish: Please do close NET-3777, this essentially fixes what has been reported.

kkavish (approved):

dependency/vault_read_test.go
```diff
@@ -711,3 +711,63 @@ func TestShimKVv2Path(t *testing.T) {
...
+			},
+		},
+	}))
+	assert.False(t, deletedKVv2(&api.Secret{
+		Data: map[string]interface{}{
+			"metadata": map[string]interface{}{
+				"deletion_time": "foo",
+			},
+		},
+	}))
```
marcboudreau, line 772: I'd add two more tests that confirm:
1. that if metadata is missing, `false` is returned from **deletedKVv2**
2. that if metadata's type is something other than `map[string]interface{}`, `false` is returned from **deletedKVv2**

marcboudreau (approved):
Looks good. I did leave a suggestion for 2 additional test cases for your consideration.

dependency/vault_read_test.go
VioletHynes: Thank you! Added both :)

PR #1880, opened 2024-02-13
Net 7571

kkavish:
- amending string conversion so that map key could also have dc+ns+partition while tracking the dependency.

### DIFF
--- a/dependency/catalog_services.go
+++ b/dependency/catalog_services.go
@@ -120,10 +120,22 @@ func (d *CatalogServicesQuery) CanShare() bool {
 
 // String returns the human-friendly version of this dependency.
 func (d *CatalogServicesQuery) String() string {
+	var name string
 	if d.dc != "" {
-		return fmt.Sprintf("catalog.services(@%s)", d.dc)
+		name = name + "@" + d.dc
 	}
-	return "catalog.services"
+	if d.namespace != "" {
+		name = name + "@" + d.namespace
+	}
+	if d.partition != "" {
+		name = name + "@" + d.partition
+	}
+
+	if len(name) == 0 {
+		return "catalog.services"
+	}
+
+	return fmt.Sprintf("catalog.services(%s)", name)
 }
 
 // Stop halts the dependency's fetch function.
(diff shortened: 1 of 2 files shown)


dependency/catalog_services.go
```diff
@@ -120,10 +118,22 @@ func (d *CatalogServicesQuery) CanShare() bool {
...
 // String returns the human-friendly version of this dependency.
 func (d *CatalogServicesQuery) String() string {
+	var name string
 	if d.dc != "" {
-		return fmt.Sprintf("catalog.services(@%s)", d.dc)
+		name = name + "@" + d.dc
 	}
-	return "catalog.services"
+	if d.namespace != "" {
+		name = name + "@" + d.namespace
```
tauhid621, line 126: Wondering if we also need to add some info that this is the namespace?

Eg: Two string would be same if they have namespace and partition as same.

Eg: catalog.services(@foo), catalog.services(@foo). The first foo is namespace while the second is partition.

Maybe having the format `catalog.services(ns={namespace},partition={part}@dc)` would be better?

This would not be needed if we do not worry about the strings to be same for two queries having the same value of partition and namespace.

dependency/catalog_services.go
```diff
@@ -120,10 +118,22 @@ func (d *CatalogServicesQuery) CanShare() bool {
 
 // String returns the human-friendly version of this dependency.
 func (d *CatalogServicesQuery) String() string {
+	var name string
 	if d.dc != "" {
-		return fmt.Sprintf("catalog.services(@%s)", d.dc)
+		name = name + "@" + d.dc
```
tauhid621, line 123: NIT: Better to use fmt.Sprintf here and the next few places.
```suggestion
		name = fmt.Sprintf("%s@%s", name, d.dc)
```

tauhid621 (commented):
Left a minor comment

dependency/catalog_services.go
kkavish: valid point, I don't have much context on whether someone would actually have that kind of deployment but the point is THEY CAN have such a deployment.

I guess it's better to plan for the above case. It'd be prudent that way, we'd need to change it for every dependency except the `list.peering` one. (which can be taken care of in subsequent PRs)

dependency/catalog_services.go
kkavish: `fmt.Sprintf` is generally used when you need specific format, here we're just concatenating and using `fmt.Sprintf` would be overkill (same pattern in org's codes).

tauhid621 (approved):

PR #1882, opened 2024-02-18
[NET-7377] Consul ENT tests

kkavish:
## Changes proposed in this PR:
- CT ENT tests for all consul resource supporting partition and ns.
- bug fix for map keys, keys did not contain partition and ns, so could have been overridden if same key in a different ns and partition is added as a dependency.

## How I've tested this PR:

- CI
- Locally (you can install consul ent and provide a license through env variable before running each test)

## Checklist:
- [X] Tests added
- [X] CHANGELOG entry added


dependency/catalog_node_test.go
```diff
@@ -103,67 +128,248 @@ func TestNewCatalogNodeQuery(t *testing.T) {
...
+			testCase{
+				tenancyHelper.AppendTenancyInfo("local", tenancy),
+				"",
+				&CatalogNode{
+					Node: &Node{
+						Node:            testConsul.Config.NodeName,
+						Address:         testConsul.Config.Bind,
+						Datacenter:      "dc1",
+						TaggedAddresses: map[string]string{
+							//"lan": "127.0.0.1",
```
tauhid621, line 147: Is this intended?

dependency/catalog_nodes_test.go
```diff
@@ -134,32 +139,90 @@ func TestNewCatalogNodesQuery(t *testing.T) {
...
+			testCase{
+				tenancyHelper.AppendTenancyInfo("partition and namespace", tenancy),
+				fmt.Sprintf("?partition=%s&ns=%s@dc1", tenancy.Partition, tenancy.Namespace),
+				[]*Node{
+					{
+						Node:            testConsul.Config.NodeName,
+						Address:         testConsul.Config.Bind,
+						Datacenter:      "dc1",
+						TaggedAddresses: map[string]string{
+							//"lan": "127.0.0.1",
```
tauhid621, line 176: Same comment as above

test/tenancy_helper.go
```diff
@@ -0,0 +1,149 @@
...
+			cases = append(cases, generationFunc(tenancy)...)
+		}
+	}
+	return cases
+}
+
+func (t *TenancyHelper) GetUniquePartitions() map[*api.Partition]interface{} {
+	partitions := make(map[*api.Partition]interface{})
+	for _, tenancy := range t.TestTenancies() {
+		partitions[&api.Partition{
```
tauhid621, line 144: Would it return unique partitions? I believe it would lead to duplicates as the map key is pointer.

test/tenancy_helper.go
```diff
@@ -0,0 +1,149 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package test
```
tauhid621, line 4: Would be nice to have some test cases for this file.

tauhid621 (commented):
Looks good overall. Left some minor comments

dependency/catalog_node_test.go
kkavish: yes, these are making the tests flaky.
They work with one consul version and not with another one.
I've investigated them and have taken a side note to check whether the problem is CT or Consul itself.

test/tenancy_helper.go
kkavish: yes sir, you're correct.

test/tenancy_helper.go
kkavish: Done

tauhid621 (approved):

dependency/catalog_service.go
```diff
@@ -151,6 +151,12 @@ func (d *CatalogServiceQuery) String() string {
 	if d.dc != "" {
 		name = name + "@" + d.dc
 	}
+	if d.partition != "" {
+		name = name + "@partition=" + d.partition
+	}
+	if d.namespace != "" {
+		name = name + "@ns=" + d.namespace
+	}
```
lkysow, line 159: @kkavish good find, I think we need to add `d.peer` as well?

PR #1886, opened 2024-02-23
Flaky ENT Tests fix

kkavish:
- fixing flaky tests, yay!
- test using CI and local runs, you need to have an ENT build in path and a valid license set as environment variable.


dependency/kv_list_test.go
```diff
@@ -299,17 +299,17 @@ func TestKVListQuery_Fetch(t *testing.T) {
 					{
 						Path:  "test-kv-list/prefix/foo",
 						Key:   "foo",
-						Value: fmt.Sprintf("bar-%s-%s", tenancy.Partition, tenancy.Namespace),
+						Value: "bar-default-default",
```
tauhid621, line 302: Is this change needed?

dependency/catalog_nodes_test.go
```diff
@@ -239,6 +263,18 @@ func TestCatalogNodesQuery_Fetch(t *testing.T) {
 							//"consul-network-segment": "",
 						},
 					},
+					{
+						Node:            "node" + tenancy.Partition,
+						Address:         testConsul.Config.Bind,
+						Datacenter:      "dc1",
+						TaggedAddresses: map[string]string{
+							//"lan": "127.0.0.1",
```
tauhid621, line 271: Since this is commented.  I feel that we should remove it.

dependency/dependency_test.go
```diff
@@ -241,6 +236,7 @@ func (c *ClientSet) createConsulTestResources() error {
 		if err := testClients.createConsulPeerings(tenancy); err != nil {
 			return err
 		}
+		time.Sleep(200 * time.Millisecond)
```
tauhid621, line 239: Is this needed? The creation of resources should be in sync right?

tauhid621 (approved):
Left minor comments

dependency/dependency_test.go
kkavish: this is to make sure that consul reaches a state of entropy.
the falky tests were due to wrong way of creating resource but the consul servers were not reaching a state of entropy and thus were not evicting the resources sometimes.

dependency/catalog_nodes_test.go
kkavish: we could, but I want to keep it to investigate why it's not working intermittently.

dependency/dependency_test.go
lkysow: If you changed all the registrations so they're not to the local agent but to a fake node then agent entropy won't matter so there should be no need for this sleep

kkavish: noted, I had it in place so that if someone accidentally writes to the local agent, the tests would keep on failing, without getting successful intermittently.

What do you suggest?

PR #1897, opened 2024-03-18
Create ServerErrCh to surface server errors back to caller.

divyaac:
Creates a channel, ServerErrCh that will surface server errors back to the caller. This channel will hold a maximum of one error at a time. 

The purpose of this, is to allow Vault Agent and Vault Proxy to re-authenticate if the error message from the server indicates that the token is an "invalid token". More information on this initiative here: https://hermes.hashicorp.services/document/1yTtWJEvyoYLPkYRQ3lxs6dHmqAVyC177BMEBEXuu8fE?draft=false

### DIFF
--- a/manager/runner.go
+++ b/manager/runner.go
@@ -23,7 +23,7 @@ import (
 	"github.com/hashicorp/consul-template/template"
 	"github.com/hashicorp/consul-template/watch"
 
-	multierror "github.com/hashicorp/go-multierror"
+	"github.com/hashicorp/go-multierror"
 )
 
 const (
@@ -37,6 +37,10 @@ type Runner struct {
 	// ErrCh and DoneCh are channels where errors and finish notifications occur.
 	ErrCh  chan error
 	DoneCh chan struct{}
+
+	// ServerCh is a channel to surface error responses from the server up the calling stack
+	// and will only hold a maximum of one error at a time
+	ServerCh chan error
 
 	// config is the Config that created this Runner. It is used internally to
 	// construct other objects and pass data.
@@ -199,6 +203,7 @@ func NewRunner(config *config.Config, dry bool) (*Runner, error) {
 	runner := &Runner{
 		ErrCh:         make(chan error),
 		DoneCh:        make(chan struct{}),
+		ServerCh:      make(chan error),
 		config:        config,
 		dry:           dry,
 		inStream:      os.Stdin,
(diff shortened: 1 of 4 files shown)


watch/watcher.go
```diff
@@ -121,6 +125,15 @@ func (w *Watcher) ErrCh() <-chan error {
 	return w.errCh
 }
 
+// ServerCh returns a read-only channel of errors returned by the server
+// as a response to each Consul instance
```
VioletHynes, line 129:
```suggestion
// as a response to each consul-template instance
```

manager/runner.go
```diff
@@ -38,6 +38,10 @@ type Runner struct {
 	ErrCh  chan error
 	DoneCh chan struct{}
 
+	// ServerCh is a channel to surface error responses from the server up the calling stack
+	// and will only hold a maximum of one error at a time
+	ServerCh chan error
```
VioletHynes, line 43: This is a little bit of a nit, but `ServerCh` doesn't immediately indicate to me that this will contain errors, and it has me asking myself: what's the difference between `ErrCh` and `ServerCh`? Would something like `TemplateErrCh` be more clear? Maybe I'm not fully understanding the difference

watch/view_test.go
```diff
@@ -113,6 +126,8 @@ func TestPoll_once(t *testing.T) {
 		t.Errorf("expected no data (should have stopped), but received view data")
 	case err := <-errCh:
 		t.Errorf("error while polling: %s", err)
+	case err := <-serverErrCh:
+		t.Errorf("server error while polling : %s", err)
```
VioletHynes, line 130:
```suggestion
		t.Errorf("server error while polling: %s", err)
```

VioletHynes (approved):
Looks good! I have some minor comments, but none are blocking.

A test would be great, but I think it would be difficult to test this in consul-template directly and we should be able to have plenty of Agent tests around this, so I don't see that as blocking either!

manager/runner.go
divyaac: This is a good question. I wanted to try to distinguish between any template error (for example it could be a lookup error) and an error that is coming directly from the server it is making a request to. 
TemplateServerErr seemed weirder because it could imply that the template server had an issue (especially when it is being referenced in the Agent code). 

If you think one of these is better though I'm open to changing it.

divyaac: Actually - I'll change this to ServerErrCh !

VioletHynes: Hmm, that's a great point. I'm okay with keeping it as-is, but I still worry it could be confusing between the template server and the Vault server. Though I think any name we pick will probably be imperfect

Some more options to consider:
- `VaultRequestErrCh`
- `RequestErrCh`
- `TemplateRenderingErrCh`

I'm happy to keep it as is, like I've said, and naming things is hard, but if there's something clearer we can think of I think that'd be great. If you like any other option I'd say go for it, otherwise, it's fine as-is :D

VioletHynes: I like `ServerErrCh` more than `ServerCh`, thanks for humouring me!

PR #1898, opened 2024-03-19
prevent dependencies from being added after stop

tgross:
A Nomad user reported an issue where `View.poll` goroutines were being leaked when using large templates. When the runner's `Run` method is in-flight and the runner concurrently receives `Stop`, it's possible for the watcher to stop dependencies via stop and then immediately recreate them. Because the `Stop` method has returned, the caller will assume that all goroutines have been stopped, and this causes the `View.poll` goroutines to be leaked. The race condition for this case is very narrow, but is most likely to occur when the user has both a very large number of dependencies, short secret leases, and the `Wait` configuration is nil.

Add a `stopped` flag to the `Watcher` which is set on `Stop`, and check whether this field has been set before adding a new dependency.

Ref: https://github.com/hashicorp/nomad/issues/20163

### DIFF
--- a/manager/runner_test.go
+++ b/manager/runner_test.go
@@ -1295,7 +1295,7 @@ func TestRunner_stoppedWatcher(t *testing.T) {
 	t.Cleanup(stopNow)
 
 	// Test server to simulate Vault cluster responses.
-	reqNum := atomic.Uintptr{}
+	reqNum := atomic.Uint64{}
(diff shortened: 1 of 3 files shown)


tgross:
Looks like there's some cross-test pollution to figure out here. Working on detangling that now and will mark ready-for-review once I have it fixed.

shoenig (approved):
LGTM!

PR #1899, opened 2024-03-19
NET-5879 - consul-template support for sameness groups

jmurret:
consul 1.19 will allow passing the `sameness-group` param to the `/health` endpoints.  This PR allows consul template to utilize that functionality.


dependency/dependency.go
```diff
@@ -64,18 +64,19 @@ type ServiceTags []string
...
+	Near                string
+	Choose              string
+	RequireConsistent   bool
+	VaultGrace          time.Duration
+	WaitIndex           uint64
+	WaitTime            time.Duration
+	ConsulPeer          string
+	ConsulPartition     string
+	ConsulNamespace     string
+	ConsulSamenessGroup string
```
jmurret, line 79: Need to add to Merge() as well.

.golangci.yml
```diff
@@ -88,5 +89,4 @@ linters-settings:
 
 run:
   timeout: 10m
-  concurrency: 4
-  skip-dirs-use-default: false
```
jmurret, line 92: lint error that this is deprecated and to use`issues.exclude-dirs-use-default` instead.

dependency/vault_pki.go
```diff
@@ -133,7 +133,7 @@ func goodFor(cert *x509.Certificate) (time.Duration, bool) {
 
 	// after we have the 'time the certificate should be rotated', figure out how
 	// far it is from now to sleep
-	sleepFor := time.Duration(rotationTime.Sub(now))
```
jmurret, line 136: linter error for unneeded conversion

dependency/dependency_test.go
```diff
@@ -239,6 +229,34 @@ func (c *ClientSet) createConsulTestResources() error {
...
+		Partition: failoverPartition,
+	})
+	sg := &api.SamenessGroupConfigEntry{
+		Kind:               QuerySamenessGroup,
+		Name:               name,
+		Partition:          partition,
+		DefaultForFailover: true,
+		Members:            members,
+	}
+	log.Printf("sameness group: %+v", sg)
```
lkysow, line 245: these log lines needed?

dependency/dependency_test.go
```diff
@@ -239,6 +229,34 @@ func (c *ClientSet) createConsulTestResources() error {
...
+	_, _, err := c.consul.client.ConfigEntries().Set(sg, &api.WriteOptions{})
+	if err != nil {
+		return err
+	}
+
+	sgs, _, err := c.consul.client.ConfigEntries().List(QuerySamenessGroup, &api.QueryOptions{})
```
lkysow, line 251: why do we need to list them at the end here?

dependency/dependency_test.go
```diff
@@ -545,17 +563,25 @@ func (v *nomadServer) DeleteVariable(path string, opts *nomadapi.WriteOptions) e
 func (c *ClientSet) createConsulPartitions() error {
 	for p := range tenancyHelper.GetUniquePartitions() {
 		if p.Name != "" && p.Name != "default" {
-			partition := &api.Partition{Name: p.Name}
-			_, _, err := c.Consul().Partitions().Create(context.Background(), partition, nil)
-			if err != nil {
-				return err
+			err2 := c.createConsulPartition(p.Name)
```
lkysow, line 566: nit (can ignore): why err2?

dependency/health_service.go
```diff
@@ -122,18 +124,25 @@ func healthServiceQuery(s string, connect bool) (*HealthServiceQuery, error) {
...
+		connect:       connect,
+		namespace:     queryParams.Get(QueryNamespace),
+		peer:          queryParams.Get(QueryPeer),
+		partition:     queryParams.Get(QueryPartition),
+		samenessGroup: queryParams.Get(QuerySamenessGroup),
+	}
+
+	if qry.samenessGroup != "" && qry.peer != "" {
+		return nil, fmt.Errorf("health.service: cannot specify both peer and sameness-group")
+	}
```
lkysow, line 143: nit/personal preference/feel free to ignore: to me it makes more sense to do the validation _before_ constructing the object.

dependency/health_service_test.go
```diff
@@ -230,9 +248,7 @@ func TestNewHealthServiceQuery(t *testing.T) {
 		tc := test.(testCase)
 		t.Run(fmt.Sprintf("%d_%s", i, tc.name), func(t *testing.T) {
 			act, err := NewHealthServiceQuery(tc.i)
-			if (err != nil) != tc.err {
-				t.Fatal(err)
-			}
+			assert.Equal(t, tc.err, err)
```
lkysow, line 251: thanks for making this change, makes a lot of sense to validate the actual error text

docs/templating-language.md
```diff
@@ -775,12 +775,16 @@ Query [Consul][consul] for services based on their health.
 
 The `<TAG>` attribute is optional; if omitted, all nodes will be queried.
 
-The `<QUERY>` attribute is optional; if omitted, the `default` Consul namespace, `default` partition will be queried. `<QUERY>` can be used to set the Consul [namespace](https://developer.hashicorp.com/consul/api-docs/health#ns-2), partition, or [peer](https://developer.hashicorp.com/consul/api-docs/health#peer). `<QUERY>` accepts a url query-parameter format, e.g.:
```
lkysow, line 778: Should we document what will be returned with sameness group and how it will behave with failures and updates to the config entry?

dependency/health_service_test.go
```diff
@@ -675,6 +691,136 @@ func TestHealthServiceQuery_Fetch(t *testing.T) {
...
+	partitionTwo := "sg-partition-2"
+	nodeOne := "node" + partitionOne
+	nodeTwo := "node" + partitionTwo
+	samenessGroup := "test-sameness-group"
+
+	require.NoError(t, testClients.createConsulPartition(partitionOne))
+	require.NoError(t, testClients.createConsulPartition(partitionTwo))
+	require.NoError(t, testClients.createConsulSamenessGroups(samenessGroup, partitionOne, partitionTwo))
+
+	// Register services in foo partition as critical so they will fall back to default partition service
```
lkysow, line 711: This comment is misleading I think? There's no foo partition in this test and we don't fallback to the default partition IIUC.

dependency/health_service_test.go
```diff
@@ -675,6 +691,136 @@ func TestHealthServiceQuery_Fetch(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	svcs, meta, err := d.Fetch(testClients, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	healthServices := svcs.([]*HealthService)
```
lkysow, line 761:
```suggestion
        // Assert that the service instance returned is from the first listed partition.
        // This is expected because the service is healthy so it shouldn't failover
        // to other partitions.
	healthServices := svcs.([]*HealthService)
```

dependency/health_service_test.go
```diff
@@ -675,6 +691,136 @@ func TestHealthServiceQuery_Fetch(t *testing.T) {
...
+				v.NodeTaggedAddresses = filterAddresses(
+					v.NodeTaggedAddresses)
+				v.NodeMeta = filterVersionMeta(v.NodeMeta)
+			}
+		}
+
+		// Assert - verify the results
+		expectedResult := []*HealthService{
+			{
+				Node:                nodeTwo,
```
lkysow, line 803:
```suggestion
                                 // The instance should now be from the failover partition.
				Node:                nodeTwo,
```

dependency/dependency_test.go
```diff
@@ -239,6 +229,34 @@ func (c *ClientSet) createConsulTestResources() error {
...
+	}
+
+	sgs, _, err := c.consul.client.ConfigEntries().List(QuerySamenessGroup, &api.QueryOptions{})
+	if err != nil {
+		return err
+	}
+	log.Printf("sameness groups: %+v", sgs)
+
+	return nil
+}
```
lkysow, line 258: For a realistic scenario, should we create the sameness group in both partitions with the members reversed?

dependency/dependency_test.go
```diff
@@ -239,6 +229,34 @@ func (c *ClientSet) createConsulTestResources() error {
...
 }
 
+func (c *ClientSet) createConsulSamenessGroups(name, partition, failoverPartition string) error {
+	members := append([]api.SamenessGroupMember{}, api.SamenessGroupMember{
+		Partition: partition,
+	}, api.SamenessGroupMember{
+		Partition: failoverPartition,
+	})
+	sg := &api.SamenessGroupConfigEntry{
+		Kind:               QuerySamenessGroup,
```
lkysow, line 239: nit: does the API package expose the config entry kind constants? If not I'd just use the literal string here rather than tie this to our query which just happens to use the same name right now

dependency/health_service_test.go
```diff
@@ -20,45 +22,45 @@ func TestNewHealthServiceQuery(t *testing.T) {
...
 				nil,
-				true,
+				fmt.Errorf(`health.service: invalid format: "?ns=%s"`, tenancy.Namespace),
 			},
 			testCase{
 				tenancyHelper.AppendTenancyInfo("invalid query param (unsupported key)", tenancy),
 				"name?unsupported=test",
 				nil,
-				true,
+				fmt.Errorf(`health.service: invalid query parameter key "unsupported" in query "unsupported=test": supported keys: ns,peer,partition`),
```
lkysow, line 63: shouldn't the list of supported keys include `sameness-group`? Maybe we forgot to update something somewhere?

dependency/health_service_test.go
```diff
@@ -675,6 +691,136 @@ func TestHealthServiceQuery_Fetch(t *testing.T) {
 	}
 }
 
+// TestHealthServiceQuery_Fetch_SamenessGroup tests fetching services with sameness group
+// and asserts that failover works when a service in the sameness group fails and consul-template
+// reacts to the change.
+func TestHealthServiceQuery_Fetch_SamenessGroup(t *testing.T) {
```
lkysow, line 697: There's some behavior I don't think we're currently testing:
* Querying for a service in a sameness group in a partition that the SG isn't defined in returns nothing (or an error?).
* If a service becomes healthy again in partitionOne, it returns partitionOne again
* If a second service instance comes up, it is also returned.
* If a service instance is deregistered it is no longer retured.
* If a sameness group is deleted it returns an error (what happens if it gets recreated?)
* If a sameness group ordering or list of members changes the returned service instance is updated
* This works across peers (i.e. multiple DCs)

Maybe some of those tests are too difficult to write or don't make sense because they're double-testing things? Interested in your thoughts here.

lkysow (commented):
Looks great. Only thing "blocking" from me is the question about other things to test.

dependency/dependency_test.go
jmurret: it was debugging for the long line below.  I am removing these remnants.

dependency/health_service.go
jmurret: good call. thank you.

dependency/dependency_test.go
jmurret: good call.

dependency/health_service_test.go
jmurret: I'll cross reference the blocking query tests in consul, but I added most of this there so that this was just a test to make sure that consul-template hooks into what the api is doing.

dependency/dependency_test.go
jmurret: i think this was probably incidental from an IDE refactor or something.  will fix.

dependency/health_service_test.go
jmurret: that is a good observation. let me check consul.

dependency/dependency_test.go
jmurret: removed in edd0f7394427f64c6fc9de7aa74dd6bac9c32f80

dependency/dependency_test.go
jmurret: removed in edd0f7394427f64c6fc9de7aa74dd6bac9c32f80

dependency/health_service.go
jmurret: elevated the guard clause in edd0f7394427f64c6fc9de7aa74dd6bac9c32f80

dependency/health_service_test.go
jmurret: fixed in edd0f7394427f64c6fc9de7aa74dd6bac9c32f80

dependency/dependency_test.go
jmurret: fixed in edd0f7394427f64c6fc9de7aa74dd6bac9c32f80

dependency/health_service_test.go
lkysow: Ahh understood. Then for this test can you add a comment to the effect of:

This test ensures that consul-template re-runs when the blocking query updates. The different behaviors of the blocking query, e.g. "when a service becomes unhealthy it should failover, when it becomes healthy again it should fail back", are tested in the Consul codebase.

dependency/health_service.go
```diff
@@ -122,18 +124,25 @@ func healthServiceQuery(s string, connect bool) (*HealthServiceQuery, error) {
...
-		name:      m["name"],
-		near:      m["near"],
-		tag:       m["tag"],
-		connect:   connect,
-		namespace: queryParams.Get(QueryNamespace),
-		peer:      queryParams.Get(QueryPeer),
-		partition: queryParams.Get(QueryPartition),
-	}, nil
+	if queryParams.Get(QuerySamenessGroup) != "" && queryParams.Get(QueryPeer) != "" {
+		return nil, fmt.Errorf("health.service: cannot specify both peer and sameness-group")
```
lkysow, line 128:
```suggestion
		return nil, fmt.Errorf("health.service: cannot specify both %s and %s", QueryPeer, QuerySamenessGroup)
```

dependency/health_service.go
```diff
@@ -261,6 +271,9 @@ func (d *HealthServiceQuery) String() string {
 	if d.peer != "" {
 		name = name + "@peer=" + d.peer
 	}
+	if d.samenessGroup != "" {
+		name = name + "@samenessGroup=" + d.samenessGroup
```
lkysow, line 275: nit: do these usually use camel case or dash-separated?

dependency/dependency.go
```diff
@@ -175,7 +181,8 @@ func GetConsulQueryOpts(queryMap map[string]string, endpointLabel string) (url.V
 			switch key {
 			case QueryNamespace,
 				QueryPeer,
-				QueryPartition:
+				QueryPartition,
+				QuerySamenessGroup:
 			default:
 				return nil,
 					fmt.Errorf("%s: invalid query parameter key %q in query %q: supported keys: %s,%s,%s", endpointLabel, key, queryRaw, QueryNamespace, QueryPeer, QueryPartition)
```
lkysow, line 188: @jmurret I think this is where you should add `QuerySamenessGroup`

dependency/health_service_test.go
jmurret: this validation was in consul-template.  fixed this in 2896a992ebd4710ccb9e67c6962ca4f1207b7181

dependency/health_service_test.go
jmurret: Will do.  I will cross-reference your suggestions above and in the other comment and open a PR for the missing scenarios in consul.

dependency/health_service.go
jmurret: fixed in a2260c48e2712a81c5e0846d0527428ed257557c

dependency/dependency_test.go
jmurret: resolved in edd0f7394427f64c6fc9de7aa74dd6bac9c32f80

dependency/health_service_test.go
jmurret: updated test docstring in 7c90b52ed98516de357533004645677e66074442.  will resolve this when a PR for missing behavior assertions is opened in Consul.

lkysow (approved):

docs/templating-language.md
jmurret: I've updated this in 7002f6235037a0c136fbff7e4594349fbbbad4c6.

dependency/health_service_test.go
jmurret: @lkysow thank you for calling these out.  In adding coverage I found a bug when the sameness groups does not exist or is deleted.  New tests and bug fix are in this PR: https://github.com/hashicorp/consul-enterprise/pull/9445

Here is a summary of what you asked about and confirmation each is in the `TestHealth_ServiceNodes_BlockingQuery_SamenessGroupParam` in consul enterprise after additional tests (and some already existed)
- If a service becomes healthy again in partitionOne, it returns partitionOne again
- If a service instance is deregistered it is no longer returned.
- If a sameness group ordering or list of members changes the returned service instance is updated
- If a second service instance comes up, it is also returned.
- If a sameness group is deleted it returns an error (what happens if it gets recreated?)
- Querying for a service in a sameness group in a partition that the SG isn't defined in returns an error.
- Querying for a Sameness group that does not exist returns an error (should be same code flow as above but worth having a test).

nathancoleman (commented):
LGTM

PR #1905, opened 2024-04-03
feat: add pkiSign function to use vault pki sign api endpoint

n-marton:
This pull requests intends to add a function to leverage the `sign` api endpoint in vault pki.

This can be really useful if one plans to use a large number of certificates with low TTL, where the usage of the `issue` endpoint can cause high load on the vault servers. By using the sign endpoint one can shift the heaviest part of the problem (the private key generation), from server side to client side.


eduardolmedeiros:
hey guys, any chance to have a look on that?
appreciate.

alexandrenas:
Hello team, could please someone have a look? It will be nice to have it soon. Thanks in advance!

dependency/vault_pki.go
```diff
@@ -136,6 +140,11 @@ func (d *VaultPKIQuery) Fetch(clients *ClientSet, opts *QueryOptions) (interface
 	default:
 		return PemEncoded{}, nil, err
 	}
+	// In the case that we are using sign vault endpoint we wont have an private key in the response
+	// so we should pass the one the we generated
```
aslamovamir, line 144:
```suggestion
	// when using the sign Vault endpoint, the response will not include a private key
	// therefore, we should pass the one we generated
```

dependency/vault_pki_test.go
```diff
@@ -25,13 +25,18 @@ func init() {
...
-	d1, _ := NewVaultPKIQuery("pki/issue/example-dot-com", "/unique_1", nil)
+	d1, _ := NewVaultPKIQuery("pki/issue/example-dot-com", "/unique_1", nil, nil)
 	id1 := d1.String()
-	d2, _ := NewVaultPKIQuery("pki/issue/example-dot-com", "/unique_2", nil)
+	d2, _ := NewVaultPKIQuery("pki/issue/example-dot-com", "/unique_2", nil, nil)
 	id2 := d2.String()
 	if id1 == id2 {
 		t.Errorf("IDs should be unique.\n%s\n%s", id1, id2)
 	}
+	d3, _ := NewVaultPKIQuery("pki/sign/example-dot-com", "/unique_1", nil, nil)
```
aslamovamir, line 35:
```suggestion
	d3, _ := NewVaultPKIQuery("pki/sign/example-dot-com", "/unique_3", nil, nil)
```
Shouldn't this be 3?

dependency/vault_pki_test.go
```diff
@@ -25,13 +25,18 @@ func init() {
...
+	id3 := d3.String()
+	if id1 == id3 {
```
aslamovamir, line 37:
```suggestion
	if id1 == id3 {
```
```suggestion
	if id1 == id3 || id1 == id2 {
```
Shouldn't we make sure all the 3 ids are unique against all of them?

dependency/vault_pki_test.go
```diff
@@ -25,13 +25,18 @@ func init() {
...
+		t.Errorf("IDs should be unique.\n%s\n%s", id1, id3)
```
aslamovamir, line 38:
```suggestion
		t.Errorf("IDs should be unique.\n%s\n%s\n%s", id1, id3, id2)
```
I think we need to compare all the 3?

docs/templating-language.md
```diff
@@ -765,6 +766,57 @@ to separate files from a template.
...
+means, the private key generation is happening on the consul template side, this
+can be quite useful if one generates a high number of certificates with low ttl
+which can put high load on the vault servers.
+
+The templating behaviour is the same as we have in `pkiCert` with a few special 
+attributes. You also need to pass `key_type=rsa|ec|ed25519` in alignment with your
+role on vault server. If you have `use_csr_common_name` and/or `use_csr_sans` as true
+on in your role, you should also pass them here, so the CSR will be appended with those
+values (they can have any value `use_csr_sans=value` or `use_csr_common_name=value` the 
+code only check for the key).
```
aslamovamir, line 782:
```suggestion
Query [Vault][vault] for a PKI certificate. This is quite similar to `pkiCert`;
however, instead of using the `issue` API endpoint, it uses the `sign`. This also 
means that the private key generation happens on the Consul template side. This
can be particularly useful when generating a high number of certificates with a low 
TTL, which can put a high load on the Vault servers.

The templating behavior is the same as in `pkiCert`, with a few special attributes.
You also need to pass `key_type=rsa|ec|ed25519` in alignment with your
role on the Vault server. If you have `use_csr_common_name` and/or `use_csr_sans`
to`true` in your role, you should also pass them here so the CSR is appended with those
values. They can have any value (`use_csr_sans=value` or `use_csr_common_name=value`),
as the code only check for the key. 
```

template/funcs.go
```diff
@@ -456,7 +464,193 @@ func pkiCertFunc(b *Brain, used, missing *dep.Set, destPath string) func(...stri
...
+			if k == "use_csr_sans" {
+				useCSRSans = true
+			}
+		}
+
+		var csrTemplate x509.CertificateRequest
+
+		// if we passed use_csr_common_name, that means serverside will expect the commonname from the csr
+		// so besides adding that param to the csr template, we also remove it from the map we pass later
+		// to vault, this way we spare a warning from the server side
```
aslamovamir, line 542:
```suggestion
		// if we passed use_csr_common_name, serverside will expect the commonname from the csr
		// so besides adding that param to the csr template, we also remove it from the map and pass it later
		// to vault; this way, we spare a warning from the server side
```

n-marton:
> Hi @n-marton , thanks a lot for opening the PR and your contribution. Your changes look great! I left some feedback. Also, we need unit tests added in here to ensure your new functions are covered in unit tests (to test they work in different use cases and edge cases and in case someone changes it in the future, unit tests catch errors): https://github.com/hashicorp/consul-template/blob/main/template/funcs_test.go

hey @aslamovamir 

thanks for getting back on this MR, I haven't wrote a test there because the cert issue func also did not have one, in fact that part is kind lacking heavily all together, however I am happy to write one for my new func, IF that is the only one thing we miss to merge this feature

aslamovamir:
> > Hi @n-marton , thanks a lot for opening the PR and your contribution. Your changes look great! I left some feedback. Also, we need unit tests added in here to ensure your new functions are covered in unit tests (to test they work in different use cases and edge cases and in case someone changes it in the future, unit tests catch errors): https://github.com/hashicorp/consul-template/blob/main/template/funcs_test.go
> 
> hey @aslamovamir
> 
> thanks for getting back on this MR, I haven't wrote a test there because the cert issue func also did not have one, in fact that part is kind lacking heavily all together, however I am happy to write one for my new func, IF that is the only one thing we miss to merge this feature

Hi @n-marton, yup I see - I honestly am not part of the codeowners for this repo or never contributed to it so did not know other functions also lack unit tests here. But would be great to have it written for yours if thats okay. All the rest looks good to me in your PR and can approve it once we have unit test coverage.

template/funcs.go
```diff
@@ -456,7 +464,192 @@ func pkiCertFunc(b *Brain, used, missing *dep.Set, destPath string) func(...stri
...
+		}
+
+		// generating private keys and also pem encode them for later usage
+		if keyType == "rsa" {
+			key, err := rsa.GenerateKey(rand.Reader, keyBits)
+			if err != nil {
+				return nil, err
+			}
+			privateKey = key
+			csrTemplate.SignatureAlgorithm = x509.SHA512WithRSA
```
victorr, line 574: What about looking for a `signature_bits` parameter to set the signature algorithm? Please consider it, as well as `use_pss` parameter for the RSA case.

https://github.com/hashicorp/vault/blob/371ffc4bd47a70c09e6b5c59cdd659ec09b719fa/sdk/helper/certutil/helpers.go#L1297-L1309

docs/templating-language.md
```diff
@@ -765,6 +766,57 @@ to separate files from a template.
...
+### pkiSign
+
+Query [Vault][vault] for a PKI certificate. This is quite similar to `pkiCert`;
+however, instead of using the `issue` API endpoint, it uses the `sign`. This also 
+means that the private key generation happens on the Consul template side. This
+can be particularly useful when generating a high number of certificates with a low 
+TTL, which can put a high load on the Vault servers.
+
+The templating behavior is the same as in `pkiCert`, with a few special attributes.
+You also need to pass `key_type=rsa|ec|ed25519` in alignment with your
```
victorr, line 778: Please mention `key_bits` here.

victorr (commented):
This is a wonderful PR, thank you so much for your contribution, it should prove very useful.

I have added some comments for your consideration. The most important one is about supporting `alt_names`. Please also consider addressing the other comments, or at least mentioning in the documentation the the defaults you have used for the signature algorithms.

I am not one of the code owners for this project, but we should have someone from the right team to review the PR soon.

PR #1921, opened 2024-05-03
Fixes Namespace Concat Bug for KV2 secrets

divyaac:
Pushing this first:

Vault-ent tests that test this new functionality: https://github.com/hashicorp/vault-enterprise/pull/5811 (Note, these will not pass in CI, until this pr is pushed a new consul-template version is used)

After pushing this PR:
1. Push new consul-template version (there is a release process that is time consuming)
2. Modify the consul-template version in the test pr above : https://github.com/hashicorp/vault-enterprise/pull/5811
3. Tests should pass, version is updated, will push the linked test pr.


dependency/vault_read.go
```diff
@@ -208,6 +208,16 @@ func shimKVv2Path(rawPath, mountPath string) string {
 		if strings.HasPrefix(p, "data/") || strings.HasPrefix(p, "metadata/") {
 			return rawPath
 		}
+
+		// If the raw path contains "/data/", it means the namespace on the Vault client
+		// is different from the namespace prefixing the rawPath. We want to
+		// keep the rawPath as is, and have the Vault client can pass the  namespace as it's header
```
VioletHynes, line 214:
```suggestion
		// keep the rawPath as is, and have the Vault client can pass the namespace as its header
```

VioletHynes: "its" is the possessive form of it, "it's" means "it is" :)

dependency/vault_read.go
```diff
@@ -208,6 +208,16 @@ func shimKVv2Path(rawPath, mountPath string) string {
...
+		// so the concatenation is handled by the Vault server.
+		if strings.Contains(p, "/data/") {
```
VioletHynes, line 216: Hmm, I wonder if this might be too naive -- in particular, what if I have a secret that's at `secret/foo/bar/data/baz`?

It's unlikely, but it's possible.

dependency/vault_read.go
```diff
@@ -208,6 +208,16 @@ func shimKVv2Path(rawPath, mountPath string) string {
 		if strings.HasPrefix(p, "data/") || strings.HasPrefix(p, "metadata/") {
 			return rawPath
 		}
+
+		// If the raw path contains "/data/", it means the namespace on the Vault client
+		// is different from the namespace prefixing the rawPath. We want to
```
VioletHynes, line 213: Is this true? Would this be true for KVv1 secrets, too?

VioletHynes (commented):
Some comments so far

dependency/vault_read.go
VioletHynes: Hmm, another concern: if I'm accessing KVv2 secrets without `/data/` today, then this function will automatically add it for me -- we should try to preserve that behaviour if possible, so that it works for all namespaces

dependency/vault_read.go
divyaac: Do you mean if secret/foo/bar is a path that's some combination of a namespace and the mount path? And baz is the secret name?

dependency/vault_read.go
divyaac: > Hmm, another concern: if I'm accessing KVv2 secrets without /data/ today, then this function will automatically add it for me -- we should try to preserve that behaviour if possible, so that it works for all namespaces

I believe this should still work - below there's an extra if statement that adds `"data/" `to the path if it doesn't exist, and that will still be preserved.

dependency/vault_read.go
VioletHynes: I mean for example, `secret` is the mount path, and `foo/bar/data/baz` is where the secret lived. This could or could not be prefixed by a namespace

dependency/vault_read.go
VioletHynes: Aha! Thank you for assuaging my concerns there!

I think as part of testing this (and before we merge this code ideally) we should triple check to make sure this works for secrets that rely on the bit of the code that adds `"data/"` when it's missing _and_ we have this namespace setup where it's in the client as well as the path. I'm a little concerned about that situation since we return here before we get to the bit we add `"data/"`.

dependency/vault_read.go
divyaac: That makes sense. I modified this.

dependency/vault_read.go
divyaac: That makes sense, and I modified this to work correctly. I also have some vault-enterprise tests here that test this: https://github.com/hashicorp/vault-enterprise/pull/5811

dependency/vault_read.go
```diff
@@ -196,18 +196,30 @@ func deletedKVv2(s *api.Secret) bool {
...
+		if !strings.HasSuffix(clientNamespace, "/") {
+			clientNamespace += "/"
+		}
+		// Extract client namespace from mount path if it exists
+		rawPathNsAndMountPath := strings.TrimPrefix(mountPath, clientNamespace)
+
+		// Trim (mount path - client namespace) from the raw path
+		p := strings.TrimPrefix(rawPath, rawPathNsAndMountPath)
+
+		log.Printf("[ERR] divya modified raw path: %s", p)
```
VioletHynes, line 215: I don't think this was meant to be in the commit :)

dependency/vault_read.go
```diff
@@ -196,18 +196,30 @@ func deletedKVv2(s *api.Secret) bool {
...
 		return path.Join(mountPath, "data")
 	default:
-		p := strings.TrimPrefix(rawPath, mountPath)
+
+		// Canonicalize the client namespace path to always having a '/' suffix
+		if !strings.HasSuffix(clientNamespace, "/") {
+			clientNamespace += "/"
+		}
+		// Extract client namespace from mount path if it exists
+		rawPathNsAndMountPath := strings.TrimPrefix(mountPath, clientNamespace)
```
VioletHynes, line 210: Hmm, what about if I have a namespace structure like this:
```
violet@violet-TX54KFJWXM vault-enterprise % vault namespace create a
Key                Value
---                -----
custom_metadata    map[]
id                 MMl1a
path               a/
violet@violet-TX54KFJWXM vault-enterprise % VAULT_NAMESPACE=a vault namespace create a
Key                Value
---                -----
custom_metadata    map[]
id                 BDwyU
path               a/a/
```

In this situation, I might have `a/` on my path but `a/` on my client with the intent that the request goes to `a/a/`. I think that this code might trim one of the `a/`s and make the request go to namespace `a/`?

dependency/vault_read.go
```diff
@@ -196,18 +196,28 @@ func deletedKVv2(s *api.Secret) bool {
...
 	case rawPath == mountPath, rawPath == strings.TrimSuffix(mountPath, "/"):
 		return path.Join(mountPath, "data")
 	default:
-		p := strings.TrimPrefix(rawPath, mountPath)
 
+		// Canonicalize the client namespace path to always having a '/' suffix
+		if !strings.HasSuffix(clientNamespace, "/") {
+			clientNamespace += "/"
+		}
+		// Extract client namespace from mount path if it exists
```
VioletHynes, line 209: Under what circumstances might the mountPath contain the clientNamespace 'unintentionally'? If I'm trying to render a secret in `ns1`, what might cause the namespace to be in both?

dependency/vault_read_test.go
```diff
@@ -652,66 +652,104 @@ func TestVaultReadQuery_String(t *testing.T) {
...
 			"secret/",
 			"secret/data",
+			"",
 		}, {
 			"prefix added to mount path not exact match",
 			"secret",
 			"secret/",
 			"secret/data",
+			"",
+		},
```
VioletHynes, line 721: We should add a test case for a secret residing at `a/a/secret`, with `a/` on the path and `a/` on the client

dependency/vault_read.go
divyaac: if the full mountPath is `a/b/c/secret/`

mountPath will be `a/b/c/secret/' when the client namespace = [ `nil`, `a`, `a/b`]
mountPath will be `secret/` when the client. namespace = [`a/b/c`

VioletHynes: Right, I don't think I worded my question correctly:

What I'm wondering is:
In the situation where I'm trying to access a secret at `a/b/c/secret` and my mount path requires this `TrimPrefix` call, e.g.
if my client namespace is `a/b/c/` and my mountPath is `c/secret` or similar. How does that happen? What circumstances does the call to `TrimPrefix` solve for?

dependency/vault_read.go
divyaac: I believe what will happen here is : 
clientNamespace = `a/`
rawPath = `a/secret/data`
mountPath = `a/a/secret/`

`a/` will be trimmed from `mountPath` to get `a/secret/`
`a/secret/` will be trimmed from rawPath to get `data/` 
Then rawPath will be returned giving `a/secret/data` as we want (because client namespace is `a/`)

VioletHynes: Ohh, I see now, got it! The mount path will always contain the namespace, so we should always trim it. Thanks for the explanation!

dependency/vault_read.go
VioletHynes: I think I answered my own question: this will happen any time we have a namespace, right?

VioletHynes (approved):
Awesome work! This is a really well-thought out fix for a tricky bug. Good job!

dependency/vault_read.go
```diff
@@ -196,18 +196,30 @@ func deletedKVv2(s *api.Secret) bool {
...
-func shimKVv2Path(rawPath, mountPath string) string {
+func shimKVv2Path(rawPath, mountPath, clientNamespace string) string {
 	switch {
 	case rawPath == mountPath, rawPath == strings.TrimSuffix(mountPath, "/"):
 		return path.Join(mountPath, "data")
 	default:
-		p := strings.TrimPrefix(rawPath, mountPath)
+
+		// Canonicalize the client namespace path to always having a '/' suffix
+		if !strings.HasSuffix(clientNamespace, "/") {
```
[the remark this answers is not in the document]
VioletHynes: Yeah, though I like keeping it in `consul-template` too for anything else that happens to be using it. Like I said, out of scope for this PR, but it seems like we have a need for external repos to be able to use the Canonicalize logic without repeating it. Food for thought!

PR #1923, opened 2024-05-06
[NET-8062] Consul-CT compatibility matrix

kkavish:
What is in this PR?
- We have added the compatibility matrix in the README.md which gives everyone a view about how different versions of CT work with different versions of Consul.

Why this PR?
- https://hashicorp.atlassian.net/browse/NET-8062. Support team requested this from us.

How did we test?
- We built docker images with Consul and CT binaries inside them, started Consul and then ran all the CT tests to see whether they pass.
- If they failed we analysed them and checked whether the failure was due to a hardcoded value in the test, removed them and re-ran.
- The code is checked in here https://github.com/hashicorp/consul-template/pull/1907


david-yu (approved):

jmurret (changes requested):
Please update the PR description to also include a meaning description of the change so that when reviewing the git log, the commit has a meaning description and not just a unique identifier to a ticketing system.

PR #1940, opened 2024-05-21
[NET-9467, NET-9468] Add partitions + exportedServices funcs

nathancoleman:
### Overview
This pull request adds two new functions to consul-template:
- `partitions`, which lists all partitions in the local datacenter using the existing [`/v1/partitions` API](https://developer.hashicorp.com/consul/api-docs/admin-partitions#list-all-partitions)
- `exportedServices`, which lists all exported services in a given partition using the existing [`/v1/exported-services` API](https://developer.hashicorp.com/consul/api-docs/exported-services#list-exported-services)

### Example
By combining these two functions, you can list the services imported into the local partition from all other partitions within the local datacenter.

> [!NOTE]
> The following should probably be optimized with a `{{ break }}` to end the `range` early once `$localPartition` has been identified as a consumer of a given service; however, I haven't had a chance to test with that change in place yet, so I'm noting it here.

```go
{{- $localPartition := "ap2" }}

// Loop through all partitions in the local datacenter
{{- range $partition := partitions }}

  // Loop through exported services for partition, unless it's our own partition
  {{- if not (eq $partition.Name $localPartition) }}
    {{- range $svc := (exportedServices $partition.Name) }}

      // For each service, loop through partition consumers looking for our own partition
      {{- range $consumer := .Consumers.Partitions }}
    	
        // Print only when the service is exported to our own partition
        {{- if eq $consumer $localPartition }}
        {{- $svc.Service }} -> ExportedFrom: {{ $partition.Name }} To {{ $consumer }}
        {{- end }}

      {{- end }}
    {{- end }}
  {{- end }}
{{- end }}
```


docs/templating-language.md
nathancoleman, line 1: Note: We need to add the `exportedServices` func here as well

docs/templating-language.md
```diff
@@ -506,6 +507,15 @@ To query a different data center and order by shortest trip time to ourselves:
...
 [Go's text/template][text-template] map indexing.
 
+### `partitions`
+
+Query [Consul][consul] for all partitions.
+
+```golang
+{{ range partitions }}
+{{ . }}{{ end }}
+```
```
blake, line 517:
```suggestion
{{ range partitions }}
{{ .Name }}
{{ end }}
```
```

`.` is a map. Shouldn't this be `.Name` so that the partition name is output?

File: dependency/nomad_var_get_test.go [hunk 2]
```diff
@@ -121,7 +121,6 @@ func TestNewNVGetQuery(t *testing.T) {
 			assert.Equal(t, tc.exp, act)
 		})
 	}
-	fmt.Println("done")
```
jm96441n, line 124: this was unnecessary

dependency/dependency_test.go
```diff
@@ -111,24 +125,7 @@ func TestMain(m *testing.M) {
...
-				testVault.Stop()
-				testNomad.Stop()
-				panic(r)
-			}
-		}()
-
-		exitCh <- m.Run()
-	}()
-
-	exit := <-exitCh
```
jm96441n, line 131: this channel shenanigans doesn't actually do anything

dependency/dependency_test.go
```diff
@@ -43,6 +43,20 @@ func TestMain(m *testing.M) {
 	tb := &test.TestingTB{}
 	runTestConsul(tb)
 	clients := NewClientSet()
+
+	defer func() {
```
jm96441n, line 47: moved the defer here so we actually catch panics during setup, this won't catch panics during test runs, I spent some time trying to get to a working solution but it would require checking for running instances of consul/nomad/vault that did not previously get cleaned up before running

dependency/consul_exported_services.go
```diff
@@ -0,0 +1,162 @@
...
+			Partitions:     partitions,
+			SamenessGroups: samenessGroups,
+		},
+	}
+}
+
+// NewListExportedServicesQuery parses a string of the format @dc.
+func NewListExportedServicesQuery(s string) (*ListExportedServicesQuery, error) {
+	return &ListExportedServicesQuery{
+		stopCh:    make(chan struct{}, 1),
```
rboyer, line 80: Stop channels do not need to have capacity so long as you only block on `<-stopCh` and signal with `close(stopCh)`.

dependency/consul_exported_services.go
```diff
@@ -0,0 +1,162 @@
...
+		RawQuery: opts.String(),
+	})
+
+	// This is certainly not elegant, but the partitions endpoint does not support
+	// blocking queries, so we are going to "fake it until we make it". When we
+	// first query, the LastIndex will be "0", meaning we should immediately
+	// return data, but future calls will include a LastIndex. If we have a
+	// LastIndex in the query metadata, sleep for 15 seconds before asking Consul
+	// again.
+	//
```
rboyer, line 101: The config entries endpoints can do blocking queries properly.

dependency/consul_exported_services.go
```diff
@@ -0,0 +1,162 @@
...
+	rm := &ResponseMetadata{
+		LastContact: qm.LastContact,
+		LastIndex:   qm.LastIndex,
+	}
+
+	return exportedServices, rm, nil
+}
+
+// CanShare returns if this dependency is shareable.
+// TODO What is this?
```
jmurret, line 147: remnant?
```suggestion
```

dependency/consul_exported_services_test.go
```diff
@@ -0,0 +1,153 @@
...
+	capi "github.com/hashicorp/consul/api"
+)
+
+func TestListExportedServicesQuery_Fetch(t *testing.T) {
+	testCases := map[string]struct {
+		partition        string
+		exportedServices *capi.ExportedServicesConfigEntry
+		expected         []ExportedService
+	}{
+		//"no services": {},
```
jmurret, line 17: remnant?

dependency/consul_partitions.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/pkg/errors"
+)
+
+var (
+	// Ensure implements
+	_ Dependency = (*ListPartitionsQuery)(nil)
+
+	// ListPartitionsQuerySleepTime is the amount of time to sleep between
+	// queries, since the endpoint does not support blocking queries.
+	ListPartitionsQuerySleepTime = 15 * time.Second
```
jmurret, line 21: [NIT] wonder if we want to create just one `LongPollingSleepTime` set to 15 seconds for use across all of the non blocking query endpoints?

dependency/consul_partitions.go
```diff
@@ -0,0 +1,112 @@
...
+		log.Printf("[TRACE] %s: long polling for %s", c, ListPartitionsQuerySleepTime)
+
+		select {
+		case <-c.stopCh:
+			return nil, nil, ErrStopped
+		case <-time.After(ListPartitionsQuerySleepTime):
+		}
+	}
+
+	// TODO Consider using a proper context
```
jmurret, line 69: do you want to commit this TODO?

dependency/consul_partitions.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+	}
+
+	// Use respWithMetadata which always increments LastIndex and results
+	// in fetching new data for endpoints that don't support blocking queries
+	return respWithMetadata(resp)
+}
+
+// CanShare returns if this dependency is shareable.
+// TODO What is this?
```
jmurret, line 97: remnant?
```suggestion
```

dependency/consul_exported_services.go
jm96441n: ah yeah, this comment is incorrect as we're using blocking queries here

dependency/consul_partitions_test.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func TestNewListPartitionsQuery(t *testing.T) {
+	cases := []struct {
+		name string
+		exp  *ListPartitionsQuery
+		err  bool
+	}{
+		{
+			"empty",
```
jmurret, line 22: [NIT] were you planning on adding any use cases to this? i don't think you really have to but if this test is not going to describe all/more of the expected ways that NewListPartitionsQuery() is going to set properties, I think we could just remove this test since this empty case is covered in other tests.  Not critical, but just calling that since this was here, I was expecting it to tell me more.  If this test did not exist, I would not ask for it.

jmurret (approved):
Approving, nothing blocking from my additional comments, but think you do need to address RB's comment about exported services implementing blocking queries.  You'll want to remove that logic that does this select based on the sleep time.

dependency/consul_exported_services.go
jmurret: you also want to remove the "fake it til you make it" code below it as well.

dependency/consul_partitions_test.go
jm96441n: yeah I'm not particularly sure this test has much value beyond testing that the `New` function returns something usable, I'm okay removing this

dependency/consul_partitions.go
jm96441n: so we kinda can, we do some re-assigning of these within tests and I'm not 100% on how that will play out if these tests run in parallel (since we could get a concurrent read and write) what I did though was move this to a `DefaultNonBlockingQuerySleepTime` constant and assign the individual sleep times to that const

dependency/consul_exported_services.go
jm96441n: yeah all that is gone now, I copied/pasted from another endpoint and missed removing this

PR #1942, opened 2024-05-27
Hack week: integrate with HCP Vault Secrets aka HVS

ncabatoff:
Currently oriented around HVS in a PRDE, since that's convenient for testing.  As such, hacked the tests to put test instances of Nomad and Vault on 127.0.0.2 instead of 127.0.0.1, as they would otherwise collide with PRDE proxy.  To run the tests it's therefore necessary to run `sudo ifconfig lo0 alias 127.0.0.2` on MacOS or `sudo route add -host 127.0.0.2 dev lo` on Linux.

The following env vars also need to be set:
* HCP_API_TLS=disabled
* HCP_API_ADDRESS=localhost:28081
* PRDE_SERVICE_NAME=ncabatoff01-9yir6h5q (use the part between "registry" and ".dev.pedp" in the "docker registry" output line of hcloud hcp dev proxy)
* HCP_CLIENT_SECRET
* HCP_CLIENT_ID
* HCP_PROJECT_ID
* HCP_ORGANIZATION_ID

You can test the new funcs using  `make dev && consul-template -once -dry -template "in.tpl:out.txt"`, where in.tpl looks like:
```
{{ range hvsSecrets "sample-app" }}
{{ hvsSecret (printf "sample-app/%s" .) }}
{{ end }}
```

The output (assuming you created the sample-app using the gui) should be:
```
> out.txt

Reveal your secret value as a way to familiarize yourself with Vault Secrets. You're on your way to becoming an expert!

Add a new secret.

Read your new secret via the CLI or API using the instructions below.

Go to the Integrations page and create a sync integration. Then change one of your secret values to see it updated in your sync integration.
```


watch/view.go
```diff
@@ -305,7 +305,7 @@ func (v *View) fetch(doneCh, successCh chan<- struct{}, errCh chan<- error) {
 	}
 }
 
-const minDelayBetweenUpdates = time.Millisecond * 100
+const minDelayBetweenUpdates = time.Millisecond * 1000
```
ncabatoff, line 308: This is still not really sufficient.  With this change, each HVS secret appearing in a template now only gets polled once per second.  But it wouldn't take very many templated secrets to start hitting API rate limits.

PR #1949, opened 2024-06-05
[NET-9468, NET-9902] Differentiate the ListExportedServicesQuery dependency based on the provided partition

nathancoleman:
Prior to this change, the template below would result in printing the services exported by the `default` partition twice.

```golang
{{ range exportedServices `default` }}
  ...
{{ end }}

{{ range exportedServices `ap1` }}
  ...
{{ end }}
```

This is because the `String` func -- used to uniquely identify a dependency provider -- returned the same value each time and wasn't differentiated based on its arguments. Consul-template would then consider these the same dependency and return the same set of services each time.

The fix is to include the differentiator -- the partition name in this case -- in the unique identity returned by `.String()` for the dependency. With this change, consul-template correctly identifies the code above as having two different dependencies and thus will create a separate dependency for each partition and make the appropriate calls to Consul.

> [!NOTE]
> For use with Consul CE, the provided partition name should be an empty string. This change will still work in that case as `<Dependency>.String()` will return `list.exportedServices()`.

### DIFF
--- a/template/funcs_test.go
+++ b/template/funcs_test.go
@@ -397,3 +397,21 @@ func Test_hmacSHA256Hex(t *testing.T) {
 		})
 	}
 }
+
+// Test_exportedServicesFunc verifies that calling exportedServices w/ a
+// different partition tracks an additional dependency
+func Test_exportedServicesFunc(t *testing.T) {
+	var used, missing dep.Set
+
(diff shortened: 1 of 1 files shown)


DanStough (approved):
👍

PR #1841, opened 2023-11-20
Revert "Revert "Implement systemd notification support after all templates ha…"

skpratt:
Reverts hashicorp/consul-template#1837

PR #1954, opened 2024-06-18
feat: Add support for systemd notify

dyudin0821:
- Integrated systemd notification support using github.com/coreos/go-systemd/v22/daemon.
- Updated manager/runner.go to signal readiness to systemd.
- Modified .golangci.yml to include github.com/coreos/go-systemd in linters-settings.
- Updated dependencies in go.mod and go.sum for github.com/coreos/go-systemd/v22 v22.5.0.

**Purpose**:

- This enhancement enables the application to efficiently notify systemd about its readiness, improving overall service management capabilities and alignment with modern system practices.

See the description of PR #1794 and #1841 respectively for detail info.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -198,14 +198,12 @@ func (cli *CLI) Run(args []string) int {
 			case *config.KillSignal:
 				fmt.Fprintf(cli.errStream, "Cleaning up...\n")
 				runner.StopImmediately()
-				cancel() // Cancel the context to stop the readyCh goroutine
 				return ExitCodeInterrupt
 			default:
 				// Propagate the signal to the child process
 				runner.Signal(s)
 			}
 		case <-cli.stopCh:
-			cancel() // Cancel the context to stop the readyCh goroutine
 			return ExitCodeOK
 		}
 	}
--- a/cli.go
+++ b/cli.go
@@ -13,6 +13,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/coreos/go-systemd/v22/daemon"
 	"github.com/hashicorp/consul-template/config"
 	"github.com/hashicorp/consul-template/logging"
 	"github.com/hashicorp/consul-template/manager"
@@ -105,11 +106,29 @@ func (cli *CLI) Run(args []string) int {
 		return ExitCodeOK
 	}
 
+	// Create a channel for signaling readiness to the systemd init system.
+	readyCh := make(chan struct{})
+
+	// Start a goroutine that listens for signals on the readyCh channel.
+	// When a signal (an empty struct) is received, it notifies systemd that
+	// the application is ready by calling daemon.SdNotify with SdNotifyReady.
+	// If an error occurs during the notification, it logs a warning message.
+	go func() {
+		for range readyCh {
+			_, err := daemon.SdNotify(false, daemon.SdNotifyReady)
+			if err != nil {
+				log.Printf("[WARN] failed to signal readiness to systemd: %v", err)
+			}
+		}
+	}()
+
 	// Initial runner
 	runner, err := manager.NewRunner(config, dry)
 	if err != nil {
 		return logError(err, ExitCodeRunnerError)
 	}
+
+	runner.SetReadyChannel(readyCh)
 	go runner.Start()
 
 	// Listen for monitored signals
@@ -163,6 +182,7 @@ func (cli *CLI) Run(args []string) int {
 				if err != nil {
 					return logError(err, ExitCodeRunnerError)
 				}
+				runner.SetReadyChannel(readyCh)
 				go runner.Start()
 			case *config.KillSignal:
 				fmt.Fprintf(cli.errStream, "Cleaning up...\n")
--- a/cli.go
+++ b/cli.go
@@ -4,6 +4,7 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"io"
@@ -106,6 +107,10 @@ func (cli *CLI) Run(args []string) int {
 		return ExitCodeOK
 	}
 
+	// Create a context with cancellation
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
 	// Create a channel for signaling readiness to the systemd init system.
 	readyCh := make(chan struct{})
 
@@ -114,10 +119,15 @@ func (cli *CLI) Run(args []string) int {
 	// the application is ready by calling daemon.SdNotify with SdNotifyReady.
 	// If an error occurs during the notification, it logs a warning message.
 	go func() {
-		for range readyCh {
-			_, err := daemon.SdNotify(false, daemon.SdNotifyReady)
-			if err != nil {
-				log.Printf("[WARN] failed to signal readiness to systemd: %v", err)
+		for {
+			select {
+			case <-readyCh:
+				_, err := daemon.SdNotify(false, daemon.SdNotifyReady)
+				if err != nil {
+					log.Printf("[WARN] failed to signal readiness to systemd: %v", err)
+				}
+			case <-ctx.Done():
+				return
 			}
 		}
 	}()
@@ -156,6 +166,7 @@ func (cli *CLI) Run(args []string) int {
 		case <-service_os.Shutdown_Channel():
 			fmt.Fprintf(cli.errStream, "Cleaning up...\n")
 			runner.StopImmediately()
+			cancel() // Cancel the context to stop the readyCh goroutine
 			return ExitCodeInterrupt
 		case s := <-cli.signalCh:
 			log.Printf("[DEBUG] (cli) receiving signal %q", s)
@@ -187,12 +198,14 @@ func (cli *CLI) Run(args []string) int {
 			case *config.KillSignal:
 				fmt.Fprintf(cli.errStream, "Cleaning up...\n")
 				runner.StopImmediately()
+				cancel() // Cancel the context to stop the readyCh goroutine
 				return ExitCodeInterrupt
 			default:
 				// Propagate the signal to the child process
 				runner.Signal(s)
 			}
 		case <-cli.stopCh:
+			cancel() // Cancel the context to stop the readyCh goroutine
 			return ExitCodeOK
 		}
 	}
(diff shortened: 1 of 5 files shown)


dyudin0821:
Hello @hashicorp/team-consul-india

Could you please take a look?

Local tests for me were successful
```
make test-race
==> Testing consul-template (race)
go: downloading github.com/hashicorp/consul/sdk v0.15.0
go: downloading github.com/hashicorp/go-gatedio v0.5.0
go: downloading github.com/hashicorp/go-hclog v1.6.3
go: downloading golang.org/x/sys v0.20.0
go: downloading github.com/hashicorp/hcl v1.0.0
go: downloading github.com/hashicorp/vault/api v1.10.0
go: downloading github.com/mitchellh/go-homedir v1.1.0
go: downloading github.com/mitchellh/mapstructure v1.5.0
go: downloading github.com/pkg/errors v0.9.1
go: downloading golang.org/x/exp v0.0.0-20240506185415-9bf2ced13842
go: downloading github.com/stretchr/testify v1.8.4
go: downloading github.com/hashicorp/consul/api v1.28.3
go: downloading github.com/hashicorp/go-rootcerts v1.0.2
go: downloading github.com/hashicorp/nomad/api v0.0.0-20230103221135-ce00d683f9be
go: downloading github.com/hashicorp/vault/api/auth/kubernetes v0.5.0
go: downloading github.com/hashicorp/go-syslog v1.0.0
go: downloading github.com/hashicorp/logutils v1.0.0
go: downloading github.com/coreos/go-systemd/v22 v22.5.0
go: downloading github.com/hashicorp/go-multierror v1.1.1
go: downloading github.com/mitchellh/hashstructure v1.1.0
go: downloading github.com/BurntSushi/toml v1.3.2
go: downloading github.com/Masterminds/sprig/v3 v3.2.3
go: downloading github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
go: downloading github.com/hashicorp/go-sockaddr v1.0.6
go: downloading github.com/imdario/mergo v0.3.13
go: downloading golang.org/x/text v0.14.0
go: downloading gopkg.in/yaml.v2 v2.4.0
go: downloading github.com/hashicorp/go-cleanhttp v0.5.2
go: downloading github.com/hashicorp/go-uuid v1.0.3
go: downloading github.com/hashicorp/go-version v1.6.0
go: downloading github.com/fatih/color v1.17.0
go: downloading github.com/mattn/go-isatty v0.0.20
go: downloading github.com/cenkalti/backoff/v3 v3.2.2
go: downloading github.com/go-jose/go-jose/v3 v3.0.3
go: downloading github.com/hashicorp/errwrap v1.1.0
go: downloading github.com/hashicorp/go-retryablehttp v0.7.2
go: downloading github.com/hashicorp/go-secure-stdlib/parseutil v0.1.7
go: downloading github.com/hashicorp/go-secure-stdlib/strutil v0.1.2
go: downloading golang.org/x/net v0.24.0
go: downloading golang.org/x/time v0.3.0
go: downloading github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2
go: downloading gopkg.in/yaml.v3 v3.0.1
go: downloading github.com/hashicorp/serf v0.10.1
go: downloading github.com/gorilla/websocket v1.5.0
go: downloading github.com/hashicorp/cronexpr v1.1.1
go: downloading github.com/Masterminds/goutils v1.1.1
go: downloading github.com/Masterminds/semver/v3 v3.2.0
go: downloading github.com/google/uuid v1.3.0
go: downloading github.com/huandu/xstrings v1.4.0
go: downloading github.com/mitchellh/copystructure v1.2.0
go: downloading github.com/shopspring/decimal v1.3.1
go: downloading github.com/spf13/cast v1.5.0
go: downloading golang.org/x/crypto v0.22.0
go: downloading github.com/mattn/go-colorable v0.1.13
go: downloading github.com/ryanuber/go-glob v1.0.0
go: downloading github.com/armon/go-metrics v0.4.1
go: downloading github.com/mitchellh/reflectwalk v1.0.2
go: downloading github.com/hashicorp/go-immutable-radix v1.3.1
go: downloading github.com/hashicorp/golang-lru v1.0.2
ok      github.com/hashicorp/consul-template    3.563s
ok      github.com/hashicorp/consul-template/child      2.128s
ok      github.com/hashicorp/consul-template/config     1.214s
ok      github.com/hashicorp/consul-template/dependency 15.014s
ok      github.com/hashicorp/consul-template/logging    1.504s
ok      github.com/hashicorp/consul-template/manager    7.681s
ok      github.com/hashicorp/consul-template/renderer   1.197s
?       github.com/hashicorp/consul-template/service_os [no test files]
ok      github.com/hashicorp/consul-template/signals    1.030s
ok      github.com/hashicorp/consul-template/template   1.246s
ok      github.com/hashicorp/consul-template/test       1.037s
?       github.com/hashicorp/consul-template/version    [no test files]
ok      github.com/hashicorp/consul-template/watch      4.575s
```

jmurret:
thanks @dyudin0821 Iam not familiar with go-systemd, but do you know how this will interact with non-linux operating systems that we have [releases](https://releases.hashicorp.com/consul-template/0.38.1/) for?

manager/runner.go
```diff
@@ -309,6 +310,13 @@ func (r *Runner) Start() {
 
 		if r.allTemplatesRendered() {
 			log.Printf("[DEBUG] (runner) all templates rendered")
+
+			// Send a signal to systemd that we are ready.
+			_, err := daemon.SdNotify(false, daemon.SdNotifyReady)
```
tgross, line 315: This is going to be problematic for downstream applications that consume consul-template as a library, such as Nomad. This would cause Nomad to fire a readiness signal every time an allocation has its templates rendered, which we don't want here.

We should send the notification signal from a goroutine on `main` and then send signals from here to that goroutine when it's available.

tgross (changes requested):
In addition to my note about moving the signal into the top-level `main`, we might also want to look at whether we want to bring in this entire dependency. The systemd notify protocol is rather simple. In Nomad we implemented the protocol in a single file: [`sdnotify_linux.go`](https://github.com/hashicorp/nomad/blob/main/command/agent/sdnotify_linux.go), which allowed us to have a non-Linux code path that was a no-op in [`sdnotify_default.go`](https://github.com/hashicorp/nomad/blob/main/command/agent/sdnotify_default.go)

dyudin0821:
> thanks @dyudin0821 Iam not familiar with go-systemd, but do you know how this will interact with non-linux operating systems that we have [releases](https://releases.hashicorp.com/consul-template/0.38.1/) for?
> 
> Ideally, I think in adding functionality we'd probably want to add similar behavior that is consistent across operating systems if possible.

@jmurret Thank you so much for taking the time to review my change!

 Sending notify using go-systemd is implemented in such a way that if the OS does not support systemd, the NOTIFY_SOCKET environment variable will simply not be set, and the function will return (false, nil), indicating that the function will safely handle this situation.

https://github.com/coreos/go-systemd/blob/main/daemon/sdnotify.go#L48
```
	// NOTIFY_SOCKET not set
	if socketAddr.Name == "" {
		return false, nil
	}
```

manager/runner.go
dyudin0821: Hello @tgross 
I completely agree with you. I have changed the approach to implementation according to your recommendations.
Thx

cli.go
```diff
@@ -163,6 +182,7 @@ func (cli *CLI) Run(args []string) int {
 				if err != nil {
 					return logError(err, ExitCodeRunnerError)
 				}
+				runner.SetReadyChannel(readyCh)
```
tgross, line 185: Given that we're shutting down the runners in these signal handlers, shouldn't we have a context on the ready channel that we can close to stop that goroutine as well?

tgross (changes requested):
Thanks for improving upon the original approach here @dyudin0821. I've left one more suggestion to tighten up shutdown. There was also a failing test but it looked like a panic in an unrelated area of the code? Once you've made the change I've suggested, I'd rebase on `main` just to make sure you've got the latest bug fixes from the CT team.

cli.go
dyudin0821: Hello @tgross 
Thank you very much for your contribution and time spent!
If I understood your request correctly, then the idea is to create a context with cancellation, pass the context to the goroutine listening to the channel, and сanceled the context when shutting down the runners.


I apologize if this implementation isn't perfect, as this is my first experience making such changes. Please review the changes and let me know if further adjustments are needed.

dyudin0821: As for the fallen CI pipeline with tests, I noticed that there is a similar problem for neighboring PRs. I will be glad to help you fix the dropped tests

cli.go
```diff
@@ -163,16 +193,19 @@ func (cli *CLI) Run(args []string) int {
...
 				go runner.Start()
 			case *config.KillSignal:
 				fmt.Fprintf(cli.errStream, "Cleaning up...\n")
 				runner.StopImmediately()
+				cancel() // Cancel the context to stop the readyCh goroutine
```
tgross, line 201: This is safe, but we also don't need it, because you have `defer cancel()` above and we're returning here.

```suggestion
```

cli.go
```diff
@@ -163,16 +193,19 @@ func (cli *CLI) Run(args []string) int {
...
 				return ExitCodeInterrupt
 			default:
 				// Propagate the signal to the child process
 				runner.Signal(s)
 			}
 		case <-cli.stopCh:
+			cancel() // Cancel the context to stop the readyCh goroutine
```
tgross, line 208: Same here

```suggestion
```

cli.go
tgross: > I apologize if this implementation isn't perfect, as this is my first experience making such changes. Please review the changes and let me know if further adjustments are needed.

Looks very close. I've added a suggestion for two one-line fixes and this will be good-to-merge.

> As for the fallen CI pipeline with tests, I noticed that there is a similar problem for neighboring PRs. I will be glad to help you fix the dropped tests

Oof, ok... let's not worry about it in this PR then.

tgross (approved):
LGTM! Thanks @dyudin0821!

I'll merge once CI is done and that'll go out whenever CT gets released next.

Issue #1961, opened 2024-07-04
consul-template pkiCert does not return CA_Chain data causing issues during intermediary rotation


astundzia:
### Hashicorp issue
This is for Hashicorp ticket #153442. This fixes the issues we are seeing as described in the hashicorp ticket.

### Consul Template version
latest `main`, based off of `v0.39.0`

### Configuration

```hcl
pid_file = "/var/run/vault-agent-pki.pid"

auto_auth {
  method "approle" {
    config = {
      role_id_file_path = "/etc/vault/role_id"
      secret_id_file_path = "/etc/vault/secret_id"
      remove_secret_id_file_after_reading = false
    }
  }

  sink "file" {
    config = {
      path = "/etc/vault-token"
    }
  }
}

cache {
  use_auto_auth_token = true
}

listener "tcp" {
  address = "127.0.0.1:8100"
  tls_disable = true
}

vault {
  address = "https://192.168.0.106:8200"
}

template {
  destination = "/etc/certs/certs.pem"
  wait {
    min = "5s"
    max = "30s"
  }
  contents = <<EOH
          {{- with pkiCert "pki/astundzia/issuing_ca/issue/genctl-consoleproxy-tls" "common_name=console-proxy.vpc.cloud.ibm.local" "format=pem" -}}
          {{ .Cert }}
          {{ .Key  }}
          {{ .Key | writeToFile "/etc/certs/client.key" "" "" "0644" }}
          {{ .Cert | writeToFile "/etc/certs/client.crt" "" "" "0644" }}
          {{- end }}
          {{- with secret "pki/astundzia/issuing_ca/cert/ca_chain" -}}
          {{- .Data.ca_chain }}
          {{- .Data.ca_chain | writeToFile "/etc/certs/client.crt" "" "" "0644" "append,newline" }}
          {{ end }}
EOH
}

```

### Command

Running `vault` as `vault-agent`

### Expected behavior & bug
When using vault-agent which uses consul-template, I expect the entire CAChain (the data from the `ca_chain` API) to be returned when I use `with pkiCert`.

The data returned from `with pkiCert` does not contain the full CA Chain, only the issuer (via `.CA`). This means we need to use   ` {{- with secret "pki/astundzia/issuing_ca/cert/ca_chain" -}}` to fetch the remaining bits of the ca_chain.

[...]

PR #1962, opened 2024-07-04
Enhance with pkiCert template to return full CA chain

astundzia:
This is for Github issue https://github.com/hashicorp/consul-template/issues/1961

This provides the `CAChain` data on the `with pkiCert` function. This is necessary to support rotation of a intermediary CA without causing service interruptions.

Instead of using this template:
```
          {{- with pkiCert "pki/astundzia/issuing_ca/issue/genctl-consoleproxy-tls" "common_name=console-proxy.vpc.cloud.ibm.local" "format=pem" -}}
          {{ .Cert }}
          {{ .Key  }}
          {{ .Key | writeToFile "/etc/certs/client.key" "" "" "0644" }}
          {{ .Cert | writeToFile "/etc/certs/client.crt" "" "" "0644" }}
          {{- end }}
          {{- with secret "pki/astundzia/issuing_ca/cert/ca_chain" -}}
          {{- .Data.ca_chain }}
          {{- .Data.ca_chain | writeToFile "/etc/certs/client.crt" "" "" "0644" "append,newline" }}
          {{ end }}
```

I can instead just use:
```
          {{- with pkiCert "pki/astundzia/issuing_ca/issue/genctl-consoleproxy-tls" "common_name=console-proxy.vpc.cloud.ibm.local" "format=pem" -}}
          {{ .Cert }}
          {{ .Key  }}
          {{ .CAChain }}
          {{ .Key | writeToFile "/etc/certs/client.key" "" "" "0644" }}
          {{ .Cert | writeToFile "/etc/certs/client.crt" "" "" "0644" }}
          {{- range .CAChain }}
          {{- . | writeToFile "/etc/certs/client.crt" "" "" "0644" "append" }}
          {{- end }}
          {{- end }}
```

While similar, this has two critical benefits:
1. During intermediary CA rotation, the leaf certificate never expires. Vault agent is not aware nor should it be. The `with secret` however now pulls a new `ca_chain` which causes a client certificate to be created that is `{Leaf cert v1, Issuing CA V2}`. Since The leaf cert was issued by a different issuer (the previous one) the chain that vault-agent creates is invalid (Authority/signatures do not match).
2. Since `ca_chain` is returned on the `with pkiCert`, we no longer make a high number of API calls to vault `ca_chain` endpoint.

Without this code, we could not find a way to trust a single rootCA & support rotation of a intermediary CA/issuing CA.

This passes existing tests, and I've tested it manually using a self built version of the `vault 1.8.0-beta` code.

This also retains backwards compatibility (CA object on `with pkiCert` does not change functionally.

Please review!

### DIFF
--- a/dependency/vault_pki.go
+++ b/dependency/vault_pki.go
@@ -21,11 +21,40 @@ import (
 var _ Dependency = (*VaultPKIQuery)(nil)
 
 // Return type containing PEMs as strings
-type PemEncoded struct{ Cert, Key, CA string }
+type PemEncoded struct {
+	Cert, Key, CA string
+	CAChain       []string
+}
+
+func (a PemEncoded) Equals(b PemEncoded) bool {
+	if a.CA != b.CA || a.Cert != b.Cert || a.Key != b.Key {
+		return false
+	}
+
+	if len(a.CAChain) != len(b.CAChain) {
+		return false
+	}
+
+	for i, v := range a.CAChain {
+		if v != b.CAChain[i] {
+			return false
+		}
+	}
+	return true
+}
+
+func (a PemEncoded) CaChainContains(item string) bool {
+	for _, v := range a.CAChain {
+		if v == item {
+			return true
+		}
+	}
+	return false
+}
 
 // a wrapper to mimic v2 secrets Data wrapper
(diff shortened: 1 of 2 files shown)


sgmiller (approved):
Seems reasonable.

PR #1985, opened 2024-09-20
fix(lint): govet and staticcheck issues

thevilledev:
Fix the following quality-of-life issues:

- `govet`:
  - `printf: non-constant format string`
- `staticcheck`:
  - `SA9009: ineffectual compiler directive due to extraneous space: "// go:build ignore`


thevilledev:
Apparently this was fixed in https://github.com/hashicorp/consul-template/pull/2017

PR #2014, opened 2025-01-03
Always parse template wait intervals to avoid unnecessary rendering.

vaLski:
Fixing [[GH-1899](https://github.com/hashicorp/consul-template/pull/1899)] by always parsing template wait intervals to avoid unnecessary rendering.


vaLski:
Up.

Issue #2015, opened 2025-01-13
Vault KV v2: cannot access `subkeys/` path via `consul-template`


ryansammonaiven:
Currently, the code that shims `data/` into Vault KV v2 paths only checks if the path starts with `data/` or `metadata/`. ([code](https://github.com/hashicorp/consul-template/blob/6cbdad30ecf122701936108979b7e3be892461c2/dependency/vault_read.go#L215-L219))

This effectively means that only these paths are accessible within Vault KV v2 secrets engines.

For example, attempting to access a path starting with `subkeys/` ([Vault API Docs)](https://developer.hashicorp.com/vault/api-docs/secret/kv/kv-v2#read-secret-subkeys) will result in the shim transforming the path to `data/subkeys/...`, missing the desired endpoint.

The code that shims `data/` into Vault KV v2 paths needs to check for every possible valid prefix before deciding to shim in `data/`.

PR #2016, opened 2025-01-14
support Vault KV subkeys to avoid consul-template adding /data/ to the path

sofixa:
Add support for the [Vault KV subkeys](https://developer.hashicorp.com/vault/api-docs/secret/kv/kv-v2#read-secret-subkeys) API path.

Cf. https://github.com/hashicorp/consul-template/issues/2015

### DIFF
--- a/dependency/vault_read.go
+++ b/dependency/vault_read.go
@@ -212,9 +212,9 @@ func shimKVv2Path(rawPath, mountPath, clientNamespace string) string {
 		// Trim (mount path - client namespace) from the raw path
 		p := strings.TrimPrefix(rawPath, rawPathNsAndMountPath)
 
-		// Only add /data/ prefix to the path if neither /data/ or /metadata/ are
+		// Only add /data/ prefix to the path if neither /data/, or /metadata/ or /subkeys/ are
 		// present.
-		if strings.HasPrefix(p, "data/") || strings.HasPrefix(p, "metadata/") {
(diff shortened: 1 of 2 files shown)


sreeram77:
Hi @sofixa, couple of test cases are failing in TestShimKVv2Path. Could you please take a look at it?

sofixa:
Hi @sreeram77 , I can't test locally because there is no documentation on how to do that. It just vaguely says "install Nomad, Consul, Vault"... which I have, and am running locally. 

```
failed to start consul server: api unavailable
FAIL    github.com/hashicorp/consul-template/dependency 3.142s
```

Which isn't a very helpful.

How can I actually run this locally to be able to debug the failing test / my failing code?

sreeram77:
Hi @sofixa 
`make test` should ideally work. In your case, you can alternatively can try `go test ./... -run TestShimKVv2Path` which will run only the test cases for the modified function. If you're still getting the same error, there might be something wrong with the consul installation/binary.

sofixa:
What's the setup I need for Consul? Any test I try to run, it tries to start the Consul server via consul/sdk/testutil, but fails at https://github.com/hashicorp/consul/blob/4bca5c5d219a7d616f6ae19cca1bc5af913455c3/sdk/testutil/server.go#L467 . How could I even debug this? It doesn't seem related to my Consul setup, because it starts a test server via the SDK directly.

Adrian

sofixa:
Managed to work around the failing Consul dependency by creating a separate file calling the same function and `go run`-ing it directly. It was a dumb extra `/` in the test cases, fixed, all tests should be OK now.

sreeram77 (approved):
LGTM!

PR #2022, opened 2025-02-14
Update CHANGELOG and version files for release 0.40.0 

anandmukul93:
Updated the release version and changelog files for release details.

PS: some CI tests are failing temporarily for consul-enterprise due to license and vault connection issues and need to be fixed. The release works fine with consul and is prioritized over such issues.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ test:
 # test-race runs the test suite.
 test-race:
 	@echo "==> Testing ${NAME} (race)"
-	@go test -timeout=60s -race -tags="${GOTAGS}" ./... ${TESTARGS}
+	@go test -v -timeout=120s -race -tags="${GOTAGS}" ./... ${TESTARGS}
 .PHONY: test-race
 
 # _cleanup removes any previous binaries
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -54,7 +54,7 @@ test:
 # test-race runs the test suite.
 test-race:
 	@echo "==> Testing ${NAME} (race)"
-	@go test -timeout=60s -race -tags="${GOTAGS}" ./... ${TESTARGS}
+	@go test -v -timeout=120s -race -tags="${GOTAGS}" ./... ${TESTARGS}
```
srahul3, line 57: NIT: If you can explain how did you concluded to timeout to be increased from 60s to 120s?

srahul3 (approved):
✅ LGTM

Makefile
anandmukul93: the current testing time seems to be hitting just less than 1 min. We use this target from ci.yaml
https://github.com/hashicorp/consul-template/blob/f33df88a2f65ace9dd8660ceca84945999532235/.github/workflows/ci.yml#L39 
It is for future safeguarding since If the timeout is reached, the test process is killed with a panic.

PR #2042, opened 2025-03-24
fix: timeout issues for list.peerings

### LINKED ISSUE #2041: API peerings timeout for blocking queries
I keep running into these errors from Nomad after a few minutes:

```
Template failed: list.peerings: Get "http://127.0.0.1:8500/v1/peerings?index=41156796&stale=&wait=300000ms": context deadline exceeded
```

This is probably due to the `wait` query parameter being higher than the `DefaultContextTimeout` the internal Go-API uses. 

Related https://github.com/hashicorp/nomad/issues/25508

--- 
Previous issue: When using the following template, the template would not get re-rendered as more services were exported/imported. This might be fixed in the more recent versions in Consul Template. 

```go-template
{{ range peerings }}
{{ range .StreamStatus.ImportedServices }}
# {{ . }}
{{ end }}
{{ end }}

### PR DESCRIPTION
**Pull request by EtienneBruines**
The previously used `DefaultContextTimeout` does not work well with the user-definable `WaitTime`, which can easily exceed this global timeout.

Fixes #2041

### DIFF
--- a/dependency/consul_peering.go
+++ b/dependency/consul_peering.go
@@ -99,12 +99,7 @@ func (l *ListPeeringQuery) Fetch(clients *ClientSet, opts *QueryOptions) (interf
 		RawQuery: opts.String(),
 	})
 
-	// list peering is a blocking API, so making sure the ctx passed while calling it
-	// times out after the default wait time.
-	ctx, cancel := context.WithTimeout(context.Background(), DefaultContextTimeout)
-	defer cancel()
-
-	p, meta, err := clients.Consul().Peerings().List(ctx, opts.ToConsulOpts())
+	p, meta, err := clients.Consul().Peerings().List(context.Background(), opts.ToConsulOpts())
 	if err != nil {
 		return nil, nil, errors.Wrap(err, l.String())
 	}
--- a/dependency/dependency.go
+++ b/dependency/dependency.go
@@ -44,11 +44,6 @@ const (
 	TypeNomad
 )
 
-const (
-	// DefaultContextTimeout context wait timeout for blocking queries.
-	DefaultContextTimeout = 60 * time.Second
-)
-
 // Dependency is an interface for a dependency that Consul Template is capable
 // of watching.
 type Dependency interface {

### CODE REVIEW

File: dependency/consul_peering.go [hunk 1]
```diff
@@ -99,10 +99,15 @@ func (l *ListPeeringQuery) Fetch(clients *ClientSet, opts *QueryOptions) (interf
...
-	defer cancel()
+	// The first result needs to be "immediate", in which case we use the default timeout settings.
+	// For all other requests, we use a 'blocking' query, therefore setting the appropriate timeout.
+	// We add one second to allow for the Consul client to finish sending the response before
+	// declaring it an exceeded deadline.
+	var ctx context.Context
+	if opts.WaitTime != 0 {
+		ctx, cancel := context.WithTimeout(context.Background(), opts.WaitTime + time.Second)
+		defer cancel()
+	}
```
tgross, line 110: I'm not sure either the original or this patch work the way I'd expect to see with blocking queries, with the caveat that I work on Nomad and not Consul. But their [blocking queries](https://developer.hashicorp.com/consul/api-docs/features/blocking) behavior should be similar.

When we make a blocking query, we wait for an API response to come back. If there is no change before the `wait` time expires, we get a _non-error_ response back with the current value of the query, whether or not it has changed. If we look at how the [`Peerings.List`](https://github.com/hashicorp/consul/blob/api/v1.31.2/api/peering.go#L271-L295) API works, the context on this request does _not_ change the `wait` parameter! It sets the request timeout from the client. So that means when we hit this context deadline, the server is probably still waiting to send any update and will be sad because we're exiting early.

So if you're making a blocking query, the only reason you want to set a context other than `context.Background()` here (as we did for `consul_partitions.go`), is if you want the client side to bail out early. As we just saw, waiting until the deadline isn't an error in our use case here. But if you use a context, we'll get a context deadline error and return the error to the caller, instead of saying "ok, no change, nothing to do". Which is the behavior we expect of consul-template.

tgross (changes requested):

dependency/consul_peering.go
EtienneBruines: If I understand your comment and  the [TODO-comments to the Peerings APIs](https://github.com/hashicorp/consul/blob/api/v1.31.2/api/peering.go#L245) correctly, it seems like we do not need *any* context at all, other than perhaps `context.Background()`. Not in this case at least. I like that - it simplifies the code.

tgross (approved):
LGTM!

dependency/consul_peering.go
```diff
@@ -99,12 +99,7 @@ func (l *ListPeeringQuery) Fetch(clients *ClientSet, opts *QueryOptions) (interf
 		RawQuery: opts.String(),
 	})
 
-	// list peering is a blocking API, so making sure the ctx passed while calling it
-	// times out after the default wait time.
-	ctx, cancel := context.WithTimeout(context.Background(), DefaultContextTimeout)
```
sreeram77, line 104: const `DefaultContextTimeout` is unused and can be removed.

sreeram77 (changes requested):

sreeram77 (approved):
LGTM! Thank you for your contribution!

PR #2046, opened 2025-04-09
fix: pin golangci-lint version to v1.62.0 in workflow

sreeram77:
- pin golangci-lint version to v1.62.0 instead of using latest version in the lint workflow since lint action is failing with latest version.


Vikramarjuna (approved):

mrgupta-hashi (approved):
LGTM

Issue #2031, opened 2025-03-18
funcs: sprig function setup is computationally costly


jrasell:
I have done some recent work on the Nomad team to understand its template runner profile. Nomad utilizes consul-template as a library to provide the [job specification template block](https://developer.hashicorp.com/nomad/docs/job-specification/template) functionality.

The test I performed to generate the flame graph attached included taking a 120s profile of a Nomad client which was running allocations utilizing the `nomadVar` template function.
![Image](https://github.com/user-attachments/assets/956816ae-0b4e-4105-a4bf-dc09e777abc1)

Sprig function setup code reference: https://github.com/hashicorp/consul-template/blob/faf8404946d4ef55174e98c27a232e50761c6efd/template/template.go#L440-L443

The code is costly in two ways, particularly considering the `sprig.TxtFuncMap` function returns a map with 210 entries:
- the string concatenation used to generated the map key for the function name which is unavoidable no matter where the setup occurs
- the growth of the map which the sprig functions are entered into, which is initialised [here](https://github.com/hashicorp/consul-template/blob/faf8404946d4ef55174e98c27a232e50761c6efd/template/template.go#L330) and could potentially be improved

## Proposal
Move the template function setup functionality into a more efficient process to improve the overhead seen on all template execute calls.

It could also be possible to have a configuration option to disable to use of sprig functions, however, library implementors such as Nomad would need to plumb this through. This would be time consuming for both us as engineers but also Nomad users, as they would need to update all their configurations to make use of this.

To further demonstrate the overhead, I ran a quick experiment using the process described above with Nomad built using a hacked consul-template version which had removed the sprig setup code:
![Image](https://github.com/user-attachments/assets/62952490-48a4-4970-a92d-8f2f1b2af001)

thevilledev:
I have a proposal to improve this in #2052. The string concatenation cost is then incurred only once when the package loads, not on every template execution.

PR #2052, opened 2025-04-18
feat(template): optimize sprig function init

thevilledev:
Pre-compute Sprig template functions during package init.

Avoids repeated map allocation and string concatenation for Sprig functions on every template execution, addressing performance overhead reported in #2031.

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -34,6 +34,19 @@ var (
 	ErrMissingReaderFunction = errors.New("template: missing a reader function")
 )
 
+var (
+	sprigFuncMap template.FuncMap
+)
+
+func init() {
+	sprigFuncMap = make(template.FuncMap)
+	// Add the Sprig functions to the funcmap
+	for k, v := range sprig.TxtFuncMap() {
+		target := "sprig_" + k
+		sprigFuncMap[target] = v
+	}
+}
+
 // Template is the internal representation of an individual template to process.
 // The template retains the relationship between its contents and is
 // responsible for it's own execution.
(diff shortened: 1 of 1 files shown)


jrasell (commented):
Hi @thevilledev and thanks for raising this PR.

I am not a code owner of this repository, so don't have the final say at the moment. I did want to note that while this init function avoids the string concatenation overhead, this will not avoid the map regrowth cost also seen. It would be nice if we could look into this too, even if the investigation determines avoiding this is not possible.

I am also curious on your thoughts of how using a private global variable and init function might impact applications such as Nomad, which utilise consul-template to run and manage any number of templates on a single host via a single control application.

In the PR code at the moment, it would be possible to create the funcmap with a fixed/known size, as it's possible to know the size of the sprig map before creating this I believe. This would avoid unnecessary map growth work in this area.

template/template.go
```diff
@@ -34,6 +34,19 @@ var (
...
 )
 
+var (
+	sprigFuncMap template.FuncMap
+)
+
+func init() {
+	sprigFuncMap = make(template.FuncMap, len(sprig.TxtFuncMap()))
+	// Add the Sprig functions to the funcmap
+	for k, v := range sprig.TxtFuncMap() {
```
dduzgun-security, line 44: Using `HemerticTxtFuncMap` instead of `TxtFuncMap` would remove some sprig functions that may be problematic.

dduzgun-security (commented):
Thanks a lot for your contribution @thevilledev 🙌 

Added a small comment and will ask the team to review + add a deprecating note on sprig functions that `HemerticTxtFuncMap` removes before cutting a new release.

template/template.go
thevilledev: Done!

thevilledev:
Consul Enterprise related test is fixed in #2053 - but even with the fix it fails, possibly due to an outdated license in the repository secrets.

Vikramarjuna:
LGTM

PR #2064, opened 2025-06-04
fix: use 3.21 instead of latest for alpine in Dockerfile

sreeram77:
- Use 3.21 version instead of latest for alpine in the Dockerfile.


jainam51297:
Nice

PR #2078, opened 2025-08-06
add fileContents function for templating

markcampv:
This PR adds a new fileContents template function to Consul Template.

The fileContents function reads the contents of a specified file and returns it as a string. If the file does not exist or cannot be read, it returns an empty string. This makes it safe to use in conditional template logic, such as comparisons or fallbacks using functions like default, ne, or regexMatch.

This function is particularly useful in templates that need to make decisions based on the contents of existing files, such as determining whether to overwrite a file or trigger related logic when file contents differ.

### DIFF
--- a/docs/templating-language.md
+++ b/docs/templating-language.md
@@ -1877,6 +1877,20 @@ For example:
 {{ key "my/key/path" | writeToFile "/my/file/path.txt" "my-user" "my-group" "0644" "append,newline" }}
 ```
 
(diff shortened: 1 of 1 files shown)


tgross:
I have no opinion on this PR itself but should @hashicorp/consul-selfmanage-maintainers accept it, I'd ask if you all wouldn't mind letting us on Nomad know so we can add this to the denylist.

PR #2096, opened 2025-09-26
template: distinguish templates with identical contents but different destinations or sources

pkazmierczak:
Consul Template should distinguish templates that have the same body but
different sources or destinations. In cases where a slice of templates is
submitted to the runner with same contents and a dependency, templates with the
same contents body will effectively be "merged" into one. See
hashicorp/nomad#26800 for an example of a bug caused by this. 

## PCI review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## UNRELEASED
 
 IMPROVEMENTS:
-* Distinguish templates with identical contents but different destinations or sources
+* Distinguish templates with identical contents but different destinations or sources [GH-2096](https://github.com/hashicorp/consul-template/pull/2096)
 
 # 0.41.2 (September 19, 2025)
 


template/template.go
```diff
@@ -186,7 +186,7 @@ func NewTemplate(i *NewTemplateInput) (*Template, error) {
 	}
 
 	// Compute the MD5, encode as hex
-	hash := md5.Sum([]byte(t.contents))
+	hash := md5.Sum([]byte(t.source + t.contents + t.destination))
```
tgross, line 189: It probably would be a good idea to add null bytes between each of these just to make it robust against very silly inputs.

Should we swallow the pill and swap this out for a hash function that doesn't trip zealous security people while we're here? (ex. sha256, which also has hardware support on some architectures.) I'm fine with the answer being "not in this PR" 😁

tgross (dismissed):
LGTM! (Once we figure out why CI is failing)

dduzgun-security (approved):
LGTM, thanks for changing the `md5` to `sha256` part too.

kswap (approved):
LGTM

PR #2143, opened 2026-04-09
cve: fix sandbox path bypass in file helper through symlink attack

sreeram77:
The file template helper enforces sandbox_path during template evaluation, but the later dependency fetch reads the original raw path without re-checking the sandbox. This creates a time-of-check to time-of-use gap. If an attacker can retarget a symlink or equivalent link target after validation but before dependency fetch, consul-template can read an out-of-sandbox file and then render the cached contents even after the link is restored to an in-sandbox target.

## PCI review checklist


anilvpatel (approved):
LGTM

PR #2156, opened 2026-07-08
Prevent path redirection in writeToFile via symlinked path components

Surabhi-1605:
## Summary

The `writeToFile` template helper opened the final user-supplied path directly with `os.Create` / `os.OpenFile`, which follow symlinks in both intermediate directory components and the final filename. This allowed template output to be redirected outside the intended directory, or to overwrite an existing file via a pre-positioned symlink — a particular concern given `writeToFile` is documented for writing certificate and private-key material.

Unlike the main renderer's `AtomicWrite` (temp-file + rename, which replaces the directory entry rather than following it), `writeToFile` wrote directly to the named path and lost that protection.

## Changes

- Validate the write path before opening it:
  - Reject a symlinked **parent directory** component (`os.Lstat`), which would redirect the whole write into an attacker-controlled directory.
  - Reject a symlinked **final filename** component (`os.Lstat`), which would clobber an unintended target.
- Open the destination with `O_NOFOLLOW` (via a per-platform constant; no-op on Windows) so the open itself fails if the final component is a symlink, closing the TOCTOU window between the pre-check and the open.
- Apply ownership and permissions through the **open file descriptor** (`f.Chown` / `f.Chmod`) instead of the path, so metadata always targets the file that was opened rather than a path that could be swapped afterward.
- Fail closed: only create the parent directory when it genuinely does not exist (`os.IsNotExist`), and return immediately if the destination cannot be stat'd for any reason other than non-existence (e.g. permission denied), instead of proceeding to open.

`os.Lstat` is used rather than `filepath.EvalSymlinks` containment (as used by the sandboxed `file` helper) because `writeToFile` has no configured sandbox boundary, and an `EvalSymlinks`-based ancestor check produces false positives on OS-managed symlinks (e.g. macOS `/var -> /private/var`).

### Known limitation

Only the immediate parent directory is checked, not grandparent/ancestor components. This is deliberate: `writeToFile` has no sandbox to anchor a full ancestor walk, and walking every ancestor breaks compatibility on systems with OS-managed symlinks in common path prefixes. `O_NOFOLLOW` only guards the final path component; a residual TOCTOU on ancestor directories remains, consistent with this limitation.

## Tests

Adds `Test_writeToFile_symlink_rejection`:
- `rejects_final_component_symlink` — errors, sensitive file byte-verified unchanged
- `rejects_directory_component_symlink` — errors, redirect target byte-verified unchanged
- `allows_normal_path` — succeeds
- `append_rejects_final_component_symlink` — errors in append mode, sensitive file unchanged

Sub-tests skip gracefully where `os.Symlink` is unsupported and skip entirely as root. All pre-existing `Test_writeToFile` cases continue to pass.

### DIFF
--- a/template/funcs.go
+++ b/template/funcs.go
@@ -1930,7 +1930,11 @@ func writeToFile(path, username, groupName, permissions string, args ...string) 
 	}
 
 	resolvedPath := filepath.Join(dirPath, filepath.Base(cleanPath))
-	if fi, lErr := os.Lstat(resolvedPath); lErr == nil && fi.Mode()&os.ModeSymlink != 0 {
+	fi, lErr := os.Lstat(resolvedPath)
+	if lErr != nil && !os.IsNotExist(lErr) {
+		return "", fmt.Errorf("writeToFile: failed to stat %q: %w", resolvedPath, lErr)
+	}
+	if lErr == nil && fi.Mode()&os.ModeSymlink != 0 {
 		return "", fmt.Errorf("writeToFile: refusing to write to symlink %q", path)
 	}
 
--- a/template/funcs.go
+++ b/template/funcs.go
@@ -1897,25 +1897,51 @@ func writeToFile(path, username, groupName, permissions string, args ...string) 
 	}
 	perm := os.FileMode(p_u)
 
+	// Validate the write path to prevent symlink redirection attacks.
+	//
+	// os.Lstat does not follow the final path component, so it reveals whether
+	// that component is itself a symlink. We check both the immediate parent
+	// directory and the final file component:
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 IMPROVEMENTS:
 * update: go version to 1.26.2 [GH-2144](https://github.com/hashicorp/consul-template/pull/2144)
+* cve: fix path redirection and unintended file overwrite in `writeToFile` helper through symlink attack
```
panman90: you don't have to update change log of olderversion

panman90 (approved):
LGTM

PR #2163, opened 2026-07-29
runner: isolate template errors instead of aborting all templates [fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT]

### LINKED ISSUE #32035: Vault Agent aborts evaluation of all temlates and reschedules all if one fails
**Describe the bug**
If using Vault Agent to render multiple templates, if one fails then all of the other ones are aborted. Then, even the successfully rendered templates are also rescheduled to run again. This is problematic when combined with `exec {}` blocks within the template as those are also re-executed on all rescheduled templates which on our case causes the nomad agents to [constantly restart](https://github.com/hashicorp/nomad/issues/4593).

**To Reproduce**
1. Write two `template {}` blocks for Vault Agent
2. Make one of them contain an `exec {}` template that always fails
3. See that both templates keep re-rendering despite only one failing, alternatively that the functioning one never renders at all.

**Expected behavior**
Only the failing template should fail and be rescheduled, not all templates.

**Environment:**
* Vault Server Version (retrieve with `vault status`): v2.0.1
* Vault CLI Version (retrieve with `vault version`): v2.0.3
* Server Operating System/Architecture: Ubuntu 24.04

**Additional context**
N/A

waterWang:
## Description

When one template in a multi-template Vault Agent configuration encounters a fatal error (e.g., an `exec {}` template function fails), the consul-template runner's `Run()` method returns immediately after the first failure, skipping all remaining templates.

This means:
1. Template A with a failing `exec {}` block causes templates B, C, D to be skipped
2. The runner stops and sends the error to `ErrCh`
3. On restart, ALL templates are re-processed from scratch
4. The failing template fails again, creating an infinite loop

## Fix

Collect template errors and continue processing the remaining templates:
- When a template has a fatal error, log it, add to the error list, and continue to the next template
- If at least one template rendered successfully, return `nil` (don't stop the runner — the failed template will be retried on the next dependency change)
- If ALL templates failed, return an error so the caller can handle it

## Changes

`manager/runner.go` — 20 lines added, 1 deleted

## Related issue

Fixes hashicorp/vault#32035

### DIFF
--- a/manager/runner.go
+++ b/manager/runner.go
@@ -645,10 +645,14 @@ func (r *Runner) Run() error {
 		depsMap: make(map[string]dep.Dependency),
 	}
 
+	var templateErrs []error
+
 	for _, tmpl := range r.templates {
 		event, err := r.runTemplate(tmpl, runCtx)
 		if err != nil {
-			return err
+			log.Printf("[ERR] (runner) template %s failed: %s", tmpl.ID(), err)
+			templateErrs = append(templateErrs, err)
+			continue
 		}
 
 		// If there was a render event store it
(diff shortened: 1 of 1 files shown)


tgross (changes requested):
No one can review this till the CLA is signed . I'd ask reviewers to consider how this impacts other downstream projects like Nomad `template` blocks. Especially given the bug reported in Vault this is trying to fix was not acknowledged by the Vault team so maybe they intend for this behavior.

tgross:
Closing as AI slop spam

PR #2166, opened 2026-08-05
dependency: fix runaway credential creation on renewal failure

### LINKED ISSUE #2165: bug: RenewBehaviorErrorOnErrors causes runaway dynamic credential creation on renewal failure
### Consul Template version

v0.42.1 (embedded in Vault Agent v1.17.0; also reproduced on v1.16.1)

### Configuration

```hcl
# vault-agent.hcl
vault {
  address = "http://127.0.0.1:8200"
}

auto_auth {
  method "token_file" {
    config {
      token_file_path = "/home/vault/.vault-token"
    }
  }
}

template {
  contents    = "{{ with secret \"database/creds/my-role\" }}{{ .Data.username }}:{{ .Data.password }}{{ end }}"
  destination = "/tmp/database-config"
}
```

```liquid
{{ with secret "database/creds/my-role" }}
{{ .Data.username }}:{{ .Data.password }}
{{ end }}
```

```liquid
# database/roles/my-role (bad renew_statement that triggers the bug)
creation_statements  = ["CREATE ROLE ... WITH LOGIN PASSWORD '' VALID UNTIL ''; GRANT SELECT ON ALL TABLES IN SCHEMA public TO ...;"]
default_ttl          = "1m"
max_ttl              = "1h"
renew_statements     = ["PLEASE ALTER ROLE Samuel Lee;"]  # intentionally invalid SQL
```

### Command

```shell
vault agent -config=vault-agent.hcl -log-level=warn
```

### Debug output

The following repeats continuously at ~100ms intervals while the bad
`renew_statement` is in place:

```
[WARN] agent: vault.read(database/creds/my-role): failed to renew: Error making API request.
  URL: PUT http://vault.default.svc:8200/v1/sys/leases/renew
  Code: 400. Errors:
  * failed to renew entry: resp: (*logical.Response)(nil) err: 1 error occurred:
    * ERROR: syntax error at or near "PLEASE" (SQLSTATE 42601)
[WARN] agent: vault.read(database/creds/my-role): renewer done (maybe the lease expired)
[WARN] agent: vault.read(database/creds/my-role): failed to renew: ...
[WARN] agent: vault.read(database/creds/my-role): renewer done (maybe the lease expired)
... (repeats)
```

And on the database side:

```sql
teams=# SELECT count(*) FROM pg_roles WHERE rolname LIKE '%v-kubernet-my-role%';
 count
-------
   561   -- after ~1 minute with a 1m TTL; 300K+ observed in production
```

Compare with a role that has no `renew_statement` (or a valid one):

```sql
teams=# SELECT count(*) WHERE rolname LIKE '%v-kubernet-my-role%';
 count
-------
     1
```

### Expected behavior

When a lease renewal fails due to a bad `renew_statement` or intermittent
database connectivity, consul-template should apply exponential backoff and
keep retrying renewal until the original lease has genuinely expired, then
fetch exactly one new credential — i.e. one new database role per TTL.

### Actual behavior

On the first renewal error, the `LifetimeWatcher` exits immediately.
`renewSecret()` in `dependency/vault_common.go` treats the closed `DoneCh`
as a normal completion and returns `nil`. `VaultReadQuery.Fetch()` then
unconditionally calls `fetchSecret()`, minting a new `database/creds/my-role`
lease and a new database role. Because `fetchSecret` succeeds, the view loop's
retry counter resets to zero and the rate-limiter floor of 100ms is the only
throttle — producing roughly 600 new database roles per minute with a 1-minute
TTL. In production this resulted in 300,000+ unexpected database role
creations, causing a memory crunch on the database.

The root cause is a single line in `dependency/vault_common.go`. The
`LifetimeWatcher` is created with `RenewBehaviorErrorOnErrors`, which is the
legacy watcher mode: it exits immediately on the first renewal error with no
backoff and no regard for remaining lease time.

Changing it to `RenewBehaviorIgnoreErrors` fixes the issue. With
`IgnoreErrors`, the watcher applies exponential backoff (initial 10s, max 5m,
bounded by the remaining lease duration) and only signals `DoneCh` when the
original lease has genuinely expired — limiting re-fetches to at most once per
TTL regardless of how many transient errors occur.

### Steps to reproduce

1. Start a Vault dev server and configure a database secrets engine with a role
   that has an invalid `renew_statement`, e.g. `["PLEASE ALTER ROLE ...;"]`,
   and a short `default_ttl` (e.g. 1m).
2. Run Vault Agent with a template reading `database/creds/my-role`:
   `vault agent -config=vault-agent.hcl -log-level=warn`
3. Observe the agent logs — the `failed to renew` / `renewer done` pair repeats
   at ~100ms intervals immediately after the first renewal attempt.
4. Query the database: `SELECT count(*) FROM pg_roles WHERE rolname LIKE '%my-role%'`
   — the count grows by ~600 per minute instead of staying at 1.

### References

- reproduced on Vault Agent v1.16.1 and v1.17.0, both standalone (no
injector) and via the Vault Agent injector on Kubernetes.
- All retry-related Vault Agent annotations were tested; none affect this
  behaviour because the issue is inside the consul-template `LifetimeWatcher`
  configuration, not the agent retry layer

Siddharth-Kannan:
## PCI review checklist

Fixes #2165

## Problem

When a Vault lease renewal fails — due to a bad `renew_statement` on a database
role, or intermittent database connectivity — consul-template mints a new
dynamic credential on every view-loop iteration (~600 new database roles per
minute with a 1-minute TTL) instead of waiting out the original lease.

The root cause is in `dependency/vault_common.go`. The `LifetimeWatcher` was
created with `RenewBehaviorErrorOnErrors`:

```go
renewer, err := clients.Vault().NewLifetimeWatcher(&api.LifetimeWatcherInput{
    Secret:        vaultSecret,
    RenewBehavior: api.RenewBehaviorErrorOnErrors,
})
```

RenewBehaviorErrorOnErrors is the legacy watcher mode — it closes DoneCh
and exits immediately on the first renewal error. renewSecret() treats any
DoneCh close as a clean completion and returns nil. VaultReadQuery.Fetch()
then unconditionally calls fetchSecret(), minting a new credential.
Because fetchSecret succeeds, the view loop's retry counter resets to zero and
the 100ms rate-limiter floor is the only throttle.

Fix
```go
Change RenewBehaviorErrorOnErrors to RenewBehaviorIgnoreErrors:

renewer, err := clients.Vault().NewLifetimeWatcher(&api.LifetimeWatcherInput{
    Secret:        vaultSecret,
    RenewBehavior: api.RenewBehaviorIgnoreErrors,
})
```

RenewBehaviorIgnoreErrors is the current default mode of LifetimeWatcher.
On a renewal error it applies exponential backoff (initial 10s, max 5m) bounded
by the remaining lease duration, and only signals DoneCh when the original
lease has genuinely expired — so fetchSecret is called at most once per TTL
regardless of how many transient errors occur in between.

Testing
Tested manually with Vault Agent v1.16.1 and v1.17.0 using a database role with an intentionally
invalid renew_statement. Without this fix, pg_roles grows by ~600 rows per
minute. With this fix, the count stays at 1 for the full TTL.

- [x] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [x] If applicable, I've documented the impact of any changes to security controls.

### DIFF
--- a/dependency/vault_common_test.go
+++ b/dependency/vault_common_test.go
@@ -276,7 +276,7 @@ type renewalTestDep struct {
 	stopCh      chan struct{}
 }
 
-func (d *renewalTestDep) stopChan() chan struct{}          { return d.stopCh }
+func (d *renewalTestDep) stopChan() chan struct{}         { return d.stopCh }
 func (d *renewalTestDep) secrets() (*Secret, *api.Secret) { return d.secret, d.vaultSecret }
 func (d *renewalTestDep) CanShare() bool                  { return false }
 func (d *renewalTestDep) Fetch(*ClientSet, *QueryOptions) (interface{}, *ResponseMetadata, error) {
--- a/dependency/vault_common_test.go
+++ b/dependency/vault_common_test.go
@@ -5,8 +5,12 @@ package dependency
 
 import (
 	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


sanikachavan5 (changes requested):
RenewBehaviorIgnoreErrors helps for typical TTLs, but leases shorter than its 5–15s initial backoff may still trigger immediate re-fetching. Could we add a regression test for renewal failures, including a short TTL, to ensure credential creation remains bounded?

dependency/vault_common_test.go
```diff
@@ -186,3 +190,98 @@ func setupVaultPKI(clients *ClientSet) {
...
+	}
+
+	// Simulate the Fetch() loop: call renewSecret in a tight loop and count
+	// returns. The watcher should block via backoff — 0 returns expected in 5s.
+	var returns int32
+	go func() {
+		for {
+			select {
+			case <-stopCh:
+				return
```
sanikachavan5: The test currently counts the expected return caused by close(stopCh), so the assertion can nondeterministically see one return. Could we run renewSecret once, use a result channel to assert it does not return before the timeout, then close stopCh and wait for ErrStopped? This would also ensure the goroutine is cleaned up.

sanikachavan5 (changes requested):

dependency/vault_common_test.go
Siddharth-Kannan: That makes sense, I have added this change. It now checks for the duration of the initial backoff, and asserts that it does not return before the timeout.

sanikachavan5 (approved):
looks good to me!

PR #2167, opened 2026-08-06
template: preserve existing file when writeToFile receives empty content

charand2004:
## Problem
When Vault Agent restarts while a PKI certificate is still within its 
TTL, the agent skips re-rendering but still executes the writeToFile 
directive — overwriting the private key file with blank content.
Fixes: https://hashicorp.atlassian.net/browse/VAULT-38287
## Solution
Added a guard in writeToFile that skips the write if the incoming 
content is empty and the destination file already has content.
## Testing
- Added regression tests for empty content with existing file
- Added regression tests for empty content with new file
- All existing tests pass

Fixes: https://hashicorp.atlassian.net/browse/VAULT-38287

### DIFF
--- a/template/funcs.go
+++ b/template/funcs.go
@@ -1943,36 +1943,42 @@ func writeToFile(path, username, groupName, permissions string, args ...string) 
 	// above and the open below (no-op on Windows, which lacks O_NOFOLLOW).
 	var f *os.File
 	shouldAppend := strings.Contains(flags, "append")
-	// Guard: when content is empty and we are not in append mode, do not
-	// truncate an existing non-empty file.  This prevents Vault Agent from
-	// blanking a pkiCert side-file (e.g. the private key) when it restarts
-	// within a lease TTL and consul-template re-executes the template body
-	// without re-rendering the destination file (VAULT-38287).
-	if !shouldAppend && content == "" {
-		if existing, err := os.ReadFile(resolvedPath); err == nil && len(existing) > 0 {
-			return "", nil
-		}
-	}
-	if shouldAppend {
-		f, err = os.OpenFile(resolvedPath, os.O_APPEND|os.O_WRONLY|os.O_CREATE|openNoFollow, perm)
-		if err != nil {
-			return "", err
-		}
-	} else {
-		f, err = os.OpenFile(resolvedPath, os.O_RDWR|os.O_CREATE|os.O_TRUNC|openNoFollow, perm)
-		if err != nil {
-			return "", err
-		}
-	}
-	defer f.Close()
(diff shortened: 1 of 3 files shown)


rana (approved):
LGTM, thank you

Surabhi-1605:
LGTM

template/funcs.go
```diff
@@ -1943,6 +1943,16 @@ func writeToFile(path, username, groupName, permissions string, args ...string)
 	// above and the open below (no-op on Windows, which lacks O_NOFOLLOW).
 	var f *os.File
 	shouldAppend := strings.Contains(flags, "append")
+	// Guard: when content is empty and we are not in append mode, do not
+	// truncate an existing non-empty file.  This prevents Vault Agent from
+	// blanking a pkiCert side-file (e.g. the private key) when it restarts
+	// within a lease TTL and consul-template re-executes the template body
+	// without re-rendering the destination file (VAULT-38287).
+	if !shouldAppend && content == "" {
+		if existing, err := os.ReadFile(resolvedPath); err == nil && len(existing) > 0 {
```
sanikachavan5: are there no other cases where the content can be empty? this seems like a broad case.

sanikachavan5 (changes requested):

template/funcs.go
charand2004: Do you mean that the fix itself is narrow(there are other cases that return blank files that need to be fixed) or broad(There are cases where someone would WANT to pass empty content)?

sanikachavan5: I mean the latter: the fix is broad because there are legitimate cases where users may intentionally pass empty content to truncate or clear an existing file. With this change, writeToFile would silently preserve the old content instead, potentially leaving stale data.
Could we scope the behavior specifically to the Vault PKI reuse case, or make preservation of existing content explicitly opt-in?

sanikachavan5 (changes requested):

template/funcs.go
charand2004: Gotcha, I'll make it opt-in via a preserve_on_empty flag so default behavior is unchanged.  Does that sound good?

sanikachavan5: Yep, making it opt-in sounds good and keeps the default behavior unchanged.
One thing to watch: even when preserving the content, we should still apply the configured ownership and permissions. It’d also be good to cover the flag’s interaction with newline in the tests.

charand2004: I'll make sure ownership and permissions are still  applied on the early return, and I'll add test cases covering the  interaction between preserve_on_empty and newline. Thanks!

charand2004:
The watch package panic is unrelated — changes only touch template/funcs.go and template/template_test.go.

sanikachavan5 (approved):
looks good to me!

PR #2168, opened 2026-08-18
Introduce a tls.Config field on VaultConfig

ncabatoff:
Context: https://hashicorp.atlassian.net/browse/VAULT-46384 adds support for authenticating to Vault using a client cert whose private key is stored in a TPM, meaning it can't be referenced as a file.

To accommodate this, introduce a `TLSConfig *tls.Config` field on `VaultConfig` that allows callers to supply a pre-built TLS configuration instead of using the file-based TLS config fields. .

## PCI review checklist

### DIFF
--- a/config/vault.go
+++ b/config/vault.go
@@ -4,6 +4,7 @@
 package config
 
 import (
+	"crypto/tls"
 	"fmt"
 	"time"
 
@@ -67,6 +68,12 @@ type VaultConfig struct {
 
 	// SSL indicates we should use a secure connection while talking to Vault.
 	SSL *SSLConfig `mapstructure:"ssl"`
+
+	// TLSConfig allows direct configuration of TLS settings for the Vault client.
+	// This is intended for programmatic use only and cannot be set via configuration files.
+	// When present, this takes precedence over the SSL field configuration.
+	// It is an error to set both TLSConfig and SSL fields.
+	TLSConfig *tls.Config `mapstructure:"-" json:"-"`
 
 	// Token is the Vault token to communicate with for requests. It may be
 	// a wrapped token or a real token. This can also be set via the VAULT_TOKEN
@@ -166,6 +173,8 @@ func (c *VaultConfig) Copy() *VaultConfig {
 		o.SSL = c.SSL.Copy()
 	}
 
+	o.TLSConfig = c.TLSConfig
+
 	o.Token = c.Token
 
 	o.VaultAgentTokenFile = c.VaultAgentTokenFile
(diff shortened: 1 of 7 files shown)


config/vault.go
```diff
@@ -277,7 +291,12 @@ func (c *VaultConfig) Merge(o *VaultConfig) *VaultConfig {
 }
 
 // Finalize ensures there no nil pointers.
-func (c *VaultConfig) Finalize() {
+func (c *VaultConfig) Finalize() error {
+	// Validate that both TLSConfig and SSL are not configured simultaneously
+	if c.TLSConfig != nil && c.SSL != nil && c.SSL.Enabled != nil && *c.SSL.Enabled {
```
sanikachavan5: TLSConfig is always combined with the default enabled SSL config during Runner.init, so this rejects the intended programmatic configuration and leaves Vault partially finalized, leading to a nil-pointer panic. Please let TLSConfig take precedence over SSL as documented.

config/vault.go
```diff
@@ -166,6 +173,8 @@ func (c *VaultConfig) Copy() *VaultConfig {
 		o.SSL = c.SSL.Copy()
 	}
 
+	o.TLSConfig = c.TLSConfig
```
sanikachavan5: Copy shares the mutable *tls.Config, breaking its deep-copy contract and potentially racing with active TLS handshakes. Please use c.TLSConfig.Clone() here.

sanikachavan5 (changes requested):

config/vault.go
ncabatoff: Fixed in 5f3ead4041a42f8bcb5d1efb9431d5899df4c8e6

config/vault.go
ncabatoff: Fixed in 5f3ead4041a42f8bcb5d1efb9431d5899df4c8e6

config/vault.go
```diff
@@ -229,6 +239,10 @@ func (c *VaultConfig) Merge(o *VaultConfig) *VaultConfig {
 		r.SSL = r.SSL.Merge(o.SSL)
 	}
 
+	if o.TLSConfig != nil {
+		r.TLSConfig = o.TLSConfig
```
sanikachavan5: here as well, 
```suggestion
		r.TLSConfig = o.TLSConfig.Clone()
```

sanikachavan5 (changes requested):

sanikachavan5 (approved):
Looks good to me!

PR #2182, opened 2026-09-02
support cross-namespace Nomad service queries

### LINKED ISSUE #2072: Namespace support for `nomadService(s)`
When using the `nomadService` or `nomadServices` functions, only services from the default namespace are returned. It would be nice if we could support querying services from other namespaces (or even all namespaces via `*`).

The `nomadVar` family of functions all already support namespaces using the following syntax: `nomadVar "path/to/variable@namespace.region"`. `nomadService` supports a similar format like `nomadService "tag=value.service_name@region"`.

## Possible solutions

1. Ideally we could extend the nomad service query syntax to match the one of `nomadVar` and get `"tag=value.service_name@namespace.region"`. However, that would break backwards compatibility because right now the region is assumed to appear right after the `@`. Changing the regex to assume that `@x` would be a region `x` and `@x.y` would be namespace `x` and region `y` would break existing setups, because regions may contain `.`, so we can't assume anything before the `.` is a namespace.
2. We could add two new functions with the same functionality as `nomadService` and `nomadServices`, but with support for namespaces. Those could use an updated query syntax matching the one of `nomadVar`.
3. Since the service name can't contain a `.`, we could use a syntax like `service_name.namespace@region`. This is still confusing when seen side-by-side with a `nomadVar` invocation.
4. Use a different separator than `@` and `.` for declaring the namespace.
5. We could swap namespace and region to keep the backwards compatibility, but that is strange when seen side by side with a `nomadVar` invocation (`nomadVar "var@namespace.region"` vs `nomadService "service@region.namespace"`).
6. The namespace could be passed as an additional (optional) parameter to the `nomadService` and `nomadServices` functions:
    - `nomadServices` currently takes 0 or 1 parameters. We could add an optional second parameter, e.g. `nomadServices "region" "namespace"`.
    - `nomadService` has two possible ways of invocation, either with 1 or 3 parameters. Adding an additional optional namespace parameter to each of those overloads would be possible without any ambiguities.

I think breaking backward compatibility is out of the question, so `1.` is out. I'm also not a big fan of the _almost the same as `nomadVar` but not quite_ syntaxes (i.e. `3.`, `4.` and `5.`). `6.` would be possible but would lead to further deviation from `nomadVar`, so my preferred solution right now would be `2.`.

I'd be happy to write up a PR once there's a consensus regarding the route to take.

## Related Issues

hashicorp/nomad#14177

tgross:
Add support in the Nomad service(s) dependencies for including a Nomad namespace in the queries. The syntax for this matches the namespace syntax of the existing Consul service queries. The `nomadServices` query will also support wildcard namespaces.

Fixes: https://github.com/hashicorp/consul-template/issues/2072
Ref: https://hashicorp.atlassian.net/browse/NMD-920
Ref: https://github.com/hashicorp/nomad/issues/14177

---

In addition to the added unit tests, I built a Nomad binary with this change included and ran an end-to-end test with Nomad jobs that register themselves with Nomad service discovery. The template text of the jobspec was as follows:

```
# all services
{{ range nomadServices "?ns=prod" }}
service {{ .Name | toLower }} {{ .Tags }}{
{{end}}
# end

# one service
{{ range nomadService "httpd-web?ns=prod" }}
upstream  {{ .Address }} {{ .Port }}
{{ end }}
# end
```

The full jobspec was as follows:

<details><summary>jobspec</summary>

```hcl
job "httpd" {

  group "web" {

    network {
      mode = "bridge"
      port "www" {
        to = 8001
      }
    }

    service {
      name     = "httpd-web"
      provider = "nomad"
      port     = "www"
    }

    task "http" {

      driver = "docker"

      config {
        image   = "busybox:1"
        command = "httpd"
        args    = ["-vv", "-f", "-p", "8001", "-h", "/local"]
        ports   = ["www"]
      }

      template {
        data = <<EOF
# all services
{{ range nomadServices "?ns=prod" }}
service {{ .Name | toLower }} {{ .Tags }}{
{{end}}
# end

# one service
{{ range nomadService "httpd-web?ns=prod" }}
upstream  {{ .Address }} {{ .Port }}
{{ end }}
# end

EOF

        destination = "local/out.txt"
      }

      resources {
        cpu    = 100
        memory = 100
      }

    }
  }
}
```

</details>

Setup:

```sh
$ nomad namespace apply prod
$ nomad job run -namespace prod ./httpd.nomad.hcl
$ nomad namespace apply dev
$ nomad job run -namespace dev ./httpd.nomad.hcl
```

See that the expected services are registered with Nomad:

```sh
$ nomad service list -namespace prod
Service Name  Tags
httpd-web     []

$ nomad service info -namespace prod httpd-web
Job ID  Address              Tags  Node ID   Alloc ID
httpd   192.168.1.194:29615  []    e9d0aa14  e53a9790
```

And see that consul-template in the Nomad binary has queried the services cross-namespace:

```sh
$ nomad alloc exec -namespace dev 8a cat local/out.txt
# all services

service httpd-web []{

# end

# one service

upstream  192.168.1.194 29615

# end
```

---

## PCI review checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# UNRELEASED
+
+IMPROVEMENTS:
+* nomad: add support for querying services from specific Nomad namespaces in `nomadService` and `nomadServices` [GH-2182](https://github.com/hashicorp/consul-template/pull/2182)
+
 # 0.42.1
 
 IMPROVEMENTS:
--- a/dependency/dependency.go
+++ b/dependency/dependency.go
@@ -16,21 +16,23 @@ import (
 )
 
 const (
-	dcRe           = `(@(?P<dc>[[:word:]\.\-\_]+))?`
-	keyRe          = `/?(?P<key>[^@\?]+)`
-	filterRe       = `(\|(?P<filter>[[:word:]\,]+))?`
-	serviceNameRe  = `(?P<name>[[:word:]\-\_]+)`
-	queryRe        = `(\?(?P<query>[[:word:]\-\_\=\&]+))?`
-	nodeNameRe     = `(?P<name>[[:word:]\.\-\_]+)`
-	nearRe         = `(~(?P<near>[[:word:]\.\-\_]+))?`
-	prefixRe       = `/?(?P<prefix>[^@\?]+)`
-	tagRe          = `((?P<tag>[[:word:]=:\.\-\_]+)\.)?`
-	regionRe       = `(@(?P<region>[[:word:]\.\-\_]+))?`
-	nvPathRe       = `/?(?P<path>[^@]+)`
-	nvNamespaceRe  = `(@(?P<namespace>[[:word:]\-\_]+))?`
-	nvListPrefixRe = `/?(?P<prefix>[^@]*)`
-	nvListNSRe     = `(@(?P<namespace>([[:word:]\-\_]+|\*)))?`
-	nvRegionRe     = `(\.(?P<region>[[:word:]\-\_]+))?`
+	dcRe                     = `(@(?P<dc>[[:word:]\.\-\_]+))?`
+	keyRe                    = `/?(?P<key>[^@\?]+)`
+	filterRe                 = `(\|(?P<filter>[[:word:]\,]+))?`
+	serviceNameRe            = `(?P<name>[[:word:]\-\_]+)`
+	queryRe                  = `(\?(?P<query>[[:word:]\-\_\=\&]+))?`
+	nodeNameRe               = `(?P<name>[[:word:]\.\-\_]+)`
+	nearRe                   = `(~(?P<near>[[:word:]\.\-\_]+))?`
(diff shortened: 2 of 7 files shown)


jrasell (dismissed):
LGTM.

I think we will need a changelog entry for this too.

tgross:
Added a changelog and added the missing support for wildcards on the list endpoint.

tgross:
We won't have a PR for Nomad until this has been merged and released of course, but I've got a documentation PR for Nomad standing by in https://github.com/hashicorp/web-unified-docs/pull/3295

tgross:
~A somewhat painful part of this that came up during testing is that if you query `nomadServices` and then want to loop over those services and make subsequent `nomadService` calls, the namespace of those services is not present. This makes wildcard queries less useful. It looks like we can implement this simply on the Nomad side but then we'll need CT to update the Nomad API client to pick up the struct changes the user will need to query it. Depending on whether this gets reviewed first we can always do that under a separare PR.~

Nope! It's already in the API. Fixed here.

